Issue #43, opened 2016-08-31
epiphany section in basic infos for characters not saved


bartart3d:
When filling in the data about the characters, the epiphany section is not saved into the file.

olivierkes:
Fixed thanks to @gedakc.

PR #94, opened 2017-05-07
Fixes: epiphany section in basic infos for characters not saved #43

gedakc:
This one line commit fixes issue #43.  The change permits the epiphany section in basic infos for characters to be load/saved.

Because this commit changes the file manuskript/load_save/version_1.py, I was not sure if a new version needed to be created.

In my testing, a project containing epiphany data was able to load/save both with and without this one line change.


olivierkes:
Thanks !
I think it's safe to add this without creating a new version.

PR #103, opened 2017-05-22
Add project name to main window title

gedakc:
This is a minor enhancement to add the project name to the main window title when a project is loaded.

With project loaded:
**My Novel - Manuskript**

Without project loaded (same as before):
**Manuskript**


olivierkes:
Great, thanks !

PR #107, opened 2017-05-31
Adding german translation

ScullyBlue:
I added a german translation for the tool. Its only the informal version (german "Du"), instead of the formal language (german "Sie"), but mostly I tried to use neutral, passive sentence, so it would be not so obviously.

I hope, its in the right format. It was my first time with Qt Linguist ... Often says by generating the ts-File "cannot be called without context" and the Files für english and spanish are much longer ...


olivierkes:
Thanks a lot, I merged your translation to the `develop` branch.

olivierkes:
Hey @ScullyBlue , we're about to release manuskript 0.5.0.
Would you agree to update your german translation?
That would be awesome !

gedakc:
While we await updated translations I will manually test Manuskript functionality in an effort to minimize the chance of regressions slipping into the release.

ScullyBlue:
Oh, yes, I would update the german translation. But befor next week I have not enough time. I hope its ok, if I give it later.

olivierkes:
Entschuldigung @ScullyBlue , forgot to answer. Sure it's ok, send it when you have time :)
Thanks a lot !

PR #110, opened 2017-06-06
Change message from warning to note for failed to load translator string

gedakc:
The message for failing to find and load a language translator was
misleading.

There is no problem running Manuskript with the default English
language (no translation loaded).  The word **Warning** implied that
something was wrong with manuskript and could be confusing to users.


olivierkes:
That's wise !

PR #114, opened 2017-06-12
Request confirmation if create project would overwrite existing file(s)

gedakc:
Previously creating a new project with an existing filename would
blindly overwrite the existing project file(s).  This could result in
an author accidentally losing all of their work, assuming they did not
have a backup.

This enhancement pops up a warning asking the user if they wish to
overwrite the existing project filename.


olivierkes:
Great !

Issue #102, opened 2017-05-20
Character Importance-Slider memorize importance of last character ... partly


ScullyBlue:
I have created one character with high importance (Main) and when I create a new one, the importance-slider is also on high importance, but in the sidebar the second character stand under low importance (Minor). If I want the character move to the category Main I must set the importance-slider back to Minor and then to Main.

Not very important, but at first it confused me, because the importance of the character was high (on Slider) and the character stands in the list under low importance.

I used manuscript 0.3.0 on Linux Mint 18, 64-bit.

olivierkes:
Thanks for reporting that. Annoying stuff... I'll have a look when possible.

gedakc:
@ScullyBlue I created a Pull Request to address this issue.  The PR enhancement makes the Character Importance slider work the same way as the Plot Importance slider currently works.  More specifically with the enhancement, newly added records will default to the lowest importance.

This seemed to me to be a more correct choice because a user could be on any Character Tab (Basic infos, Summary, Notes, Detailed Infos) when the **+** button is clicked to add a new character.  Only the Basic infos tab shows the Importance slider, which is a piece of data for the current character only.

ScullyBlue:
Thanks, gedakc, that sounds good.

ScullyBlue:
My fault, wrong button ...

gedakc:
The above referenced fixes have been committed to the develop branch for inclusion in the upcoming 0.5.0 release.

gedakc:
@olivierkes can you close this report?  The fix to address the issue has already been committed (see above).

PR #121, opened 2017-06-22
Fixes: add character button does not set importance slider to default…

gedakc:
… value

See issue #102.

By default when a character is added to the character model, the
importance value is set to "0" (Minor).  However when the new
character is selected, the importance slider remains set to the value
of the previously selected character.

Steps to Reproduce:

1.  Click add character (plus sign) button.
2.  Select this "New character".
3.  Select "Basic infos" tab and change Name to "Bob".
4.  Move Importance slider all the way to the right (Main).
5.  Click add character (plus sign) button.
6.  Select this "New character".

    Note that the "New character" is shown in the character list pane
    under "Minor", but the Importance slider is still all the way to
    the right (Main).

This enhancement ensures that the corresponding character UI
importance slider is also set to the same default value.

This enhancement aligns the Character importance slider with the way the Plot importance slider works.  More specifically all new records are added at the lowest importance level.

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -156,7 +156,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         """
         c = self.lstCharacters.currentCharacter()
         if not c:
-            self.tabPlot.setEnabled(False)
+            self.tabPersos.setEnabled(False)
             return
 
         self.tabPersos.setEnabled(True)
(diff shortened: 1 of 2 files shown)


gedakc:
In the second commit listed above I fixed the numbering in the steps in the comment section of the commit.

gedakc:
@olivierkes, while my second commit does appear to fix the issue, I think there might be a more appropriate solution.

With Model-View-Controller, I think that the model is supposed to not know anything about the Views.  The Model is only supposed to issue messages when something has changed.  Views, via the Controller, are supposed to subscribe to the messages if they need to do something based on a change.

In this situation, the character model **should not** tell the character importance slider (view) to update it's values.  _I think this is what I implemented with the second commit._

I will investigate further, but I am also open to suggestions.

gedakc:
Hi @olivierkes.  I believe these new commits resolve three issues related to character importance, character tab disabling, and plot importance.  This new code should be in alignment with Model-View-Controller concepts.

Please review and either comment or commit.

olivierkes:
Seems right to me. Good work, thanks!

PR #153, opened 2017-09-29
Add about manuskript dialog

gedakc:
Add Help -> About dialog window
    
Add **Help -> About** menu entry that invokes an **About** dialog window for Manuskript.
    
Show details such as logo, version, website, copyright, license, and software versions in use (Python, PyQT, and Qt).

Additionally, single source the package version.
Define one place to store the Manuskript version so that it can be used in multiple locations.


olivierkes:
Good idea. I modified it a little bit so that we can copy/paste all the software versions in one go.

gedakc:
The modifications look good to me.  It should help us with tickets because now users can copy/paste all the software versions in one operation.  :+1:

PR #193, opened 2017-11-05
Expand german translation

ScullyBlue:
As I promissed: The update of german translation! To the best of one's knowledge and belief.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,28 +1,31 @@
 # Change Log
 
-## [Unreleased](https://github.com/olivierkes/manuskript/tree/HEAD)
-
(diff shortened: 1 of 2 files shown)


olivierkes:
Vielen Dank, @ScullyBlue !!

Your translation has been merged to develop, we might make a 0.5.1 release soon.

_(And I realize I don't understand PR much: why, when I changed the base branch from master to develop, it included the changelog commit? — but that's something else...)._

gedakc:
> (And I realize I don't understand PR much: why, when I changed the base branch from master to develop, it included the changelog commit? — but that's something else...).

@olivierkes was the _develop_ branch rebased against the current _master_ which includes the **Update CHANGELOG.md** commit b4427fb356?

If so then when you merged the PR that might explain why commit b4427fb356 was also included.

PR #203, opened 2017-11-09
[WIP] Add Travis CI support

katafrakt:
For now it builds MacOS package when tag is pushed.

This is work in progress, open for discussion. Commits should be squashed before merge, as they don't contain any valuable development history.


.travis.yml
```diff
@@ -0,0 +1,23 @@
+# according to the internet Python envs are hard and it's better to build them from scratch
+language: cpp
+os:
+- osx
+sudo: required
+install:
+- if [ "$TRAVIS_OS_NAME" = "osx" ]; then bash package/prepare_osx.sh; fi
+script:
+- python3 package/dependency_test.py
```
katafrakt: This is a very basic test if main application loads. In the future it would be cool to have some real tests in here. This is run on every push, not on tag push.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+before_deploy:
+- export FILENAME=manuskript-$TRAVIS_BRANCH-$TRAVIS_OS_NAME.zip
+- pyinstaller manuskript.spec --clean
+- cd dist && zip $FILENAME -r manuskript && cd ..
+- ls dist
+deploy:
+  provider: releases
+  api_key:
+    secure: XVcO7GsEEQxNymr7h567BxkERBZittpoC3oMZN+F2rzIoeabo/wjjBSyhMvgJKDYm1rFjQro2tr9r1ubMCjUyBP7eioOIY+7clX7pmxKPYosYgrXXx2WnaWgv8zkVk7KkXmGyFxjnp7jyTNytjFlcOeXl1cvLHum+a56vD6pINQft9V1eQ3NuPZnM729pEndL2TkI0o246QNNi1XBxIeX/o/15zx0BTMG5z8M0oGBW7VaZC6IOZ9NED7L4VGAkVizb3ONLypnQRGfFdWWE3o6gfYTLhk9fGekmlykbpOjgQcxpUe29AExKVToXfDWYUhAzaNsrpDaXXn7C0RDBVRiJYN10CiEojaZOTq7F2nrnE5Y5x1k3oXElXVAT9JQoFw935aJpEM7g5WFwmPikXoHkTFtoXWyVHcUyn2ByYuA7wu61u9NVKFvMOeH0Qg/U3IQhXvu5FekMAbDbu2H0VfKCu1+RnCepp3JFVFIPB1nr7TsmAuJDD2Z1qJ1O7S66DFWt05VLh2ruCL/lKjQIY9tP8cSaVrYEiXFBDam5JyhILYMyKWonYZWbZ4B+5Vzersq+ZjXXRkdSDHX4DuKtKiI5z2oP1ijLU0i7gkBK7UMdaKq/2l3mT3JDK88oaL5K6BeB9C50p0k7rlQtC920SvmmqDjiDelJqaCtcqzEw13H4=
```
katafrakt: This must be changed for access token having write permissions for Manuskript's main repository releases. It can be generated with Travis CLI tool.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  file: dist/$FILENAME
+  overwrite: true
```
katafrakt: Allows overwriting previously released packages. Useful for testing but can be useful for fixing errors spotted right after release or for corked packages. Requires use of `git push --force --tags`

olivierkes:
Thank you very much !

Just to be sure, you wrote `[WIP]`, are *you* still working on it? Or should it be merged now?

katafrakt:
Maybe `[WIP]` was not the best choice. My idea was that we maybe should discuss some of the options before merge. But if you don't have any comments, I can squash the commits and we can merge. I'm not planning any more work on that apart from results from discussion.

olivierkes:
OK :) so I'll read that carefully soon, and we'll wait also for @gedakc comments if he has any.

gedakc:
First, thank you @katafrakt for your efforts to get manuskript to automatically build for Mac OS X.  I'm sure there was lots of learning involved to figure out the how Travis CI works, and also how to enhance Manuskript to work properly with Travis and PyEnchant.

Next, a disclaimer on my part.  I'm fairly new to lots of the github features so I'm not exactly sure how to work with multiple conversations on code sections.  As such I will try to summarize my comments in a this single comment.

1.  The automated testing sound like a good idea to me.  The dependency test you created is good start.

2. Thanks for the tip about needing to change the secure API key to one that will permit writing into @olivierkes project release area.  
   Is there a web link you could provide to inform either Olivier or myself how to do this?

3. Regarding the `overwrite` setting perhaps it might help to place a comment in the `.travis.yml` file indicating that by setting `overwrite` to true then `git push --force --tags` will create a new PyInstaller release.

4. Does the PR enable PyEnchant for Mac OS X?  
   From looking through the code it appears to me that the pyenchant stuff is commented out.  Then again I'm still learning github so maybe it is an error on my part.

Again thanks hugely for helping with Manuskript.  :-)

katafrakt:
> Is there a web link you could provide to inform either Olivier or myself how to do this?

[Here](https://github.com/travis-ci/travis.rb#readme) is the repo for Travis CLI. You need to install it (basically it's a Ruby gem, so `gem install travis`) and then run it with `travis setup releases`. It asks for credentials then asks a couple of questions, after which it generates a `.travis.yml` file with secure token.

This is a process I have found, I'm not sure if there is easier way. Steps I described were mentioned [in the docs](https://docs.travis-ci.com/user/deployment/releases/#Authenticating-with-an-OAuth-token).

> Does the PR enable PyEnchant for Mac OS X?

No, I haven't tried that yet.

.travis.yml
olivierkes: What does it mean on every push? Every time any one pushes anything to the repo? And then where do you see the output? (Sorry if it's a stupid question, I'm not sure I understood exactly how travis works yet :p)

olivierkes:
Looks good to me. (Just asked a question above, but that's a detail)
Thanks again !!

> Here is the repo for Travis CLI. You need to install it (basically it's a Ruby gem, so gem install travis) and then run it with travis setup releases. It asks for credentials then asks a couple of questions, after which it generates a .travis.yml file with secure token.

I guess we can also get a token on https://travis-ci.org/, in case.

@gedakc, do you have time to do that? Or would you like me to do it?

gedakc:
>> Here is the repo for Travis CLI. You need to install it (basically it's a Ruby gem, so gem install travis) and then run it with travis setup releases. It asks for credentials then asks a couple of questions, after which it generates a .travis.yml file with secure token.

> I guess we can also get a token on https://travis-ci.org/, in case.

> @gedakc, do you have time to do that? Or would you like me to do it?

I can take a look at getting this setup.  I am sure that it will work for your account, and I think it should work for me because you granted me write access to the repository.

.travis.yml
katafrakt: >What does it mean on every push? Every time any one pushes anything to the repo? 

Yes. Also, on every pull request made etc.

>And then where do you see the output?

Not a stupid question at all ;) You can see history of builds for my fork here, for example: https://travis-ci.org/katafrakt/manuskript/builds

Also, you'd need to enable Travis on main repository. It should be possible here: https://travis-ci.org/profile/olivierkes?offset=0

olivierkes:
Thanks for the answers to my questions [above](https://github.com/olivierkes/manuskript/pull/203#discussion_r150383567).

So I authorized travis-co.org. I'll see for the oauth token now.

Just one more question: why is language `cpp` and not `python`?

gedakc:
My guess is that @katafrakt set the language to `cpp` to enable more control in the future over which version of `python` is installed.  Currently the `package/prepare_osx.sh` installs `python3`, not `python` which is version 2.

@olivierkes since you've started looking into this already, I will hold off working on this auto build enhancement.  I am available for testing if/when needed.

katafrakt:
>Just one more question: why is language cpp and not python?

I have no experience in setting up Travis for Python projects, but I read somewhere that it's easier to set up `cpp` (it is most basic setup as for dependencies) and build Python env manually, than telling Travis to build it for you. So @gedakc is right.

katafrakt:
I squashed the commits and here's first ever build for manuskript (of course, doing not very much :wink: ): https://travis-ci.org/olivierkes/manuskript/builds/300958039

So after fixing things with access token I thinks we are good to go.

olivierkes:
Thanks a lot @katafrakt!

olivierkes:
And it passes ;)

image

Issue #188, opened 2017-11-04
Manuskript does not start with PyEnchant on MacOS


katafrakt:
This happens with PyEnchant enabled on MacOS.

```
Debug: Web rendering engine used: QWebEngineView
Running manuskript version 0.5.0.
Note: No translator found or loaded for locale en_US.
Traceback (most recent call last):
  File "bin/manuskript", line 13, in <module>
    main.run()
  File "/Users/pawel/dev/github/manuskript/bin/../manuskript/main.py", line 64, in run
    launch()
  File "/Users/pawel/dev/github/manuskript/bin/../manuskript/main.py", line 70, in launch
    main = MainWindow()
  File "/Users/pawel/dev/github/manuskript/bin/../manuskript/mainWindow.py", line 54, in __init__
    self.setupUi(self)
  File "/Users/pawel/dev/github/manuskript/bin/../manuskript/ui/mainWindow.py", line 169, in setupUi
    self.txtSummarySentence = textEditView(self.tabSummaryPage1)
  File "/Users/pawel/dev/github/manuskript/bin/../manuskript/ui/views/textEditView.py", line 79, in __init__
    self._dict = enchant.Dict(self.currentDict if self.currentDict else enchant.get_default_language())
  File "/Users/pawel/.pyenv/versions/3.5.3/Python.framework/Versions/3.5/lib/python3.5/site-packages/enchant/__init__.py", line 555, in __init__
    raise Error(err)
enchant.errors.Error: No tag specified and default language could not be determined.
```

Brief investigation indicates that this is because PyEnchant cannot detect default language. In fact, when running a simple script:

```
import enchant;
print(enchant.list_dicts())
default = enchant.get_default_language()
print(default)
enchant.Dict(default)
```

The result is:

```
[('de_DE', <Enchant: Myspell Provider>), ('en_AU', <Enchant: Myspell Provider>), ('en_GB', <Enchant: Myspell Provider>), ('en_US', <Enchant: Myspell Provider>), ('fr_FR', <Enchant: Myspell Provider>)]
None
Traceback (most recent call last):
  File "package/dependency_test.py", line 14, in <module>
    enchant.Dict(default)
  File "/Users/pawel/.pyenv/versions/3.5.3/Python.framework/Versions/3.5/lib/python3.5/site-packages/enchant/__init__.py", line 555, in __init__
    raise Error(err)
enchant.errors.Error: No tag specified and default language could not be determined.
```

gedakc:
[...]

PR #212, opened 2017-11-13
Add PyEnchant support to OSX builds

katafrakt:
After fixing #188 PyEnchant builds pass and seem to generate working build with spellcheck enabled.

### DIFF
--- a/package/prepare_osx.sh
+++ b/package/prepare_osx.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -ev
 brew update
-brew install python3 # enchant
+brew install python3 enchant
 sudo pip3 install --upgrade pip setuptools wheel
-pip3 install pyinstaller PyQt5 lxml # pyenchant
-# brew install qt hunspell
+pip3 install pyinstaller PyQt5 lxml pyenchant
+brew install qt hunspell
 # fooling PyEnchant as described in the wiki: https://github.com/olivierkes/manuskript/wiki/Package-manuskript-for-OS-X
-# sudo touch /usr/local/share/aspell
+sudo touch /usr/local/share/aspell


katafrakt:
(@olivierkes I invited you as collaborator to my fork, because without it I cannot test deploying builds with your access token)

olivierkes:
I accepted the invitation. Tell me if I should do something else. I can even set you as a collaborator on the repo here if it helps.

katafrakt:
Weird, it still says that access is denied. I don't really have an idea how to test it further until you release new version of manuksript and we can be certain that this secure token works.

olivierkes:
Didn't you say we coud force a build for force pushing an already existing tag?

katafrakt:
We could. You'd need to re-tag and push force, for example:

```
git checkout master
git tag 0.5.0 --force
git push --tags --force
```

katafrakt:
Disregard my deleted comment 😉 

This would not work with 0.5.0, as Travis config is not merged into master. You could, however, create some additional tag such as `0.6.0-devel` and tag `develop` branch with it. Then it would build from current contents of develop branch.

olivierkes:
I just did that: https://github.com/olivierkes/manuskript/tree/0.5.0-devel

It's building. Suspense.

katafrakt:
Ok, looks like it worked (https://github.com/olivierkes/manuskript/releases/tag/0.5.0-devel) so token is valid. Maybe it's per-repository and that's why I can't use it in my fork. Nevermind, though, I will find another way to test that.

olivierkes:
Great :)

olivierkes:
I deleted the release to remove confusion for people who might want to download manuskript.

olivierkes:
@katafrakt, do you see a way to parameter travis so that it builds package for every commit in the develop branch, and uploads them to an FTP location?

I'm doing it for `.deb` packages (www.theologeek.ch/manuskript/releases/manuskript-develop.deb is the latest develop commit).

It would be useful for testers.

katafrakt:
I don't think it's possible to upload build from Travis to FTP, as you'd have to keep FTP credentials in publicly accessible file in the repository.

Probably valid architecture to do so would be to have some server creating and uploading builds and install a Github hook notifying it when to create a new build.

olivierkes:
With SemaphoreCI, I can define environnment variable that are encrypted, and use them in an ftp command (`lftp -c "open -u hfpn_semaphoreci,$FTP_PASSWORD www.theologeek.ch;"`)

olivierkes:
It seems to be possible with Travis too:

- FTP: https://docs.travis-ci.com/user/deployment/custom/#FTP
- Encrypted var: https://docs.travis-ci.com/user/encryption-keys/

I'll see if I can do it.

katafrakt:
Okay, it seems to be possible with encrypted vars https://docs.travis-ci.com/user/deployment/custom/

olivierkes:
;)

olivierkes:
I think it works. @katafrakt, could you try www.theologeek.ch/manuskript/releases/manuskript-osx-develop.zip and check if it works, just to be sure?
Thanks a lot :)

katafrakt:
@olivierkes It works 😉 

However, I discovered that when I'm opening it by double-clicking in Finder, it opens a terminal window in a background. That's not a big deal, but I just want to let you know. Also, I don't know if it's something new, as up until now I always started Manuskript from terminal.

screen shot 2017-11-30 at 17 54 16

olivierkes:
Thanks @katafrakt!

If I remember correctly the terminal thing was on purpose: since manuskript is not mature yet I thought it could help to report issues. Do you think it brings confusion?

katafrakt:
I don't know, really. I see it appearing with other applications too, so I guess it's not that strange. However, I'm a programmer, so I'm used to terminal. Not sure about "normal people" 😉

gedakc:
> @katafrakt I discovered that when I'm opening it by double-clicking in Finder, it opens a terminal window in a background.

This is the same as what occurs in Windows.

I think this is okay while manuskript is undergoing significant development.  It enables users to more easily include the console messages when reporting issues.

In the future when manuskript is out of *pre-release* status I think we might try disabling the console window by setting `console=False` in the [manuskript.spec](https://github.com/olivierkes/manuskript/blob/d32ee37e84b4e7e3732f75eb36fbe52c77e8171a/manuskript.spec#L32) file.

Issue #181, opened 2017-10-31
Odd word choices in English - Take 2


gedakc:
This issue is a continuation of suggested enhancements to improve English wording.  This first post includes two suggestions from issue #53 that will not make it into the 0.5.0 release.

* Serie -> Series

_Note that there is a trailing "**s**"_

* Informations -> Information

_Note that there is no trailing "**s**"_

gedakc:
More suggestions:

- Recents -> Recent

- Fix sentence structure for the following:

FROM:  
The basic situation, in the form of a &apos;What if...?&apos; question. Ex: &apos;What if the most dangerous evil wizard ~~could~~ wasn&apos;t able~~d~~ to kill a baby?&apos; (Harry Potter)

TO:  
The basic situation, in the form of a &apos;What if...?&apos; question. Ex: &apos;What if the most dangerous evil wizard wasn&apos;t able to kill a baby?&apos; (Harry Potter)

PR #231, opened 2017-11-23
Change words issue 181

gedakc:
This pull request includes enhancements to address issue #181 - Odd Word Choices in English - Take 2.

### DIFF
--- a/i18n/manuskript_de.ts
+++ b/i18n/manuskript_de.ts
@@ -784,7 +784,7 @@ Nutze das, wenn du YAML-Errors bekommst.</translation>
     </message>
     <message>
         <location filename="../manuskript/ui/mainWindow.ui" line="154"/>
-        <source>Book informations</source>
+        <source>Book information</source>
         <translation>Buchinformationen</translation>
     </message>
     <message>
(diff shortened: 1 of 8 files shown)


gedakc:
Hi @olivierkes,

For your information, when I created this PR it looks like a number of processes were initiated on TravisCI/SemaphoreCI.

See above.

olivierkes:
Thanks for doing this!
I might be wrong, but I think you missed one "Serie" in `manuskript/ui/mainWindow.ui` (there is 3 changes in mainWindow.py, but only two in mainWindow.ui).

(I don't know why Semaphore triggered stuff here. It's normal that it fails, because the test framework has not yet been pushed to master. In the future this will tell us for any PR if it passes the tests)

gedakc:
> @olivierkes: I think you missed one "Serie" in manuskript/ui/mainWindow.ui (there is 3 changes in mainWindow.py, but only two in mainWindow.ui)

Thanks for catching this.  I'll take a look and re-push if/when I fix the issue.

gedakc:
I had missed some of the "Serie" -> "Series" changes.

Note that I did not change the variable names.  I also did not change the entries in `manuskript/load_save/version_1.py` because these might adversely impact manuskript's compatibility with previous versions.  I wasn't sure so I left these alone.

olivierkes:
Great! Haven't tested it, but looking through it it looks good. You're right, it's safer not to touch things `version_1.py`, or that would require more extensive testing.

(Do you have the right to merge a PR btw?)

gedakc:
When I view the PR under the `olivierkes/manuskript` github repository, I can see a **Merge pull Request** button, so I assume I have permission.

Would you like me to merge the PR?

olivierkes:
> Would you like me to merge the PR?

Yet do it, so that we know you can.

gedakc:
I chose the **Rebase and merge** option in an effort to keep the git tree simple (less forks).  If you are not familair with `gitk`, install it and take a look at the output from `gitk --all`.

Issue #282, opened 2017-12-29
Install on MacOsX failed


cfrancois7:
Dear everyone,

I tried to install Manuskript on my Mac, but I failed. See the error message:

```
$ /Users/myName/Downloads/manuskript/manuskript ; exit;
[32512] Error loading Python lib '/Users/myName/Downloads/manuskript/Python': dlopen: dlopen(/Users/myName/Downloads/manuskript/Python, 10): Symbol not found: _clock_getres
  Referenced from: /Users/myName/Downloads/manuskript/Python (which was built for Mac OS X 10.12)
  Expected in: /usr/lib/libSystem.B.dylib
 in /Users/myName/Downloads/manuskript/Python
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.
```

It seems the builder looks for the `python`directory inside the manuskript repository, but there is no `python` dir inside.

gedakc:
Which version of manuskript are you using (0.6.0 is the latest)?

Are you [running from source code on OS X](https://github.com/olivierkes/manuskript/wiki/Run-manuskript-on-OS-X), or did you [install from the PyInstaller binary package](http://www.theologeek.ch/manuskript/download/), for example `manuskript-0.6.0-osx.zip`?

cfrancois7:
I followed the instruction from "http://www.theologeek.ch/manuskript/download/".
It was `manuskript-0.6.0-osx.zip`.

version of MacOsX: 10.11.6 (El Capitan)

gedakc:
The current Mac OS X PyInstaller package is automatically built using [Travis CI](https://travis-ci.org/).  It appears that the current [default for Travis CI is Mac OS X 10.12.6](https://docs.travis-ci.com/user/reference/osx/#OS-X-Version).

I suspect that the `manuskript-0.6.0-osx.zip` PyInstaller package does not work on your computer because you are running an older version of Mac OS X.

@cfrancois7, another option you might consider is to [Run Manuskript from Source Code on Mac OS X](https://github.com/olivierkes/manuskript/wiki/Run-manuskript-on-OS-X).


### Question for @olivierkes and/or @katafrakt:

Do you think it would be feasible to automatically build the Mac OS X PyInstaller package using an older version such as **OS X 10.10** which is listed as supported by Travis CI?

[...]

PR #287, opened 2018-01-02
Build MacOS release with XCode 7.3 image

katafrakt:
7.3 is the oldest image available on TravisCI that works. Older 6.4 version does not play well with Enchant, resulting in a build error. This could probably be averted by trying to build with 6.4 and without Enchant support, but I'm not sure if it's worth it.

Relates to #282, tested with El Capitan and Sierra.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: cpp
 os:
 - osx
+osx_image: xcode7.3
 sudo: required
 install:
 - if [ "$TRAVIS_OS_NAME" = "osx" ]; then bash package/prepare_osx.sh; fi


katafrakt:
Travis had a worse day yesterday. Could you rerun the build? It should be ok now.

olivierkes:
Thanks @katafrakt. I restarted the build.

olivierkes:
Thanks, it passed.
Merging to develop, since anyway new releases are created on tag, so next one will be with 0.7.0. Right?

katafrakt:
Yes, new releases are published on tag. So depending on when 0.7.0 is planned to go live, you may wait or backport it to master and release 0.6.1 for macOS.

Issue #252, opened 2017-12-03
New character templates.


siliconserf:
Feature Request: User Editable Templates for Character, World, and Plot entries. 

In the current version, Manuskript allows adding characters to the project. You hit the '+' icon and the code creates the new character object from a hard coded template. What I would like to see is to have a Template section (above Debug) that has user editable versions of addable items. This allows "pre-loading" items like "Characters" with "boilerplate" information. As an example, there are questions that I would like to have show up (optionally) in the Character/Notes section when I add a new character. Snowflake includes a lot of suggested data entries for a character bio. Currently you have to add each of those by hand for each character. Multiple character templates would support different defaults for different types of characters (important, secondary, walk-on, and only referred to).

olivierkes:
Thanks for those ideas.

Right now, I'm refactoring the code, and I want to make characters/word/plot items similar to outline items (editor / text items). This would simplify a lot coding because the same code to search for one type of item could be use for the other.

But this could also allow for more radical UI changes. For example:

![image](https://user-images.githubusercontent.com/1192933/33555359-60f20832-d900-11e7-98b5-d6e13bc7904c.png)

Some possibilities:

  - Every of those items *could* be regrouped in the same view (as in the fake screenshot above)
  - We could use split view to have a parallel view of a text and a character for example.
  - Characters and plots could be within "folders" or "groups" (instead of having one long list categorized only by importance)
  - The metadata pane could be used also for characters
  - …
  - And finally, to answer your initial post, a template system could be created for any of those type of items (I'm thinking of a normal folder that we would create item within, and one we declare this folder as "template folders" any item within can be used as a base for creating a new item)

What do you think? (I'd like also to have @hugo-id and @gedakc's take on that)

olivierkes:
(and also @thoronaug ^^)

siliconserf:
Excellent! That is very much what I had in mind. And I thought the logic would apply nicely to other kinds of objects. Glad to see the efficiency of coding supports this.

anicede:
That tree view is marvellous, @olivierkes, it seems it might save a lot of clicking back and forth.

One thing I've been hungry for is plots in plain text rather than xml as I do a lot of outlining on the go on my tablet. Would this be possible?

gedakc:
Thanks @olivierkes for asking my opinion.

[...]

PR #294, opened 2018-01-06
Fix a number of typos in the code

sebras:
The last commit "Fix typo in pandoc file format" is perhaps controversial. I don't know the code well enough to determine if this will prevent Manuskript from loading already saved files. Perhaps it needs more work, but I need guidance in that case.


sebras:
I noticed one of these while translating the software to Swedish, so I that prompted me to fix it.

gedakc:
Thank you @sebras for your work to fix typos in the code.

I just had a *very quick* look.  Could you back out the "fix typo in pandoc file" commit?

The reason is because the word "Serie" is used in the stored file format.  If this word is changed then we must add another version to load_save.  Because this typo is not visible in the user interface I think we should hold off on fixing it until other improvements require an update to the file format.

sebras:
Done. Though I think there is a high likelihood that the typo in the file format will remain forever.

I'm not sure if this is a typo which is Manuskript-specific or if it is related to pandoc itself, but I sense it is the former.

gedakc:
@olivierkes will you have a chance to review this PR?

If not then I can take a more detailed look at it in the next few days.

olivierkes:
Thank you very much @sebras! 

I'm very honored that you took the time to correct my terrible spelling. (I could be tempted to excuse myself saying English is not my first language, but I fear I would make as many mistakes in french ^^…)

Anyway I'll have to test it (in a few days). As @gedakc pointed out, we'll need to merge to develop, but that should not change much. And as for the typo in the file format, I could fix it with a hack so that it doesn't remain there forever. (But what is the typo again? Is it Series → Serie ?)

gedakc:
> But what is the typo again? Is it Series → Serie ?

Currently "[Serie](https://github.com/olivierkes/manuskript/blob/d32ee37e84b4e7e3732f75eb36fbe52c77e8171a/manuskript/load_save/version_1.py#L134)" is used in the file format.  The correct spelling is "Series" (with the "s" at the end).

> ... as for the typo in the file format, I could fix it with a hack so that it doesn't remain there forever.

Adding a hack usually increases the complexity of the code unnecessarily.  My vote would be to:

- Either ignore the spelling mistake because it's only in the file format and likely only seen by programmers.
- Or fix the spelling mistake properly by adding another file format version.  This option does seem like overkill to me if done only for such a simple misspelling.

If we end up changing the format for characters (issue #252) then that might be a good time to slide in a correction for the file format spelling mistake "serie" -> "series".

sebras:
> Adding a hack usually increases the complexity of the code unnecessarily

You could have the parser accept both "Serie" and "Series" but only write new files using "Series". That way it is fixed apart from old files until they are saved anew. Would that be suitable for you?

sebras:
I did a new version rebased on the **develop** branch.

gedakc:
> You could have the parser accept both "Serie" and "Series" but only write new files using "Series". That way it is fixed apart from old files until they are saved anew. Would that be suitable for you?

My preference is for no such hack to be added to the code.  Of course that is my opinion only.  @olivierkes has the final say as he created Manuskript.

sebras:
> My preference is for no such hack to be added to the code.

I'm happy either way, I just wanted for you (both) to consider it. :)

gedakc:
I just thought of another reason why I would prefer not to do the hack, and that is because it would break backward compatibility with older versions of Manuskript that recognize and load version 1 files.

Imagine the following scenario:

1. User A creates a project using 0.6.0 and passes it to a User B.  This fie is saved under file format **1**.

2. User B opens the project using upcoming 0.7.0 (which includes the hack), makes some updates, and passes it back to User A.  The file format is still saved as version **1**.

3. User A with 0.6.0 is now unable to open the project due to the file format "Serie" -> "Series" change.  The file format says it is version **1**, but in fact it is no longer the original version **1**.  In my opinion this breaks proper version control.

I still think that if we are going to fix this spelling mistake in the file format, then it must be made when a new file format version is created (e.g., **2**).

sebras:
> this breaks proper version control

Ah, yes, I didn't think about sharing of documents.

> it must be made when a new file format version is created

Agreed, then it should probably happen along with fixing issue #252 as you mentioned before. :)

olivierkes:
Good point @gedakc, I agree with that view. Hadn't thought about sharing documents either. 

So we leave it with the typo until next document format.

gedakc:
Thank you @sebras for this patch set of spelling corrections.  I have reviewed the changes, tested the code and all appears to be well.  Hence I committed these changes to the **develop** branch.

One odd thing that occurred is that this PR did not automatically close.  I suspect that it is due either to being originally based on *olivierkes:master*, or because I rebased the change against the latest **develop** branch.  I tried editing the title properties to indicate the enhancements were for the **develop** branch but this does not seem to have had any effect.

Regardless the git repository looks okay and the two commits are now part of the **develop** branch.

sebras:
> One odd thing that occurred is that this PR did not automatically close.

Ok, I might have done something silly. I know git quite well, but how github works is sometimes still a mystery to me...

> Regardless the git repository looks okay and the two commits are now part of the develop branch.

Excellent. I'm happy to hear that. Good luck with the next release! :)

Issue #273, opened 2017-12-18
Spell check is crashing the program


v-i-o-l-e-t:
Hey there!

I just downloaded Manuskript about 20 minutes ago and am loving all of it, but I've had some problems with spell check.

-It's enabled, but not checking words. I've tried just typing random letters and the program doesn't underline any of them or bring up an option to fix when I right-click them.
-I tried pressing F9 to manually check and the program crashed
-I also tried disabling spell check, but the program crashed. When I started again it was still enabled.
-The italics, bold and underline buttons also don't work, but if I type the formatting for them directly in the box, everything is fine. This probably isn't related, but thought I would bring it up.

I'm running version 0.6.0 on Windows 10 (if you need more information, I'll try to find it. Not much of a computer person.)

It's not a huge problem as I'll export to Word when I'm finished but I thought you should know.

Thanks in advance!

Edit:
Okay, so, I went into full screen and enabled spell check, and now it's working in full screen, no crashing at all. I should have mentioned earlier that when trying to enable spell check, I was in the fiction layout, no full screen. I'm assuming spell check only working in full screen is not a feature, but if it is, sorry for creating a bug thread.

v-i-o-l-e-t:
Okay, so, I went into full screen and enabled spell check, and now it's working in full screen, no crashing at all. I should have mentioned earlier that when trying to enable spell check, I was in the fiction layout, no full screen. I'm assuming spell check only working in full screen is not a feature, but if it is, sorry for creating a bug thread.

gedakc:
Thank you @v-i-o-l-e-t for your interesting in Manuskript.

Would you be able to provide a screen shot or a sample of the text that demonstrates the spell check problem?

The spell checker *should* work in both normal and full screen modes.

Following is a screen shot illustrating spell check working on Manuskript 0.6.0 in Windows 10.

![manuskript-0 6 0-win10-spellcheck](https://user-images.githubusercontent.com/10405019/34170001-81454e62-e4a6-11e7-9a0d-9c9949e60ea2.png)

v-i-o-l-e-t:
Well, this is embarrassing, but it's working now. I went to take a screenshot and tried enabling it once more and there's no problem. Both the normal and full screen view are highlighting the misspelled word. I haven't changed anything since I reported this, so I'm not sure what's changed. I can still provide screenshots, if you'd like.

[...]

PR #303, opened 2018-01-15
Avoid crash on spellcheck by ensuring enchant dictionary exists

gedakc:
See issue #273.

@olivierkes do you think we should also test for dictionary existence in [textEditView::setDict(self, d)](https://github.com/olivierkes/manuskript/blob/dc9d67e8890e44d1c8d3fecc14c6b469552b4570/manuskript/ui/views/textEditView.py#L428), or is this only called using a known to exist dictionary listed in the GUI?

### DIFF
--- a/manuskript/ui/views/textEditView.py
+++ b/manuskript/ui/views/textEditView.py
@@ -427,14 +427,15 @@ class textEditView(QTextEdit):
 
     def setDict(self, d):
         self.currentDict = d
-        self._dict = enchant.Dict(d)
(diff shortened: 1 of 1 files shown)


olivierkes:
I think it doesn't cost much in terms of performance to add a check in `setDict`, just to be on the safe side.

gedakc:
Thanks for your input.  I will add a check for dictionary existence to each call of `setDict`.

gedakc:
This PR has been merged into the **develop** branch to be included in the next release of Manuskript.

PR #310, opened 2018-01-29
Spelling: LaTex, Pandoc, XeLaTex Unicode. https link

comradekingu:
As per
https://www.latex-project.org/
https://pandoc.org/installing.html
https://www.sharelatex.com/learn/XeLaTeX
https://en.wikipedia.org/wiki/Unicode

### DIFF
--- a/manuskript/exporter/manuskript/HTML.py
+++ b/manuskript/exporter/manuskript/HTML.py
@@ -15,8 +15,8 @@ except ImportError:
 
 class HTML(markdown):
     name = "HTML"
-    description = qApp.translate("Export", "Basic HTML output using python module 'markdown'.")
-    InvalidBecause = qApp.translate("Export", "python module 'markdown'.")
+    description = qApp.translate("Export", "Basic HTML output using The Python module 'markdown'.")
+    InvalidBecause = qApp.translate("Export", "Python module 'markdown'.")
     icon = "text-html"
 
     exportVarName = "lastManuskriptHTML"
--- a/manuskript/exporter/manuskript/HTML.py
+++ b/manuskript/exporter/manuskript/HTML.py
@@ -15,7 +15,7 @@ except ImportError:
 
 class HTML(markdown):
     name = "HTML"
-    description = qApp.translate("Export", "Basic HTML output using The Python module 'markdown'.")
+    description = qApp.translate("Export", "Basic HTML output using the Python module 'markdown'.")
     InvalidBecause = qApp.translate("Export", "Python module 'markdown'.")
     icon = "text-html"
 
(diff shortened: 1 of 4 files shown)


gedakc:
Thanks @comradekingu for the patch with spelling mistake corrections, and for providing links to make verification of the upper/lower case spellings easier.

One very minor question I have relates to the spelling of *pandoc*.  In the *pandoc* page link I generally observe *Pandoc* with an upper case "P" when it is used at the being of a sentence.  Most often I observed a lower case "p" *pandoc* when it was used after the first word of a sentence.

For example under the [Compiling from source](https://pandoc.org/installing.html#compiling-from-source) section it includes the following portion of a sentence with a lower case "p" pandoc:

_...if you want to hack on pandoc or use a non-released version..._

My question is why should be we upper case "P" for pandoc in the middle of the sentence?

Personally I'm okay with either upper or lower case"P" spelling of *pandoc*.

comradekingu:
@gedakc Saw the title and the logo. Very consistent oddity. Solved by putting it at the beginning of the sentence.

gedakc:
@comradekingu the updates you made work for me.  :-)

If I don't hear any objections then I will commit these changes to the **develop** branch in the next day or so.

gedakc:
@comradekingu it appears that you continue to find spelling mistakes.  This is good and we appreciate the patches.  :-)

Because this appears to be a work-in-progress I will hold off on committing any changes.

When you are complete then please let me know.  At that time I will review the changes again and they can be collapsed into a single commit and merged with the **develop** branch.

manuskript/exporter/pandoc/plainText.py
```diff
@@ -18,7 +18,7 @@ class markdown(abstractPlainText):
 
 class reST(abstractPlainText):
     name = "reST"
-    description = qApp.translate("Export", """reStructuredText is a lightweight markup language...""")
```
gedakc: I think it would be an improvement to remove the ellipses "..." and end the statement with a period.

manuskript/exporter/manuskript/HTML.py
```diff
@@ -15,8 +15,8 @@
 
 class HTML(markdown):
     name = "HTML"
-    description = qApp.translate("Export", "Basic HTML output using python module 'markdown'.")
```
gedakc: Another minor suggestion.  I think the word "the" should be in lowercase.  For example:  "Basic HTML output using **t**he Python module 'markdown'."

gedakc:
Thank you @comradekingu for finding these spelling mistakes and creating a PR to address these.

I have squashed the multiple commits into a single commit and pushed these to the **develop** branch for inclusion in the next release of manuskript.

Issue #283, opened 2017-12-29
Underline causes false spelling error


siliconserf:
I tried this on 0.6.0 released (Win and Ubuntu), and latest development version (installed today). Keying in an underline to get italics causes spell checking to highlight the leading word as misspelled with an underline and no italicizing of that word. (The underline seems to over ride the italics effect - shouldn't they combine?) Any following words are correctly italicized. Without any italics the same word does not trip a spelling error. Using asterisks to get bold doesn't cause any problem. Here is a snippet:

![image](https://user-images.githubusercontent.com/32206614/34446741-851cf96a-ec92-11e7-95b5-9beaa7110f56.png)

siliconserf:
Closely related to the above: I found a problem attempting to set the word _is_ in italics in a sentence. The editor detects the start italics underscore, but fails to detect the closing underscore. "is" gets italicized along with all following text until the editor finds another underscore further on in the text. I think there is a problem with the editor's response to underscores. Fixing one problem might well clear the second.

gedakc:
Thank you @siliconserf for providing details on this issue including a screen shot and your thoughts on the related issue with *Italics* highlighting.

I created Pull Request  #316 to fix the issue which also includes screen shots before and after.

If you are able to test the PR then that would be great.

gedakc:
In an effort to make sure the enhancement in PR #316 works correctly, I tested with a number of combinations and permutations of *Italic* and **Bold** text.

![manuskript-bold-italic-tst](https://user-images.githubusercontent.com/10405019/36439319-c38f3f9e-1629-11e8-893e-355d23ef1dd2.png)

As best I can tell, the enhancement makes it so that all the tests I created worked properly.

gedakc:
The enhancement in PR #316 has been committed to the **develop** branch for inclusion in the next release of manuskript.

PR #316, opened 2018-02-15
Make word match for spell check exclude underscore from words

gedakc:
See issue #283.

Adds (?!_) to perform negative lookahead to exclude "\_" from pattern match.

https://stackoverflow.com/questions/14858346/regular-expressions-how-to-express-w-without-underscore

Following are example screen shots before and after applying this patch.

**Before** patch which shows words with underscores falsely identified as spelling errors:

---
manuskript-before-patch

---

**After** patch which correctly identifies spelling errors:

---
manuskript-after-patch

---


gedakc:
The enhancement contained in this PR has been committed to the **develop** branch for inclusion in the next release of manuskript.

Issue #334, opened 2018-03-26
Show character count progress indicator


lechbaczynski:
Feature: Show characters (with spaces) count for the whole text, here:

![obraz](https://user-images.githubusercontent.com/1127623/37926701-519350ea-3138-11e8-8ddd-3c646631a354.png)

gedakc:
Thank you @lechbaczynski for your interest in Manuskript.  I noticed you also provided a PR.

Normally publishers, writers, and sometimes readers are interested in the word counts for works of fiction and non-fiction.  However in my experience I have not encountered a situation where the character count is important.

What is the purpose of adding character count, and who uses it?

One reason I ask is because Manuskript has been criticised in the past for being too cluttered.  We are trying to keep distractions to a minimum so that writers can focus on creating their work.  That is why I am questioning the need for a character count statistic in the editor.

Regards,
Curtis

lechbaczynski:
For most of (over 90% I suppose) of writing contests  - short story contests, and also some novel contests, etc., in Poland there is a limit in the rules, given in characters count. 
like for example 10.000 chars including spaces, or 27.000. Sometimes there is also a minimum char count (incl. spaces) stated in the rules.

See some examples:

"znaków" - means characters
"konkurs" - competition
"tys." - thousands
"opowiadanie" - short story

![obraz](https://user-images.githubusercontent.com/1127623/38460182-365efc30-3ab5-11e8-8cd3-e8020ad06df7.png)



The same for some novels ("powieść") contests:

![obraz](https://user-images.githubusercontent.com/1127623/38460217-a8259bd0-3ab5-11e8-9ffc-9981772ebbae.png)

lechbaczynski:
Real life example - yesterday I finished writing short story for competition, 
the limit was 10.000 chars including spaces.
I wrote in Manuskript run form sources form my forked github repo - with char count feature.
The feature was very useful, see the sum of characters in all scenes on the screenshot:


![obraz](https://user-images.githubusercontent.com/1127623/38460406-7489a9ac-3ab9-11e8-85ce-ffb7f7b35393.png)

lechbaczynski:
Also, other program have it
minimalistic Focus Writer, see:

http://cdn.osxdaily.com/wp-content/uploads/2011/09/focuswriter-theme.jpg

and even Scrivener:
https://www.literatureandlatte.com/assets/image/scrivener/features/windows/win-1-binder-editor-inspector.png

gedakc:
[...]

PR #339, opened 2018-04-09
- adding characters count. Implementing #334

lechbaczynski:
@gedakc , you made me do rebase interactive, cherry pick and squash ;-)


gedakc:
With manuskript we try to support screen sizes as small as 1024x600 pixels that are popular with netbooks (small laptops).

manuskript-char-count-obscure-word-count

With this patch applied I observed two issues:

1.  The format of the character count number does not match the format used for other values, such as word count and word goal.
2. The display of the character count causes the word count to be partially hidden, and the word goal to disappear.

The first issue should be easy to fix, but the second issue is not as easy due to the reduced screen area available on small laptops.

My thoughts are that if a writer is trying to attain a specific goal for the amount of written material, this is either specified in a word count, or in a character count.

If this is the case then to work around this issue perhaps we should have a setting that can toggle the display to either words (default) or chars.

What are your thoughts?

gedakc:
Your view on the GUI changes required to support goals in either words or chars as defined by a program setting seem correct to me.

A different approach might be to only display the chars count if there is sufficient screen space to display the chars couny in addition to the word count and word goal.  This would likely be a smaller change to the code base and would perhaps still meet your need for a chars count.

The reason for suggesting only displaying the chars count if and only if there is room is because Manuskript was designed to help support large written projects such as novels with many plots and characters.  For such projects the word count is of higher importance.

lechbaczynski:
'A different approach might be to only display the chars count if there is sufficient screen space to display"

This might lead to users' confusion, like" where is characters count?  Why cant't I see that? Why my colleague sees car count and I do not? Is there a switch in settings?

But it may also be like that noone will care about that :)

I could try to make it so, if I'll be able to find editor window's size and calculate char and word count + progress bar size (in pixels?) - this might be tricky.

gedakc:
Designing a User Interface is not an easy task.  When making decisions it helps if the project has clear focus backed up by a mission statement.  That way requests can be reviewed in light of the main purpose for the application and decisions made accordingly.  With that said I have not seen a mission statement.  Perhaps that is one more thing that needs to be done to help improve manuskript.

I think that adding a character count has value to manuskript users.  My concern arises when the character count takes precedence over the word count.   This is because I believe the purpose of manuskript is to help create large written works where a text editor or word processor lacks features to help with organizating such endeavours.

TheJackiMonster:
I have merged these changes from this PR to my fork and added settings to it, so the folders and texts in the tree-view can also show the amount of chars instead of words if selected. The amount of chars will only be shown (like in the upper screenshot with the problematic format at the moment) if it's activated in general settings.
But this was more an experiment than a solution.. ^^'
I will try to work on a proper solution which considers all the mentioned points here. Shall I open a new request when I'm done or can I adapt this one somehow? (I would guess not but it would be my second PR, so I really don't know. ^^)

TheJackiMonster:
Okay, I think I have something which should work... the default settings will disable the rendered char-counter next to the word-counter but it will enable a tooltip. So you can hover over the word-counter to see the amount of chars and it's still allowed to activate it for people with enough screen space.
image

Issue #331, opened 2018-03-26
Export crashes, because of encoding to 1250


lechbaczynski:
Windows, version 0.6.0

While exporting to plain text (not pandoc):

  File "manuskript\ui\exporters\exporter.py", line 107, in export
  File "manuskript\exporter\manuskript\plainText.py", line 102, in export
  File "c:\python34\lib\encodings\cp1250.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode characters in position 3968-396
9: character maps to <undefined>

Why doesn't it export to utf 8?

lechbaczynski:
This happened in downloaded compiled windows version of 0.60
This did not happen in current devel , run on windows from sources.

gedakc:
Thanks for testing reporting back that the problem does not exist with the current **develop** branch.

When the next official release is created, the **develop** branch will become the 0.7.0 release.  As such I will close this report as a bug and tag it with the 0.7.0 milestone.

lechbaczynski:
Okay, but anyway, why not use utf8?

Editor of Manuskript - good:
![obraz](https://user-images.githubusercontent.com/1127623/38452267-e9a346ae-3a40-11e8-9b16-a68e0e204927.png)

Exported file - bad (Sublime text, default):
![obraz](https://user-images.githubusercontent.com/1127623/38452270-f2cf6f28-3a40-11e8-902d-4170987834d5.png)

But if  I open it with windows 1250 encoding it looks okay:
Sublime text, opening the file explicitly in 1250:
![obraz](https://user-images.githubusercontent.com/1127623/38452297-81990a2a-3a41-11e8-9514-6e6014481793.png)


Windows Notepad seems to use 1250 by default:
![obraz](https://user-images.githubusercontent.com/1127623/38452291-6e0a97d0-3a41-11e8-9f22-07ae77ebcb5e.png)

lechbaczynski:
I am sorry, it seems that I was mistaken. The error still exist, 
but only if I use some characters that cannot be encoded in windows 1250
I exported whole short story with just Polish letters, and it was okay.
But If I insert anything else:


The editor in Manuskript:

![obraz](https://user-images.githubusercontent.com/1127623/38452316-1a0c8f7a-3a42-11e8-8ace-3bf85b301f2b.png)


The export preview:

![obraz](https://user-images.githubusercontent.com/1127623/38452320-2d55e310-3a42-11e8-8e61-2b9856976388.png)

[...]

PR #340, opened 2018-04-09
- simple fix for #331 (errror when saving in cp1250)

lechbaczynski:


lechbaczynski:
Hi, 

I tested savings in a few formats, and was okay, I did not test PDF as i did not have that option available.

So, on my machine / locale settings it works.

It might be worth testing on other machines (Linux etc) and other locales

gedakc:
'Just looking at this bug fix again.  Soon I would like to create a **0.7.0** release of Manuskript.

Because I am unsure of the implications of this change to existing Manuskript projects I would like to include this fix after the 0.7.0 release.  That way it will spend some time in the **develop** branch for other users to test it.

@olivierkes, if you know that this bug fix will not adversely affect current Manuskript projects then please let us know and we can consider including it in the upcoming **0.7.0** release.

gedakc:
Thank you @lechbaczynski for this PR to fix the compile/export crash and for your patience.

I hope you don't mind that I made some minor changes to the commit message to fix a spelling mistake and to improve the readability when this fix is included in the CHANGELOG.md file.

Thanks again for this fix.

Sincerely,
Curtis Gedak

Issue #379, opened 2018-08-23
[FEATURE IDEA] Incorporate GIT in Manuskript


TeebObeeT:
As a user of Manuskript, I want to version my project data and be able to get "back in time" if needed.

The best option is to use Git (Lab|Hub) . Similarly as the function from modern IDE (Atom, VS Code), a possibility to commit (with message) and push directly from Manuskript would be a plus.

Optionnally, possibility to clone and display diff with previous version but this might be an overkill

RiderExMachina:
I really like this idea and will see what I can do.

obw:
A Versioning Interface with a kind of PlugIn API would be better on the long run, I personally use SVN for all my personal Project (On my own server, Programming and Writing)...
When only one Person use a repository, SVN is more easy to admin then GIT...

edomaur:
I agree with @obw , a plugin API would be better : I'd like to add support for Mercurial (instead of git) but it's not the only idea I'd like to implement as plugins.

olinox14:
a good way to achieve this would be to change the format of the project file (.msk file) to a format like xml. That way you could use the versionning system you want on the file(s) itself

gedakc:
It is possible to manually use a version control system on the Manuskript project files when the **Edit -> Settings** value for **Save to one single file** is **disabled** as described in [Open and Plain Text Format](http://www.theologeek.ch/manuskript/2016/03/31/open-plain-text-file-format/).

QuantumGrizzly:
I +1 the feature request of providing a built-in tracking of revisions within Manuskript. Since we already have the folder-saving feature, we can already use git or another Source Code Management (SCM) solution outside the software. But it would be nice to manage the entire lifecycle of the project in one software.

I also like @obw's suggestion of using a plugin architecture, so while most people would probably want to use Git, it still makes it easy to implement SCM systems like SVN or Mercurial.

TheJackiMonster:
I have also started a discussion ( #871 ) to this topic for the refactoring process. I'm not fully sure yet if we should use GIT or SVN but I think we will definitely replace the old revision system with one of these.

LoveIsGrief:
As a middle-ground, maybe the way forward really is just letting another version control system handle revisions. What could be done is to change the revisions setting to something that allows one to select a version control system. What that does is simply call another executable with the save location as the current working directory upon certain triggers e.g

 - `git` is selected
 - An init button is enabled --> `git init`
 - Save: open terminal window with `git commit`
[...]

PR #401, opened 2018-10-23
Don't crash when smart remove is set to zero

### LINKED ISSUE #381: Lowering number of saved revisions below 1 crashes program
* Manuskript 0.7.0
* Windows 10 Pro

Under Settings -> Revisions: Lowering any of the counters below "1" causes an immediate crash.

katafrakt:
This is my idea to fix #381 – just use 0 when instead of trying to divide by zero. I tested it and it works, however it's definitely not perfect: `__safeDivide` method should not be part of `settingsWindow`. Can you suggest what would be a better place to put it?

### DIFF
--- a/manuskript/settingsWindow.py
+++ b/manuskript/settingsWindow.py
@@ -127,15 +127,15 @@ class settingsWindow(QWidget, Ui_Settings):
         self.chkRevisionsKeep.stateChanged.connect(self.revisionsSettingsChanged)
         self.chkRevisionRemove.setChecked(opt["smartremove"])
         self.chkRevisionRemove.toggled.connect(self.revisionsSettingsChanged)
-        self.spnRevisions10Mn.setValue(60 / opt["rules"][10 * 60])
+        self.spnRevisions10Mn.setValue(self.__safeDivide(60, opt["rules"][10 * 60]))
         self.spnRevisions10Mn.valueChanged.connect(self.revisionsSettingsChanged)
-        self.spnRevisionsHour.setValue(60 * 10 / opt["rules"][60 * 60])
+        self.spnRevisionsHour.setValue(self.__safeDivide(60 * 10, opt["rules"][60 * 60]))
         self.spnRevisionsHour.valueChanged.connect(self.revisionsSettingsChanged)
-        self.spnRevisionsDay.setValue(60 * 60 / opt["rules"][60 * 60 * 24])
+        self.spnRevisionsDay.setValue(self.__safeDivide(60 * 60, opt["rules"][60 * 60 * 24]))
         self.spnRevisionsDay.valueChanged.connect(self.revisionsSettingsChanged)
-        self.spnRevisionsMonth.setValue(60 * 60 * 24 / opt["rules"][60 * 60 * 24 * 30])
+        self.spnRevisionsMonth.setValue(self.__safeDivide(60 * 60 * 24, opt["rules"][60 * 60 * 24 * 30]))
         self.spnRevisionsMonth.valueChanged.connect(self.revisionsSettingsChanged)
-        self.spnRevisionsEternity.setValue(60 * 60 * 24 * 7 / opt["rules"][None])
+        self.spnRevisionsEternity.setValue(self.__safeDivide(60 * 60 * 24 * 7, opt["rules"][None]))
         self.spnRevisionsEternity.valueChanged.connect(self.revisionsSettingsChanged)
 
         # Views
@@ -301,6 +301,13 @@ class settingsWindow(QWidget, Ui_Settings):
             self.lstMenu.setCurrentRow(tabs[tab])
         else:
             self.lstMenu.setCurrentRow(tab)
+
+    def __safeDivide(self, dividend, divisor):
+        if divisor == 0:
+            return 0
+        else:
+            return dividend / divisor
+
     ####################################################################################################
     #                                           GENERAL                                                #
     ####################################################################################################
@@ -357,11 +364,11 @@ class settingsWindow(QWidget, Ui_Settings):
         opt = settings.revisions
         opt["keep"] = True if self.chkRevisionsKeep.checkState() else False
         opt["smartremove"] = self.chkRevisionRemove.isChecked()
-        opt["rules"][10 * 60] = 60 / self.spnRevisions10Mn.value()
-        opt["rules"][60 * 60] = 60 * 10 / self.spnRevisionsHour.value()
-        opt["rules"][60 * 60 * 24] = 60 * 60 / self.spnRevisionsDay.value()
-        opt["rules"][60 * 60 * 24 * 30] = 60 * 60 * 24 / self.spnRevisionsMonth.value()
-        opt["rules"][None] = 60 * 60 * 24 * 7 / self.spnRevisionsEternity.value()
+        opt["rules"][10 * 60] = self.__safeDivide(60, self.spnRevisions10Mn.value())
+        opt["rules"][60 * 60] = self.__safeDivide(60 * 10, self.spnRevisionsHour.value())
+        opt["rules"][60 * 60 * 24] = self.__safeDivide(60 * 60, self.spnRevisionsDay.value())
+        opt["rules"][60 * 60 * 24 * 30] = self.__safeDivide(60 * 60 * 24, self.spnRevisionsMonth.value())
+        opt["rules"][None] = self.__safeDivide(60 * 60 * 24 * 7, self.spnRevisionsEternity.value())
 
     ####################################################################################################
     #                                           VIEWS                                                  #


katafrakt:
There's a segmentation fault in Semaphore CI. Any idea what's this about?

```
manuskript/tests/test_functions.py::test_wordCount PASSED                [  7%]
manuskript/tests/test_functions.py::test_convert PASSED                  [ 15%]
manuskript/tests/test_functions.py::test_several PASSED                  [ 23%]
manuskript/tests/test_functions.py::test_outlineItemColors PASSED        [ 30%]
manuskript/tests/test_functions.py::test_paths PASSED                    [ 38%]
manuskript/tests/test_functions.py::test_mainWindow PASSED               [ 46%]
manuskript/tests/test_settingsWindow.py::test_general PASSED             [ 53%]
manuskript/tests/models/test_outlineItem.py::test_outlineItemsProperties PASSED [ 61%]
manuskript/tests/models/test_outlineItem.py::test_modelStuff PASSED      [ 69%]
manuskript/tests/models/test_references.py::test_references PASSED       [ 76%]
manuskript/tests/ui/test_welcome.py::test_autoLoad PASSED                [ 84%]
manuskript/tests/ui/exporters/test_exporters.py::test_loadExportWiget PASSED [ 92%]
manuskript/tests/ui/importers/test_importers.py::test_loadImportWiget Segmentation fault (core dumped)
```

gedakc:
It is good to hear from you again @katafrakt.  :-)

Thanks for looking at this issue.  I don't know for certain but I suspect that by letting the initial "set revision to zero" code proceed, then this results in a crash later on.  It's just a guess though because I don't know exactly what is going on with the automated testing.

On the topic of the patch, without having a solid understanding of what is going on with this revision saving code, my thought is that it would be safer to prevent users from setting the value to zero.  Perhaps even limiting the upper end value as well.

I think implementing the restriction in the GUI would make it clear to a writer that the value cannot be set to zero.   If a writer does not wish revisions saved then the feature can be turned off by disabling **Keep revisions**.

Thinking more about this I wonder how many writers are concerned about saving revisions.  In my opinion I think revision control is more important to developers of code than writers of novels.

RiderExMachina:
I too think that "0" should mean no saved revisions. I think this implementation is better than the one I wrote; however, I'm unsure of how to set that within Manuskript itself. Good work, @katafrakt on this current revision.

katafrakt:
I was reading the code related to smart revisions and I think I more or less understand it. I'm going to try to write some test cases for it, so we don't have to walk through it again in the future. My Python skill is a bit rusty though, so it's not going as fast as I wished.

gedakc:
That's great @katafrakt that you are looking further into this.  No rush on my part.  My python skills are rusty too so everything seems to take much longer than expected.

Thank you for helping with Manuskript.  :-)

gedakc:
First, my intention is not to rush or put pressure on investigating the implication of setting revision values to zero.  It is better to have a good understanding of how this works, and not rush it into a production release.

Second, For Your Information:  I'm contemplating creating a new Manuskript release in early December.

worstje:
Just looking through open PRs since 0.10.0 is on the horizon... is there anything that is stopping this from being merged as-is? I just tested on my end and the division-by-zero crash is still a thing.

I don't like the way revisions are implemented at all (very laggy and not very useful), but I think with the length of time this bug as well as the fix has lingered we should at least put a band-aid on at this point. It might not be perfect, but revisions as a whole aren't perfect right now anyways.

katafrakt:
@worstje I believe the idea was to create a thorough test suite for revisions. Unfortunately it exceeded my time resources or skills (also ability to understand how revisions should work, really), so I'm afraid I have to give up on this task.

gedakc:
Has anyone tested or investigated to see what happens when revisions are set to zero?

I previously tested and confirmed that this PR will prevent the crash from occurring, but did not have the time to investigate further.

worstje:
I am just eyeballing the logic I see in the code because that is the best I can do right now. So YMMV. (Or is it MMMV? Hmm. 😬 )

The entered value is stored in terms of a fraction second-long time spans:
- `rules[10*60=600]` is 'revisions per minute for the last 10 minutes'.
- `rules[60*60=3600]` is 'revisions per 10 minutes for the last hour'
- And so forth.
Three revisions per minute for the last 10 minutes is stored as 60/3 = 20 seconds, which is essentially a 'time-between-revisions'.

These defined timespans are then used in `OutlineItem.cleanRevisions()`.

It has two separate logic loops in which it determines whether or not to keep a revision:
 * `for r in rev` identifies the smallest timespan the revision fits into.
    *This one does not look at the configured revisions, only the timespan.*
 * `for span in revs` checks whether or not the revision meets the rule configured.

https://github.com/olivierkes/manuskript/blob/0943d813172091f77ec13751971bf2ea2c9264ff/manuskript/models/outlineItem.py#L449-L458

This is where the logic appears to go wrong for our purposes.

It sorts every revision it found by its age. Then it loops through all of them. If nothing has been stored yet, it stores it. If something has been stored before, deduces the time from the last revision stored to see if it is more than or equals the configured revisions for that entry.

**Ergo, it always stores at least one revision per configurable timespan.** This logic (lines 453-455) unconditionally stores the oldest revision encountered, and then checks subsequent revisions against the configured time-between-revisions (lines 456-458).

I think a fix for this would have to make sure that the first loop (`for r in rev`) skips a configured timespan if that timespan equals 0. That way, that revision will be taken into account for the next tier of consideration. Inserting a check between lines 

https://github.com/olivierkes/manuskript/blob/0943d813172091f77ec13751971bf2ea2c9264ff/manuskript/models/outlineItem.py#L442-L447

Between lines 444 and 445, insert something like:

`if rule[span] == 0:
    continue`

Note that I am ignoring the 'not span' check below it, that would need fine-tuning. Hell, I am not sure if the logic as a whole works out; my sleepy mind is hurting from all the time units and the way the variable names don't really talk to me as much as I'd like. I am not sure either if a 10 year old revision would eventually get deleted, or whether you'd eventually end up with revisions that never get deleted...

But hopefully this bit of analysis helps out in understanding the code and fixing up the last few issues with this PR. 😄

gedakc:
Thank you @katafrakt and @worstje for looking into this revisions crash issue.

Last night I was mulling over how to fix the revisions feature in Manuskript when I came to an epiphany.  Why are we spending time to fix revisions when I don't even know what purpose these serve?  Personally I have never used this feature, nor has my spouse.

In my opinion the revisions feature is consuming valuable time both from developers trying to maintain it, and also from users in terms of [reduced performance](https://github.com/olivierkes/manuskript/issues/523).

What is the purpose of the current revisions feature?

Who uses the revisions feature and how do they use it?

As a programmer I recognize the usefulness of storing revisions at a known state and time, but would never consider having my code changes saved on an automatic schedule.

At the moment I think we should consider setting this feature to default to **off** and marking it deprecated for the 0.10.0 release.  For the next release we might consider removing the revisions code entirely.

What are your thoughts?

worstje:
I am in full agreement.

I don't like the feature (no big surprise there I take it, given how often I think I've spoken about it) and I think that whatever role it fulfills right now can be covered more accurately by using some sort of source control in conjunction.

I know that some people want to see better revisions (#640), but that at the same time people really like the way it works now (#639).

I think deprecation is the right way to go for 0.10.0, and to draw up a proper design document that covers exactly what kind of feature-set we are looking to cover with the revision feature.

Right now it is a crappy form of manual backups, but revisions are not a Manuskript-unique feature, and it means a lot more in other software like Word.

gedakc:
> I am in full agreement.

'Sounds like we are on the same page.

Would you like to create a PR to turn off the revisions feature by default and note somewhere in the interface that it is deprecated?

Then we can quit wasting time on the current revisions feature.

For the next release we could remove the current revision feature entirely so that we have a clean slate to start with.

Word processors often have a **track change** feature that enables collaboration on a document and is what I think #640 is looking for.  #639 is looking for revisions to be manually taken, which is more like something a source code revision control system does (see issue #379).

katafrakt:
Great thought, @gedakc! Removing overly complex code and have a fresh start if sometimes better that forcibly trying to maintain it. If someone complains about revisions gone, removing PR could be reverted and we could come back to this discussion or make a high priority to bring it back in an ordered manner.

gedakc:
I plan to create a PR to be included in the 0.10.0 release which disables the Revisions setting **Keep revisions** by default for all new projects.

Before undertaking this task I performed a very brief test of the proposed commit in this PR.

- I set "Keep **0** revisions per minute for the last 10 minutes".  
- I confirmed that making this change no longer crashed Manuskript.  
- I proceeded to type in the **Editor** pane.
- I noticed while typing that a new revision was saved at least _**every second**_!

katafrakt:
Ok, then let's maybe close this PR as this is both wrong and unnecessary anymore ;)

gedakc:
'Agreed.  Closing this PR.

Issue #404, opened 2018-10-31
Plots bounce around main, secondary, and minor -- unsatisfactory solution?


jordanwagenet:
So I'm loving Manuskript, but I've encountered an annoying issue: my plots are bouncing between main, secondary, and minor seemingly at random when clicked on. 

Context: I'm in Windows 10 x64, running the current Manuskript version. My msk file is pretty fleshed out. I have my first act in the editor, I have placeholders for world building, characters, plots, chapters, etc. It all seemed to be working fine until I deleted a major plot. Soon after, the other plots started bouncing around between "main," "secondary," and "minor" in the plot sidebar, sometimes being selected/highlighted in large blocks across the plot headers. Though, I think every click is cycling them through until all of them are under the minor category. The only other change I was making was adding detailed info entries for my main character.

Reassigning plot hierarchies, restarting, and recreating some of the plots doesn't work. They still cycle through until all of them are under minor. Deleting all but the main plot didn't work either. Nothing else is misbehaving. Even the example project's plot function is working fine, though I haven't deleted a plot because it's my test standard and I don't want to fudge it up in case I need to compare something.

An apparent solution just now: Deleting and recreating all of my plots seems to fix the problem. However, this isn't a satisfactory solution.

Any advice on why this was happening?

gedakc:
Thank you @jordanwagenet for reporting this issue and including the steps you took to work-around the issue.  We're glad that you enjoy using this software.

It would help if you let us know the exact version of manuskript (this is listed under the **Help -> About** menu).  Because manuskript is available for download from multiple locations it is quite possible that the "_current_" version is not the same version everywhere.

If you are up to doing some sleuth work, you might be able to discern the cause of the issue yourself.

Manuskript stores the project data in an [open and plain-text format](http://www.theologeek.ch/manuskript/2016/03/31/open-plain-text-file-format/).  If you disable the **Save to one single file** setting, then the _project.msk_ file will be expanded into a _project_ directory.  You can peruse this directory with a file manager and a text editor to learn the structure of plots, scenes, etc.

My suspicion is that there may be a problem with either the links between plots, chars, scenes, etc., or a missing unique number to identify each plot, char, etc.

By spending some time looking at the plain text files you may be able to discern the cause of the issue.

jordanwagenet:
[...]

PR #406, opened 2018-11-05
Fix plot importance changes if delete earlier plot and click other plots

gedakc:
The root cause was a mismatch between plot IDs and plot model rows.

This issue would appear when a plot was deleted such that the plot IDs
did not match the plot model row numbers and different plots had
different importance levels.  The problem would not occur if the most
recently added plot was deleted.

For test steps see issue #404.

If someone has a chance to test this patch then I would greatly appreciate it.

I will hold off on merging this patch for a few days to wait for feedback.


deansgit:
Works for me. No more plots bouncing around. Thanks!

gedakc:
Thank you @deansgit for testing this patch and confirming that it fixes the problemm of plots bouncing around.  This fix is planned for inclusion in the next release of Manuskript.

Issue #428, opened 2018-12-05
Chinese translation filename suffix


lingsamuel:
![image](https://user-images.githubusercontent.com/14567045/49506952-5ea0c480-f8ba-11e8-8e07-082594ce496c.png)
I noticed that manukript loads the translation `manuskript_zh_CN`, but the current translation filename is `manuskript _zh`.
Maybe that's because there are two type of Chinese, Chinese(Simplified) which is translated in Weblate called `zh_CN` and Chinese(Traditional) called `zh_TW` (or `zh_HK`). In operating system, they will use code like `zh_CN` or `zh_TW` instead of just `zh`.

gedakc:
I'm not overly familiar with how Weblate works, but perhaps the base translation file was created with only the `_zh` suffix?

From a quick look at the Manuskript source code it appears that there are two Portuguese translations: [`_pt_BR`](https://hosted.weblate.org/projects/manuskript/translations/pt_BR/), and [`pt_PT`](https://hosted.weblate.org/projects/manuskript/translations/pt_PT/) so it seems it should be possible to have both `_zh_CN` and `_zh_TW`.

Can you create a new translation using the proper suffixes raised in this issue?

lingsamuel:
Created: #430

PR #430, opened 2018-12-07
translation suffix, change translation load order

lingsamuel:
Changes:
- Rename Chinese translation suffix from `zh` to `zh_CN`, there is not traditional (`zh_TW`) translation.
- Change the translation load order. When I am testing new translation suffix, I found that `zh` is recorded in my settings, but the debug log shows failed to load `zh_CN` while the actually loaded file is `zh`. I think load translation from settings should be the first choice, and then fallback to system locale if failed. And I fixed the wrong debug information by the way.


gedakc:
Thank you @lingsamuel for this enhancement, for using the **develop** branch, and for creating a descriptive commit title.

In my testing this patch worked correctly for languages such as Spanish and Russian.  However I did not observe Chinese characters when I selected "Chinese (Simplified)" and then restarted Manuskript.  Note that this might be because I don't have a "Chinese (Simplified)" locale installed on my kubuntu 16.04 computer, and is not likely a fault of this patch.

GitHub supports automated closing of issues/PRs.  See [Closing issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/).

Would you be able to edit the commit message for this patch (`git commit --amend`) and add a reference to the issue this fixes (issue #428).  After that you'll need `--force` to re-push the patch.

This creates a link between the issues and pull requests will help with future tracking / troubleshooting.

Thanks,
Curtis

lingsamuel:
You can't see Chinese because the `.ts` file only contains a little translations but it's marked 100% translated in Weblate, I don't know why.
I am working on it right now.

gedakc:
Thank you @lingsamuel for updating the commit message to be descriptive and include the issue number.

I will rebase and merge this request to be included in the next release of Manuskript.

PR #431, opened 2018-12-07
Remember last accessed directory

lingsamuel:
Maybe this can be included in settings page.


gedakc:
Thank you @lingsamuel for this improvement.

I think that remembering the last accessed directory should be the default behaviour for Manuskript.  No need to add this as option to the settings window.

In my testing this enhancement worked by remembering the directory for the open, create, and save as.  As such I will commit this to the **develop** branch.

Thanks again for this enhancement.

PR #437, opened 2018-12-12
Spelling: Program settings, Restart ensures settings

comradekingu:


gedakc:
Please provide your logic and reasoning for the words / sentences changed in this PR.

For example, why change **Application** to **Program**?

Further some changes to manuskript settings will take effect immediately (like font size), whereas others require manuskript to be restarted to take effect (like language).  In my opinion the suggested new wording seems to imply that manuskript should always be restarted.

comradekingu:
@gedakc Program is just my preference. Application more-so tied with "app", which suggests to me it would be a mobile program.

If the user is missing any recently edited setting, the other change is enough to convey the message.

gedakc:
Closing this pull request as the changes were dealt with in PR #454.  Also marking this PR as **invalid** so that it does not get included in the automatic generation of CHANGELOG.md.

PR #454, opened 2019-01-06
Spelling: Manuscript, may have to be restarted

comradekingu:


gedakc:
Hi @comradekingu,

I am in favour of shortening this sentence as you have done in this PR.  Often shorter sentences are easier to translate to other languages

Would you be able to rewrite your suggested change to use present tense?

Technical documentation is generally written in present tense.  For example use "restart" instead of "restarted".

Thanks,
Curtis

gedakc:
Thank you @comradekingu for the rewording into present tense.

We may be able to further improve on the wording.

I think some of the key points @olivierkes was trying to make with the original sentence were:

- Some settings require a restart of Manuskript to take effect
- The restart would have to be performed by you (the user) and would not occur automatically

With these points in mind, what do you think of the following?

`Some settings need you to restart Manuskript to take effect.`

comradekingu:
`You might need to restart manuskript in order for those settings to take effect properly and entirely.`
>You might need to restart

a restart (of?) is in question. 
>manuskript

_Manuscript_
>in order for those

still very unspecified, encompassing all
>settings

OK, so specified settings already mentioned, as opposed to others (?) in question, redundant.
> to take effect

OK
>properly and entirely.

Cancel that, not OK, what qualifies what? Has it now taken effect improperly? If not, why include the word "properly"?…

---

`Restarting Manuscript ensures all settings take effect.`
>Restarting

action, still could be infinite tense.
>Manuskript

Subject, still could be infinite tense.
> ensures

Qualifying statement
>all settings

scope
>take effect.

action.

---

`Some settings need you to restart Manuskript to take effect.`
>Some settings

not "of the", could be any settings
>need you to

has this happened, is this out of the ordinary. It seems still a command.
Don't see much of a point breaking the fourth wall here, the user knows who the user is, its role, and if capable of understanding what to do through other means, then all is good.
>restart

This is where Windows-users might be quick on the draw?
>Manuskript

We got almost to the end of the garden sentence, still none the wiser as to why anything needs doing.
> to take effect.

Aha.

Greater wordcount, and length, needless complexion in sentence structure, word repetition.
Somewhat easier words to understand for people with poor command of English, but then again it is to be translated 

---

- Some settings require a restart of Manuskript to take effect
Hence the "all" `Restarting Manuscript ensures all settings take effect.` In my head its presence communicates something over leaving it out.
- The restart would have to be performed by you (the user) and would not occur automatically
Short of
`To restart Manuscript now ensures all settings take effect henceforth.`
Which hammers in an non-existant nail,
I think if you were to read it like that, it ought to have been
"Restarting Manuscript ensur_ing_ all settings take effect."
As of right now, I see it as a statement. That could be an error or oversight on my part.

A (sadly) common thing to do is "Tip: …". Also, the context string feature can be used to communicate things to translators. It can be changed directly with admin privileges to Manuscript on Hosted Weblate.

I just tried to get the spelling of the program right. In this diminishing size of fish volleys, the bigger fish remain unfried.

PR #460, opened 2019-01-22
Session target

nephlm:
Adds a Tools menu Item that creates a non-modal window that tracks session and draft progress and goals.

screenshot at 2019-01-21 23 09 45

The data is updated every two seconds via a timer.

While I was doing this I noticed that the Frequency Counter window does not close when the main app closed.  Fixed that at the same time, since I was using it as a model. 

This is my first Qt development so there are likely things I did wrong in in regard to the framework so there are likely things I did wrong or at least sub-optimally.  If you point them out, I'll be happy to fix them.


athrvvvv (approved):

athrvvvv (approved):

nephlm:
I wrote this like 4 years ago, but I merged in current develop and resolved the conflicts.  Did some testing and it seems to still work without clobbering anything else.  It's got an approval and is now up to date, I'm not sure who has merge power, but I think it can be merged.

TheJackiMonster:
I've adjusted some parts to prevent crashes in projects with no text files or one empty ones. I'm not even convinced this was an issue on your side but rather the defaults of the data structure we use being empty strings (which doesn't make any sense for word counts for example).

I've also centered the dialog on startup like the others. It can still be moved freely and updates while writing. So I think this can be merged.

Issue #449, opened 2018-12-27
add a requirements.txt file


olinox14:
i would suggest to add a requirements.txt file

* it makes the requirements/dependencies more obvious
* contributors would just have to do `pip install -r requirements.txt` instead of installing manually each library

gedakc:
If I recall correctly there were challenges with the PyQt5 requirement that were not resolvable when I tried to build package files for Manuskript using a `setup.py` file.

See [Package Manuskript for Linux with Stdeb](https://github.com/olivierkes/manuskript/wiki/Package-Manuskript-for-Linux-with-Stdeb) **Note: Does Not work**.

nephlm:
I think is meant for those who want to contribute some code rather than for packaging and distribution.  I was able to install the requirements using pip and a file, at least I think I was, but I was trying to discover those requirements at the same time so by the time I did the final `pip install ...` run it might have already head PyQt5 installed.

gedakc:
For those wishing to contribute some code (and hence run from source code) there are already instructions on how to [Run Manuskript from Source Code](https://github.com/olivierkes/manuskript/wiki#install-source-code-more-work-runs-on-more-os-platforms) and these include the requirements.

nephlm:
Yeah, I found those and they were quite helpful, but as a developer used to working in a python virtualenv I tend to try not to use OS packages to install python packages unless I have to, for all the reasons that virtual envs are a thing.  

I'm not really arguing for this in any strong way, it's no trouble to maintain my own requirements.txt file, just clarifying why I might choose to do that over use the specified OS packages.

If it's something we want, I have a file that works at least in my environment and I can create a pull request.  If you quite reasonably don't want to maintain that and so the PR wouldn't be accepted then maybe we should close this ticket.

gedakc:
Please feel free to contribute a `requirements.txt` file.

I suggest testing the installation of PyQt5 via the `requirements.txt` file to see if it will work as I had no success when I tried creating a `setup.py` file as mentioned earlier.

skle46:
I tried adding a requirements.txt, but I ran into a problem that gedakc notes would probably happen earlier on. Please make changes to my file in the PR if you think something could be done to improve it.

gedakc:
Closing this issue and labeling it as invalid as was suggested in the PR as a reason why we do not use a requirements.txt file.

PR #462, opened 2019-01-27
add a requirement file example

skle46:
I created a requirements file as requested by issue #449.

This file does not work, as python-qt5 returns this error after execution of `pip3 -r install requirements.txt`

`Files/directories not found in /tmp/pip-install-[path]/python-qt5/pip-egg-info`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+python-qt5
+lxml
+webkit
+pyenchant
+markdown


gedakc:
Thank you @skle46 for testing your PR and confirming the issue with the PyQt5 requirement.

If it is not possible to get a `requirements.txt` file working fully with the requirements for Manuskript, then we should not merge this PR.

skle46:
Makes sense to me. Please use this PR as a reason why we do not use requirements.txt, otherwise someone may make a change to this PR and it would help other projects looking to add requirements to their qt5 projects.

gedakc:
Closing this PR and labeling it as **invalid** as suggested as a reason why we do not use a requirements.txt file.

PR #483, opened 2019-02-09
Fix crash when more than 6 pound signs are written alone on a line. 

kakaroto:
Easy to reproduce, enter 7 pound signs on a line with no trailing text and as soon as the 7th is entered, manuskript would crash.

The code would look for trailing pound signs and would count all the way to the beginning and beyond, resulting in an out of bounds exception.
Fix was to make sure not to include the starting pound signs when counting the trailing ones.

### DIFF
--- a/manuskript/ui/highlighters/markdownTokenizer.py
+++ b/manuskript/ui/highlighters/markdownTokenizer.py
@@ -279,7 +279,9 @@ class MarkdownTokenizer(HighlightTokenizer):
 
         if level > 0 and level < len(text):
             # Count how many pound signs are at the end of the text.
-            while escapedText[-trailingPoundCount -1] == "#":
+            # Ignore starting pound signs when calculing trailing signs
+            while level + trailingPoundCount < len(text) and \
+                    escapedText[-trailingPoundCount -1] == "#":
                 trailingPoundCount += 1
 
             token = Token()


gedakc:
Thank you @kakaroto for identifying this issue and providing a fix.

I have confirmed the crash by typing more than 6 pound signs in a row in the editor pane.

The fix looks good to me so I will merge this PR.

gedakc:
'Just found a typo in the comment:  "calculing" shoudl be "calculating".

Would you be able to fix this typo?

kakaroto:
Thanks! I've fixed the typo, and I have two other commits that fix other crashes. I'll just close this PR and send a new one with all 3 patches, all made in their own branch this time :)

gedakc:
If each patch fixes a different crash then it would be useful for each one to have it's own branch and PR.

That will make it more obvious what is fixed for each Manusrkipt release when we automatically generate the [`CHANGELOG.md`](https://github.com/olivierkes/manuskript/blob/develop/CHANGELOG.md) file.

kakaroto:
Ah ok, good point. I just grouped them as "misc crashes", I've now made 3 individual branches and I'm opening new PRs for them. Thanks!

gedakc:
Labelling as **invalid** because another PR will supercede this one.

PR #484, opened 2019-02-09
Fix crash when 7 pound signs are written alone on a line.

kakaroto:
Re-submitting PR #483.

Easy to reproduce, enter 7 pound signs on a line with no trailing text and as soon as the 7th is entered, manuskript would crash.

The code would look for trailing pound signs and would count all the way to the beginning and beyond, resulting in an out of bounds exception.
Fix was to make sure not to include the starting pound signs when counting the trailing ones.


gedakc:
Thanks for creating a separate branch for each PR, and also for fixing the typo.

I have confirmed the crash by typing more than 6 pound signs in a row in the editor pane.

I have also confirmed that the code changes in this PR do indeed avoid the crash.

I will merge this PR.

PR #485, opened 2019-02-09
Fix crash when previewing pandoc HTML with QTextEdit as web renderer…

kakaroto:
After switching from pre-built binary to development tree, QTextEdit became the web rendering engine and the pandoc HTML preview button would make it crash.

### DIFF
--- a/manuskript/exporter/pandoc/HTML.py
+++ b/manuskript/exporter/pandoc/HTML.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # --!-- coding: utf8 --!--
-from PyQt5.QtWidgets import qApp
+from PyQt5.QtWidgets import qApp, QTextEdit
 from PyQt5.QtCore import QUrl
 
 from manuskript.exporter.manuskript import HTML as MskHTML
@@ -40,4 +40,8 @@ class HTML(abstractPlainText):
         previewWidget.widget(0).setPlainText(src)
         self.preparesTextEditView(previewWidget.widget(1), settings["Preview"]["PreviewFont"])
         previewWidget.widget(1).setPlainText(html)
-        previewWidget.widget(2).setHtml(html, QUrl.fromLocalFile(path))
+        w2 = previewWidget.widget(2)
+        if isinstance(w2, QTextEdit):
+            w2.setHtml(html)
+        else:
+            w2.setHtml(html, QUrl.fromLocalFile(path))


gedakc:
I would like to test this PR, but I do not know which OS falls back to using the **QtextEdit** web rendering engine.

Which Operating System and version did you discover this issue on?

How did you install Manuskript on said computer (I'm curious if some Qt packages were omitted)?

kakaroto:
I (shamefully) am using Windows on this computer. I had just installed python 3 then did a `pip install pyqt5` then lxml then pyenchant. There shouldn't be anything else python3 related on this computer. I launch it by using the full path to the 3.7 python.exe with bin/manuskript as argument.

If you look at the manuskript HTML exporter, you'll see it's the same code that I copied into pandoc's HTML exporter: https://github.com/olivierkes/manuskript/blob/develop/manuskript/exporter/manuskript/HTML.py#L82

```
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 22:20:52) [MSC v.1916 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt5.QtWebKitWidgets import QWebView
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'PyQt5.QtWebKitWidgets'
>>> from PyQt5 import QtWebEngineWidgets
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'QtWebEngineWidgets' from 'PyQt5' (c:\Users\kakaroto\AppData\Local\Programs\Python\Python37-32\lib\site-packages\PyQt5\__init__.py)
```

I think pip is at fault, there is no QtWebEngineWidgets in their build and QWebView is apparently deprecated and only available in custom builds of pyqt (note that your windows prebuilt binaries use QWebView). 

Unrelated note: I realized that the huge slowdowns (as much as 30+ second freezing when I right click sometimes) that I had before are gone with the local python install, it seems to have been only present when using the bundled manuskript.exe from the windows binary.

gedakc:
Thank you @kakaroto for the explanation and link to previous code.  I will try to configure a VM to properly test this.

Re:  Unrelated note.  Please feel free to contribute to the discussion regarding the right-click slowdown in a recently reported issue.  Note that I didn't mention the issue number because I didn't want these two issues linked together because these two issues are unrelated.

kakaroto:
You're welcome.

Extra information. It seems that webengine is not included in the 32 bits version of pyqt only. And I can confirm that I had mistakenly installed 32bit python (see [here](https://stackoverflow.com/questions/51900007/pyqt-5-11-missing-webengine-modules)). 
I just tried python 3.7 64 bits to see if it fixed the problem, but I have two issues :
* pyenchant refuses to install with pip (online it says 'use 32 bit python')
* It still says QTextEdit is the chosen web engine and I can confirm even the 64 bit pip wheel is missing the qt web engine module

gedakc:
Thanks for the additional explanation.  I was also ran into the same issues with 64-bit Python and 32-bit PyEnchant.

To help others I added the following [Appendix A: Install Required Software Packages on Windows 7 and Higher](https://github.com/olivierkes/manuskript/wiki/Run-Manuskript-from-Source-Code-on-Windows#appendix-a--install-required-software-packages-on-windows-7-and-higher).

gedakc:
Thank you @kakaroto for identifying this crash with compile / export Pandoc preview HTML and providing code to fix the problem.

I confirmed the crash on Windows 7 (_Web rendering engine used: QTextEdit_) using the Manuskript install instructions linked to earlier in this PR.
I also confirmed that the code update does indeed fix the problem and prevent the crash.

I have regression tested on Kubuntu 16.04 (Web rendering engine used: QWebView) and Fedora 28 (Web rendering engine used: QWebEngineView) and both showed the preview correctly.  To use QWebEngineView on Fedora 28 I removed `python3-qt5-webkit` and installed `python3-qt-webengine`.

This PR looks good to me.  I will merge it into the codebase.

PR #486, opened 2019-02-09
Don't crash if a typo is made in the exporter's regular expression.

kakaroto:
This took me a while to figure out a week ago when I tried manuskript for the first time, I didn't understand why the exporter had stopped working until I opened a console to check the log and see the re.error exception.
This fixes that so if a regexp is invalid, it will show a warning window and cancel the export instead of crashing.

### DIFF
--- a/manuskript/exporter/manuskript/plainText.py
+++ b/manuskript/exporter/manuskript/plainText.py
@@ -2,7 +2,7 @@
 # --!-- coding: utf8 --!--
 import re
 from PyQt5.QtGui import QFont, QTextCharFormat
-from PyQt5.QtWidgets import QPlainTextEdit, qApp, QFrame, QFileDialog
+from PyQt5.QtWidgets import QPlainTextEdit, qApp, QFrame, QFileDialog, QMessageBox
 
 from manuskript.exporter.basic import basicFormat
 from manuskript.functions import mainWindow
@@ -41,7 +41,12 @@ class plainText(basicFormat):
 
     def output(self, settingsWidget):
         settings = settingsWidget.getSettings()
-        return self.concatenate(mainWindow().mdlOutline.rootItem, settings)
+        try:
+            return self.concatenate(mainWindow().mdlOutline.rootItem, settings)
+        except re.error as e:
+            QMessageBox.warning(mainWindow().dialog, qApp.translate("Export", "Error"),
+                                qApp.translate("Export", "Error processing regular expression : \n{}").format(str(e)))
+            return ""
 
     def getExportFilename(self, settingsWidget, varName=None, filter=None):
 


manuskript/exporter/manuskript/plainText.py
```diff
@@ -43,9 +43,9 @@ def output(self, settingsWidget):
         settings = settingsWidget.getSettings()
         try:
             return self.concatenate(mainWindow().mdlOutline.rootItem, settings)
-        except re.error:
+        except re.error as e:
             QMessageBox.warning(mainWindow().dialog, qApp.translate("Export", "Error"),
-                                qApp.translate("Export", "Error processing regular expression"))
+                                qApp.translate("Export", "Error processing regular expression : \n%s" % str(e)))
```
kakaroto: Wonder if it should be this instead : 
`      qApp.translate("Export", "Error processing regular expression : \n{}").format(str(e)))`

gedakc: Yes, I agree that it would be better to include the actual regexp error string but not translate it.  Only the static text contained in the Manuskript application should be translated.

kakaroto: Ha! That was obvious but hadn't realized the error string would get translated too. I never worked with translations before so didn't know that would be the behaviour (makes sense now that I think about it).
I'm busy today, so feel free to change it and merge if you want, otherwise, I'll apply the change tomorrow.

gedakc: I don't think the regexp error string will get translated in either situation.  However I do prefer the latter change thought because it more closely matches the existing code.  For example see https://github.com/olivierkes/manuskript/blob/9b774d327f8ef83c729f875667a074988b366f8d/manuskript/models/references.py#L495

Also an update tomorrow will be fine.  If I make the change and push it then it breaks some of the automatic tracking and accreditation that github performs.  Hence I prefer for the original author to make the change, get the credit, and have github mark the PR as merged.

kakaroto:
I fixed it, tested and amended it. Let me know if you prefer I squash the 2 commits into one.
Also, what's up with the failed semaphoreci check? Is that something from your side or from github?

gedakc:
@kakaroto, yes please do squash the two commits into one.  That will make it easier when we troubleshoot issues in the future.

The semaphoreci check is something that @olivierkes has configured in his git repo.  If I recall it tries to build .deb files using the latest code, and also runs some basic tests.  Unfortunately semaphoreci seems to be only accessible from the git repo owner's account so only @olivierkes is able to make changes.

We also use Travis CI to test code and build the macOS X packages.  Fortunately these are controlled by the [`.travis.yml`](https://github.com/olivierkes/manuskript/blob/develop/.travis.yml) file so everyone can view and contribute to it's maintenance.

kakaroto:
Ok, thanks for the info.
I've squashed the commits. Hope it's all good.

gedakc:
Thank you @kakaroto for this updated patch.

I tested with the following broken regexp and was able to confirm that unpatched manuskript crashes, and that this PR prevents the crash:

`<TAG\b[^>*>(.*?)</TAG>`

Note that the properly formed regexp is:

 `<TAG\b[^>]*>(.*?)</TAG>`

I will merge this PR for inclusion in the next release of Manuskript.

PR #492, opened 2019-02-10
Fix crash when right clicking a word in editor and enchant is not installed

kakaroto:
Error was : 
```
  File "v:\Projects\book\manuskript-git\bin\..\manuskript\ui\views\textEditView.py", line 475, in createStandardContextMenu
    valid = self._dict.check(text)
AttributeError: 'NoneType' object has no attribute 'check'
```

Disabling the entire spellcheck section by checking if self._dict is None does the trick.

### DIFF
--- a/manuskript/ui/views/textEditView.py
+++ b/manuskript/ui/views/textEditView.py
@@ -470,7 +470,7 @@ class textEditView(QTextEdit):
 
         # Check if the selected word is misspelled and offer spelling
         # suggestions if it is.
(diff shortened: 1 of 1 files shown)


gedakc:
Thank you @kakaroto for this crash fix.

Using Fedora 28 with the `python3-enchant` packaged removed I confirmed the crash.
I also confirmed that the code in this PR does indeed prevent the crash.

I will merge this PR with the codebase.

PR #503, opened 2019-02-21
Fix fullscreen editor's myScrollBar delayed destruction causing a crash

kakaroto:
This can lead to a crash if a timer was started right before deletion, so that when
the timer is launched, the self.parent() crashes.

```
  File "v:\Projects\book\manuskript-git\bin\..\manuskript\ui\editors\fullScreenEditor.py", line 314, in <lambda>
    self.timer.timeout.connect(lambda: self.parent().hideWidget(self))
RuntimeError: Wrapped C/C++ object of type myScrollbar has been deleted
```

I've had this error happen to me multiple times after pressing F11 once as soon as I start the app, but I've been unable to reproduce it after adding debug messages (and also after removing them), so it looks like a heisenbug, but this should fix the problem, even though I can't test that it works, it seems logical. I'm also unsure on why a scrollbar would get destroyed after I enter fullscreen mode once without exiting it.

### DIFF
--- a/manuskript/ui/editors/fullScreenEditor.py
+++ b/manuskript/ui/editors/fullScreenEditor.py
@@ -312,10 +312,13 @@ class myScrollBar(QScrollBar):
         self.timer = QTimer()
         self.timer.setInterval(500)
         self.timer.setSingleShot(True)
-        self.timer.timeout.connect(lambda: self.parent().hideWidget(self))
+        self.timer.timeout.connect(self.hide)
         self.valueChanged.connect(lambda v: self.timer.start())
         self.valueChanged.connect(lambda: self.parent().showWidget(self))
         self.rangeChanged.connect(self.rangeHasChanged)
+
+    def hide(self):
+        self.parent().hideWidget(self)
 
     def setColor(self, color):
         self._color = color


kakaroto:
I just had the crash again with this fix, as well as with a `self.timer = None` added (I was hoping for a `self.timer.start()` Null pointer exception instead). So this doesn't fix it.
Still investigating but I can now reproduce the issue, and working on a fix. To reproduce, keep pressing F11 repeatedly while scrolling the mouse wheel, it will eventually crash after 3 or 4 times (note, this is on Windows).
I found the issue is that `__del__` isn't called, probably due to the garbage collector not happening when the scrollbar object gets destroyed, so I think the solution now is to just do what I did on a 'destroyed' callback instead of `__del__`. I'm working on it now, then will test, and push an updated commit.

kakaroto:
I figured it out. Sorry for the first patch. You can reproduce by pressing F11 repeatedly fast enough, no need to scroll or anything. The issue was in using a lambda function for the timer. If you add prints in the `__init__` and in a `__del__`, you'll see the myScrollBar doesn't get deleted if a lambda is used (just replace the `connect(self.hide)` with `connect(lambda: self.hide())` to test).
I also used a `sip.isdeleted(self)` to confirm and test multiple times with a lambda but avoiding a crash.

gedakc:
Thank you @kakaroto for explaining how to reproduce the crash, and also providing a patch to fix the issue.

On Kubuntu 16.04 I was able to reproduce the crash by loading a project into manuskript and then pressing **F11** in rapid succession.

I plan to review the Pull Requests in order of creation.  If you would prefer a different order of review then let me know.

I will merge this PR.

Curtis

kakaroto:
Hey @gedakc, I hope you enjoyed your long weekend!
Thanks for the quick review/merge. You can review the PRs in order of creation although it might be best to review bugfixes first and new features later as bugfixes should be easier/faster to review.
In either case, take your time, I know I've spammed the PR list over the weekend and that it might take you some time to review it all.
Thanks!

gedakc:
Good suggestion.  I will review the bug fixes first before moving on to the new features, such as alternate spell checkers.  It may take me a few days as I have other project work too.

Ideally I would like to get your patches reviewed and merged with an eye towards creating a new Manuskript release in early March.

PR #504, opened 2019-02-21
Set editor theme stylesheet to QTextEdit only.

### LINKED ISSUE #440: Full-screen mode right-click menu black text on black background
When I right-click in full-screen mode, the menu that pops up has a black background, with colourful icons. Only once I hover over any of the menu items, does the font change to white and I can actually read the menu item. The rest all have a black font, on a black background. I don't seem to be able to find a place in the settings, where one can change either the background nor foreground colours of this popup menu.

kakaroto:
This prevents any child widget from inheriting the same stylesheet,
more specifically, the context menu of the full screen editor will now
appear normal instead of being black text on black background, which made
it unreadable.

Fixes #440 

**note**: I have tested this and full screen mode looks good, as well as the side panels and the context menu, but I don't know if there are any other potential child widget other than QMenu which would actually need to inherit that theme data.
I have also tried to make the background same as the theme data instead of 'transparent', but it looks bad in the menu and there is no difference in text color between selectable and non selectable menu items (such as copy/paste/undo/redo), so I preferred to opt for a standard menu instead.

### DIFF
--- a/manuskript/ui/editors/themes.py
+++ b/manuskript/ui/editors/themes.py
@@ -249,12 +249,14 @@ def setThemeEditorDatas(editor, themeDatas, pixmap, screenRect):
     # editor.setFont(f)
 
     editor.setStyleSheet("""
-        background: transparent;
-        color: {foreground};
-        font-family: {ff};
-        font-size: {fs};
-        selection-color: {sc};
-        selection-background-color: {sbc};
+        QTextEdit {{
+            background: transparent;
+            color: {foreground};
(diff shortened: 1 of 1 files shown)


gedakc:
Thank you @kakaroto for developing a fix for the issue with the fullscreen right-click menu being unreadable in some themes.

I have tested this fix and confirmed that I was able to read the right-click menu text in fullscreen mode with each of the four themes built-in to Manuskript.

Curtis

PR #506, opened 2019-02-22
Do not default spellcheck to True for new editor views

### LINKED ISSUE #474: Spellcheck On/Off setting ignored / Manuskript unresponsive
Manuskript 0.8.0 Win 10 64 4GB / OS X El Capitan 4GB

When the spellcheck is turned Off in the main menu Manuskript ignores that setting depending on from where you open a new tab. When you open tabs from the search results or the cheatsheet the spellcheck is On no matter what the setting is. Same when you reopen tabs or hold cmd key down.

Text view in the Root tab with the whole text from the sample project selected (13.088 words) it takes over a minute to open with the spellcheck turned On. Manuskript is unresponsive till the text is fully loaded. Again spellcheck is On even though turned Off in the main menu.
Editing is near impossible with a long text and spellcheck turned On. As soon as you scroll the text Manuskript is unresponsive again.

With the spellcheck turned Off evreything works fast and relieable. 
Spellcheck is just always coming back On.

Please also read my post regarding the searchbox where I have mentioned this behaviour:
(https://github.com/olivierkes/manuskript/issues/376#issuecomment-460396349)

kakaroto:
default spellcheck to the settings value instead. Otherwise any new editor tab
will have spellcheck enabled regardless of setting.

Fixes #474

### DIFF
--- a/manuskript/ui/editors/editorWidget.py
+++ b/manuskript/ui/editors/editorWidget.py
@@ -54,7 +54,7 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
         self.dictChanged.connect(self.txtRedacText.setDict, AUC)
         self.txtRedacText.setHighlighting(True)
         self.currentDict = ""
-        self.spellcheck = True
+        self.spellcheck = settings.spellcheck
         self.folderView = "cork"
         self.mw = mainWindow()
         self._tabWidget = None  # set by mainEditor on creation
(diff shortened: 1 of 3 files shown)


gedakc:
Thanks @kakaroto for developing a fix for this problem and adding a reference to issue #474 in the commit message.

Using the test steps from issue #474 I was able to confirm that this patch fixes the issue.

I will rebase and merge this fix with the **develop** branch for inclusion in the next release of Manuskript.

Issue #505, opened 2019-02-21
Add non-enchant spellcheck support


kakaroto:
I'm using manuskript on Windows with python 64 bits and there is no pyenchant support for 64 bits unfortunately. Also it appears pyenchant is not maintained anymore and even the [link](http://pythonhosted.org/pyenchant/) that manuskript opens when enchant isn't installed doesn't work.

My solution is to add support for an alternative spellchecker, and a simple "python spellchecker" search suggests [pyspellchecker](https://pyspellchecker.readthedocs.io/en/latest/index.html). I've started to add support for it and it seems to work fine so far. My patches are not yet ready for review/push, but I'm creating this issue to discuss the feature and my proposed changes before I'm done with it.

First, I wanted to keep pyenchant support, so I'm not replacing it but just adding support for something else. 
If nothing is installed, the menu will suggest installing either pyenchant or pyspellchecker and clicking on it would open both links.
![image](https://user-images.githubusercontent.com/27990/53157716-1a0af800-3590-11e9-989c-d4c827a439cb.png)
If both are available, then the Tools->Dictionary would show dictionaries from both libraries so the user can make their choice of library to use : 
![image](https://user-images.githubusercontent.com/27990/53158269-33f90a80-3591-11e9-8900-b1f6d1c75ef8.png)

If only one is available, mention that another choice is available to the user :
![image](https://user-images.githubusercontent.com/27990/53157604-da441080-358f-11e9-8bdf-6968e4588784.png)

So first question: **Does that UI look good or you prefer it to be done differently?**

I think that's good, but I have one issue: It only works as is because the dictionary names are different from pyspellchecker and pyenchant, I don't know how it would work if pyspellchecker suddenly changed dictionary names to look like pyenchant. 
**So I'm wondering if I should store the library name as part of the 'dict' setting, or not?**
I'm not entirely sure, but I think I can store the library name in the QAction without it appearing as text to the user, I'd have to check, but if not, I don't want to have the library name on every entry...

Right now, the code handles both enchant and pyspellchecker, but I think it would make a lot more sense to create an abstraction layer, a Spellchecker class that does that for us, so the code remains simple within the text editor, highlighter, main window, context menu, etc...
Once I write the abstraction class, it would be easy to add support for other spellchecker libraries, so another question: **Are you OK with adding support for pyspellchecker or do you have another library to suggest?**
[...]

PR #507, opened 2019-02-22
Refactor spellchecker code and add other spellcheck library support

kakaroto:
This is a rather big change, but should be easy to review one commit at a time. It will add an abstraction for the spellchecker then add support for pyspellchecker.
Only 'bug'/issue I have is if I run 32 bit version and select a PyEnchant dictionary, then run the 64 bit version, then spellchecking doesn't work because my settings have stored "PyEnchant:en_US" as the chosen dictionary, so I'd have to reselect the pyspellchecker dictionary for it to work again.. this could happen to a user and they'd be lost as to why it's not working.
Another issue that would need to be discussed is whether we should have the custom dictionary entries stored in the project file or not. It makes sense if you add to dictionary weird character names or fantasy world names/concepts, but if you're just fixing the dictionary to add words you often use across projects, it could be annoying. Anyway, needs to be discussed.

### DIFF
--- a/manuskript/functions/spellchecker.py
+++ b/manuskript/functions/spellchecker.py
@@ -1,0 +1,139 @@
+
+from PyQt5.QtCore import QLocale
+
+try:
+    import enchant
+except ImportError:
+    enchant = None
+
+class Spellchecker:
+    dictionaries = {}
+
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def isInstalled():
+        return enchant is not None
+
+    @staticmethod
+    def availableDictionaries():
+        return EnchantDictionary.availableDictionaries()
+
+    @staticmethod
+    def getDefaultDictionary():
+        return EnchantDictionary.getDefaultDictionary()
+
+    @staticmethod
+    def getLibraryURL():
+        return EnchantDictionary.getLibraryURL()
+
+    @staticmethod
+    def getDictionary(dictionary):
+        if not dictionary:
+            dictionary = Spellchecker.getDefaultDictionary()
+        try:
+            d = Spellchecker.dictionaries.get(dictionary, None)
+            if d is None:
+                d = EnchantDictionary(dictionary)
+                Spellchecker.dictionaries[d.name] = d
+            return d
+        except Exception as e:
+            return None
+
+class BasicDictionary:
+    def __init__(self, name):
+        pass
+
+    @property
+    def name(self):
+        raise NotImplemented
+
+    @staticmethod
+    def getLibraryName():
+        raise NotImplemented
+
+    @staticmethod
+    def getLibraryURL():
+        raise NotImplemented
+
+    @staticmethod
+    def getDefaultDictionary():
+        raise NotImplemented
+
+    @staticmethod
+    def availableDictionaries():
+        raise NotImplemented
+
+    def isMisspelled(self, word):
+        raise NotImplemented
+
+    def getSuggestions(self, word):
+        raise NotImplemented
+
+    def isCustomWord(self, word):
+        raise NotImplemented
+
+    def addWord(self, word):
+        raise NotImplemented
+
+    def removeWord(self, word):
+        raise NotImplemented
+
+class EnchantDictionary(BasicDictionary):
+
+    def __init__(self, name):
+        self._lang = name
+        if not (self._lang and enchant.dict_exists(self._lang)):
+            self._lang = self.getDefaultDictionary()
+
+        self._dict = enchant.Dict(self._lang)
+
+    @property
+    def name(self):
+        return self._lang
+
+    @staticmethod
+    def getLibraryName():
+        return "pyEnchant"
+
+    @staticmethod
+    def getLibraryURL():
+        return "https://pypi.org/project/pyenchant/"
+
+    @staticmethod
+    def availableDictionaries():
+        if enchant:
+            return list(map(lambda i: str(i[0]), enchant.list_dicts()))
+        return []
+
+    @staticmethod
+    def getDefaultDictionary():
+        if not enchant:
+            return None
+
+        default_locale = enchant.get_default_language()
+        if default_locale and not enchant.dict_exists(default_locale):
+            default_locale = None
+
+        if default_locale is None:
+            default_locale = QLocale.system().name()
+        if default_locale is None:
+            default_locale = self.availableDictionaries()[0]
+
+        return default_locale
+
+    def isMisspelled(self, word):
+        return not self._dict.check(word)
+
+    def getSuggestions(self, word):
+        return self._dict.suggest(word)
+
+    def isCustomWord(self, word):
+        return self._dict.is_added(word)
+
+    def addWord(self, word):
+        self._dict.add(word)
+
+    def removeWord(self, word):
+        self._dict.remove(word)
--- a/manuskript/functions/spellchecker.py
+++ b/manuskript/functions/spellchecker.py
@@ -1,5 +1,5 @@
 
-import os, gzip, json
+import os, gzip, json, glob
 from PyQt5.QtCore import QLocale
 from collections import OrderedDict
 from manuskript.functions import writablePath
@@ -22,6 +22,10 @@ class Spellchecker:
 
     def __init__(self):
         pass
+
+    @staticmethod
+    def registerImplementation(impl):
+        Spellchecker.implementations.append(impl)
 
     @staticmethod
     def isInstalled():
@@ -73,14 +77,6 @@ class Spellchecker:
             return urls.get(lib, None)
         return urls
 
-    
-    @staticmethod
-    def getResourcesPath(library):
-        path = os.path.join(writablePath(), "resources", "dictionaries", library)
-        if not os.path.exists(path):
-            os.makedirs(path)
-        return path
-
     @staticmethod
     def getDictionary(dictionary):
         if not dictionary:
@@ -98,7 +94,7 @@ class Spellchecker:
             d = Spellchecker.dictionaries.get(dictionary, None)
             if d is None:
                 for impl in Spellchecker.implementations:
-                    if lib == impl.getLibraryName():
+                    if impl.isInstalled() and lib == impl.getLibraryName():
                         d = impl(name)
                         Spellchecker.dictionaries[dictionary] = d
                         break
@@ -109,11 +105,24 @@ class Spellchecker:
 
 class BasicDictionary:
     def __init__(self, name):
-        pass
+        self._lang = name
+        if not self._lang:
+            self._lang = self.getDefaultDictionary()
+
+        self._customDict = set()
+        customPath = self.getCustomDictionaryPath()
+        try:
+            with gzip.open(customPath, "rt", encoding='utf-8') as f:
+                self._customDict = set(json.loads(f.read()))
+                for word in self._customDict:
+                    self._dict.create_dictionary_entry(word, self.CUSTOM_COUNT)
+        except:
+            # If error loading the file, overwrite with empty dictionary
+            self._saveCustomDict()
 
     @property
     def name(self):
-        raise NotImplemented
+        return self._lang
 
     @staticmethod
     def getLibraryName():
@@ -142,13 +151,35 @@ class BasicDictionary:
         raise NotImplemented
 
     def isCustomWord(self, word):
-        raise NotImplemented
+        return word.lower() in self._customDict
 
     def addWord(self, word):
-        raise NotImplemented
+        word = word.lower()
+        if not word in self._customDict:
+            self._customDict.add(word)
+            self._saveCustomDict()
 
     def removeWord(self, word):
-        raise NotImplemented
+        word = word.lower()
+        if word in self._customDict:
+            self._customDict.remove(word)
+            self._saveCustomDict()
+
+    @classmethod
+    def getResourcesPath(cls):
+        path = os.path.join(writablePath(), "resources", "dictionaries", cls.getLibraryName())
+        if not os.path.exists(path):
+            os.makedirs(path)
+        return path
+
+    def getCustomDictionaryPath(self):
+        return os.path.join(self.getResourcesPath(), "{}.json.gz".format(self._lang))
+
+    def _saveCustomDict(self):
+        customPath = self.getCustomDictionaryPath()
+        with gzip.open(customPath, "wt") as f:
+            f.write(json.dumps(list(self._customDict)))
+
 
 class EnchantDictionary(BasicDictionary):
 
@@ -157,11 +188,7 @@ class EnchantDictionary(BasicDictionary):
         if not (self._lang and enchant.dict_exists(self._lang)):
             self._lang = self.getDefaultDictionary()
 
-        self._dict = enchant.Dict(self._lang)
-
-    @property
-    def name(self):
-        return self._lang
+        self._dict = enchant.DictWithPWL(self._lang, self.getCustomDictionaryPath())
 
     @staticmethod
     def getLibraryName():
@@ -211,36 +238,19 @@ class EnchantDictionary(BasicDictionary):
 
     def removeWord(self, word):
         self._dict.remove(word)
-Spellchecker.implementations.append(EnchantDictionary)
 
+    def getCustomDictionaryPath(self):
+        return os.path.join(self.getResourcesPath(), "{}.txt".format(self.name))
+
+Spellchecker.implementations.append(EnchantDictionary)
 
 class PySpellcheckerDictionary(BasicDictionary):
 
     def __init__(self, name):
-        self._lang = name
-        if not self._lang:
-            self._lang = self.getDefaultDictionary()
+        BasicDictionary.__init__(self, name)
 
-        self._dict = pyspellchecker.SpellChecker(self._lang)
-        self._customDict = None
-        customPath = self.getCustomDictionaryPath()
-        try:
-            self._customDict = pyspellchecker.SpellChecker(local_dictionary=customPath)
-            self._dict.word_frequency.load_dictionary(customPath)
-        except:
-            # If error loading the file, overwrite with empty dictionary
-            with gzip.open(customPath, "wt") as f:
-                f.write(json.dumps({}))
-                
-        self._customDict = pyspellchecker.SpellChecker(local_dictionary=customPath)
-        self._dict.word_frequency.load_dictionary(customPath)
-
-    def getCustomDictionaryPath(self):
-        return os.path.join(Spellchecker.getResourcesPath(self.getLibraryName()), "{}.json.gz".format(self._lang))
-
-    @property
-    def name(self):
-        return self._lang
+        self._dict = pyspellchecker.SpellChecker(self.name)
+        self._dict.word_frequency.load_words(self._customDict)
 
     @staticmethod
     def getLibraryName():
@@ -257,9 +267,11 @@ class PySpellcheckerDictionary(BasicDictionary):
     @staticmethod
     def availableDictionaries():
         if PySpellcheckerDictionary.isInstalled():
-            # TODO: If pyspellchecker eventually adds a way to get this list
-            # programmatically or if the list changes, we need to update it here
-            return ["de", "en", "es", "fr", "pt"]
+            dictionaries = []
+            files = glob.glob(os.path.join(pyspellchecker.__path__[0], "resources", "*.json.gz"))
+            for file in files:
+                dictionaries.append(os.path.basename(file)[:-8])
+            return dictionaries
         return []
 
     @staticmethod
@@ -284,17 +296,14 @@ class PySpellcheckerDictionary(BasicDictionary):
             candidates.remove(word)
         return candidates
 
-    def isCustomWord(self, word):
-        return len(self._customDict.known([word])) > 0
-
     def addWord(self, word):
-        self._dict.word_frequency.add(word)
-        self._customDict.word_frequency.add(word)
-        self._customDict.export(self.getCustomDictionaryPath(), gzipped=True)
+        BasicDictionary.addWord(self, word)
+        self._dict.word_frequency.add(word.lower())
 
     def removeWord(self, word):
-        self._dict.word_frequency.remove(word)
-        self._customDict.word_frequency.remove(word)
-        self._customDict.export(self.getCustomDictionaryPath(), gzipped=True)
+        BasicDictionary.removeWord(self, word)
+        self._dict.word_frequency.remove(word.lower())
 
-Spellchecker.implementations.append(PySpellcheckerDictionary)
+Spellchecker.registerImplementation(PySpellcheckerDictionary)
+
+
--- a/manuskript/functions/spellchecker.py
+++ b/manuskript/functions/spellchecker.py
@@ -14,6 +14,11 @@ try:
 except ImportError:
     pyspellchecker = None
 
+try:
+    import symspellpy
+except ImportError:
+    symspellpy = None
+
 
 class Spellchecker:
     dictionaries = {}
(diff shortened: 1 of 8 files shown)


manuskript/functions/spellchecker.py
```diff
@@ -1,45 +1,111 @@
...
+            if impl.isInstalled():
+                return True
+        return False
+
+    @staticmethod
+    def supportedLibraries():
+        libs = []
+        for impl in Spellchecker.implementations:
+            libs.append(impl.getLibraryName())
+        return libs
```
kakaroto: Humm.. more python way to do it : 
`return [i.getLibraryName() for i in Spellchecker.implementations]`

worstje:
What still need to be done to make this one ready to merge in? Is there anything I can do to help speed things along?

Of all the pending pull requests, this is the one I would love to see merged the most. 😄

gedakc:
Basically it comes down to lack of available free time to both review and test this PR.

Because this is a significant change it will require testing on multiple platforms.  It would help if someone can come up with a series of steps to test with either and both spell checkers installed, and then run these tests on as many platforms as possible.  The platforms we support Linux, Windows, and Mac OS X, in both 32-bit and 64-bit versions if available.

kakaroto:
I've been using manuskript with symspellpy for weeks now without issues. I've done plenty of testing on windows with 32 and 64 bits, as well as on 64 bit linux, though it's been a month so it wouldn't hurt to retest it all and maybe someone else would find a use case I haven't thought of or tested. 

I recently pushed a PR to symspellpy that improves loading time by 40% and changes the cached dictionary format (https://github.com/mammothb/symspellpy/issues/31#issuecomment-467533508) so I need to adapt the code here to take that into account (if it fails to load the cached dict, then reload the dictionary from the txt file instead of loading an empty dict). I'll try to do that tomorrow and push a commit with the fix.

After that I suppose I'll just need to go through another round of testing to make sure everything is still working as it should. 

Steps to test (don't know if that's what you wanted) : 
* Open with no supported library installed, verify that the Tools menu shows the list of libraries supported and suggests installing them and clicking on each opens the right website link.
* Open with at least one supported library installed, but one or more missing libraries. Verify that Tools->Dictionary menu shows dictionary list from the right libraries and that the missing libraries are marked as missing.
* Verify that spellchecking works
  * Enabling spellcheck gets misspelled words to be underlined
  * Disabling spellcheck removes underlining from misspelled words
* Change dictionaries and verify that spellchecking works
  * If changing dictionary while spellcheck is disabled, it does not enable it
  * If changing dictionary while spellcheck is disabled, enabling it shows the misspelled words from the newly selected dictionary
  * If changing dictionary while spellcheck is enabled, the old misspelled words stop being underlined and the new misspelled words get underlined.
* Change dictionaries from one library to another one and verify that it behaves the same as if the dictionary selection was from the same library
* Verify that misspelled words get the right expected suggestions
  * Suggestions may vary from one library to another, that's ok
  * The misspelled word itself shouldn't appear in the suggestions list
  * If no suggestions would be available, the context menu has no 'spelling suggestions' submenu
  * Clicking on a suggestion from the menu changing the word correctly and the word stops being underlined
* Verify support for custom words
  * Verify that custom words that get added to the dictionary stop being underlined everywhere in the text
  * Verify that right-clicking a custom word will show the "Remove from custom dictionary" option
  * Verify that removing a custom word from the dictionary gets the word underlined everywhere in the text
  * Verify that custom words remain after manuskript is restarted
  * Verify that custom words that were removed from the custom dictionary do not get re-added after manuskript is restarted
* Verify that the right dictionary in the right library (even same dictionary name) is selected when opening Manuskript or switching to a different project that used a different dictionary.
* Verify that the default dictionary gets selected if a project is opened which had its dictionary set to one that becomes unavailable (uninstalled the library the dictionary was in for example)

gedakc:
@kakaroto I think your list of test steps is exactly what is needed.  You've covered many different cases which are important to test when making major changes to functionality.

If I understand correctly there are still some enhancements that you would like to make to this PR before it receives a final review before committing to the **develop** branch.  If this is not the case then let me know.

kakaroto:
Cool! 
Yes, there are a few things that I want done before I push. I'm trying to do it now but I might not have time to finish it tonight so I'll probably push the final change tomorrow.

Things I need fixed : 
* Change the save/load of the cached dictionary to be unzipped instead of gzipped (makes it 15 times faster to save, and 33% faster to load) **DONE**
* Check if the load fails by returning False (file format is incompatible) instead of relying only on an exception being raised by the load method **DONE**
* Use the new delete_dictionary_entry method that was added in this week's release, to allow removal of custom words from the dictionary instead of reloading the dictionary from disk (which takes a second to load).
  * Add check to make sure the symspellpy module version is at least >= 6.3.8 using distutils.version functions to compare versions
  * Add a way to return to the main window the library requirement so the menu can display to the user "symspellpy >= 6.3.8 not installed" instead of just "symspellpy not installed" to avoid potential confusion to the users.

kakaroto:
I've done all the changes I wanted, and did all the tests I mentioned above (and found a new use case which I added to the list) and here are the issues I found so far, testing on 32 bit and 64 bit windows) : 

- Opening Manuskript where the dictionary is not available (switching from 32 bit to 64 bit where not all the same libraries are available for example) does not make it select the default dictionary. It just remains that no dictionary is selected until the user makes a choice in the Tools menu. Enabling/disabling spellcheck has basically no effect and the user doesn't have a visual feedback to tell them to select a new dictionary.
- Same as above but after the user selects a new dictionary, spellchecking does not work until the user disables spellchecking and re-enables it.
- If no spelling suggestions are found, there is no "Add to dictionary" option in the popup menu (was there before it seems).

I unfortunately don't have much time to be fixing those. I'll try tomorrow but no guarantees.

gedakc:
Thanks @kakaroto for the update.  I will await further updates before reviewing this PR.

kakaroto:
@gedakc It was easier to fix than I thought. Fixes are pushed, feel free to test as well and let me know if you find any new issues on the other platforms.
Btw, you can use instructions from here on how to create dictionaries for symspellpy : https://github.com/olivierkes/manuskript/issues/505#issuecomment-466632839

kakaroto:
Found a new "bug", if you're in the fullscreen editor, then alt-tab to the main window and change the dictionary, it will change it in all editors of the main window but the fullscreen one will keep using the old dictionary. I don't know if this is meant to be fixed since you're not really supposed to alt-tab out of full screen mode... I won't have time to fix this though, so please fix it (probably call the setDict on self._fullscreen if it exists from the mainwindow class) or file an issue so it can be fixed later.

gedakc:
@kakaroto I've had a quick look through the code, but haven't done extensive testing.

# PR Suggestions for Improvement:

1.  In the first commit entitled "Add a Spellchecker abstraction layer and clean up code" I think it important to mention "Windows" in the commit message such as:

    _...PyEnchant which seems to be unmaintained and does not build on **Windows** 64 bit._

    The reason I say this is that to my knowledge, Linux does have 64 bit packages.

2.  As you indicated earlier there is an invalid URL in `manuskript/functions/spellchecker.py` for `http://pythonhosted.org/pyenchant/`.

    Perhaps we should use `https://pypi.org/project/pyenchant/`?

3.  I noticed that the locale code appears to be removed from `manuskript/ui/views/textEditView.py`.

    What is the reason for this removal?  
    Does this mean that a spell check language will not be automatically selected?

4. In the fifth commit entitled "spellchecker: Improve support for symspellpy" I have a suggested change in the text from "must" to "might".

    `File must be corrupted or incompatible with installed symspellpy version`.

    Unless we know for 100% sure what the issue is, I suggest changing "must" to "might".


# Suggested Path Forward

It is unfortunate that [PyEnhant now has a status of unmaintained](https://github.com/rfk/pyenchant/).  Fortunately you have stepped up to address this problem.  :-)

For this PR I would like to have a longer period of testing, but would also like to create a new release of manuskript (hopefully with your full screen enhancements -- I will add some comments there soon).

With that in mind I would like to hold off committing this PR to the **develop** branch until after a new version of Manuskript is created and released.  Then we can look at getting this merged.

Please let me know if there are reasons why path forward might not be a good idea.

kakaroto:
Hi @gedakc thanks for the suggestions. Here are my answers:

1. Good point, I'll ammend that commit and push.
2. I was hoping the pythonhosted issue was temporary, it doesn't seem to be. I'll update the url to pypi instead.
3. Nope, the code wasn't removed, it was just moved into `spellcheck.py`. It's now specific to the library, the default dictionary/locale is checked in the `getDefaultDictionary` static method which is called in case a valid dictionary isn't set. See https://github.com/olivierkes/manuskript/blob/8d8c1a09c0f3e68de6ee676184abab96430c66a3/manuskript/functions/spellchecker.py#L227
If I remember correctly, I had checked everywhere that function was used and made sure all the use cases are still covered. However, I just realized the `getDefaultDictionary` of pyspellchecker and symspellpy would need to get updated (check if the locale is an available dictionary, if not, return the first available one)
4. Sure, I can change it to 'might' since it makes more sense.

As for the path forward, I have no objections. It does touch a rather important feature and refactors it in a significant way, so a longer period of testing is a good call. 
I already use it locally, so for me it makes no real difference :)

gedakc:
Thanks you @kakaroto for the prompt response, and for explaining where the dictionary locale code migrated.  I had missed that when reviewing the code.

I'll look through some of your other PRs with an eye towards including them in the next Manuskript release.

kakaroto:
All done, let me know if there's anything else needed once you get a chance to review it properly.

gedakc:
My initial regression testing of this PR on Kubuntu 16.04 has gone well.  At this time I have only one minor suggestion:

@kakaroto would you consider adding one or both of the following two lines to the top of each new file you created for this PR?

    #!/usr/bin/env python
    # --!-- coding: utf8 --!--

These lines appear in many of the other `.py` files in the Manuskript project.  
See also:

[StackOverflow: Why do people write the #!/usr/bin/env python shebang on the first line of a Python script?](https://stackoverflow.com/questions/2429511/why-do-people-write-the-usr-bin-env-python-shebang-on-the-first-line-of-a-pyt)
[StackOverflow: Correct way to define Python source code encoding](https://stackoverflow.com/questions/728891/correct-way-to-define-python-source-code-encoding)
[StackOverflow: Is '# -*- coding: utf-8 -*-' also a comment in Python?](https://stackoverflow.com/questions/41680533/is-coding-utf-8-also-a-comment-in-python)

@worstje since you expressed interest in helping get this PR included, would you be able to test this PR on Windows 10?

worstje:
@gedakc Sure, no problem! You two were so active after I initially posted, which combined with me not having a concrete way to contribute and life picking up pushed this PR on the backburner in my mind.

My first look at it immediately throws up a huge question in my head that I should have brought way earlier. Note that I am doing my test on Win10 with a 64-bit python, which means that I didn't have pyenchant installed yet. You might say I am coming into this with the insights of someone who has done their spellchecking with eyeballing and copypasting in other tools. :)

Clicking the 'Tools' menu shows me three different 'please install this library' options. Why like this? It gives me a very overwhelming and not-user-friendly feeling.

Don't get me wrong: I understand we want what is needed in order to use spellchecking to be easily discoverable. However, if I install one of the libraries, am I always going to be faced with 'please install this other library' whenever I open that menu with a very 'loud' warning triangle that demands my attention? All I want to do is to spellcheck, not install libraries!

I'm going to be installing the two new spellcheck libraries after this and do some proper testing of the spellchecking itself, but if we are refactoring to this degree (reading the PR title), should we not consider 'spellchecking' a feature of its own and encapsulate the libraries used as a program-configurable setting? Ideally, I'd just hit a shortcut (which I don't see right now, presumably because I haven't got the libraries installed) to 'open up the spellchecker' or 'start spellchecking', and not be faced with a choice of tools to install. PyEnchant? Symspellpy? Pyspellchecker? Or something that unifies the options and uses all of them at once behind the scenes?

If this were to be about some sort of 'find/search' functionality, no program would have separate menu items for searching plain text, searching with the PCRE engine, .NET regexes or whatever other flavors are available. They just have the option to 'find', and the rest is encapsulated away. ('Replace' would often be its own menu item, or it could be combined as 'Find/replace'.) So to me, seeing three different menu options squashed together which essentially accomplish the same purpose seems like overkill.

From a users perspective, I would probably recommend giving the Settings screen a 'Spellchecker' category which allows the configuration of spellchecking functionality in general. Which spellchecking engine to use, or a combination thereof, is something most users will only want to adjust once or twice even if they have a dozen different projects. To have any 'X is not installed' messages put in there is a to me logical step. Of course, I haven't got the foggiest how dictionaries / language are handled right now, but I could see those relying on a project default with the opportunity to easily override where necessary. You don't want every visit to the spellchecker to turn into a 'huh, which library should I use?!' problem; users just want to spellcheck. And if it isn't good enough, they'll find their way to the settings to fine-tune Manuskript's spellchecking facilities to their needs.

If we leave it as-is, I would at the very least recommend putting a separator between the Frequency Analyzer and the items related to spellchecking, just to show the last three entries of the tool menu have a shared purpose of sorts.

(I realize most of these thoughts come rather late in the process; my apologies for that.)

I will post once again once I test the actual functionality in more detail.

kakaroto:
@gedakc sorry I forgot to add the file header.
I'm currently out of the country and won't be home for another 2 weeks and my laptop just died on me yesterday. Would you mind making that change yourself and committing it if you merge before then?

@worstje: those warnings shouldn't be there for 99% of people because the dependencies would be prebundled/preinstalled with Manuskript.. Just like you don't see a "install Python " message when you run manuskript the first time. But in the case where none of the libraries are available (such as running from source) then it shows the list of dependencies needed for spellcheck. 

> However, if I install one of the libraries, am I always going to be faced with 'please install this other library' whenever I open that menu with a very 'loud' warning triangle that demands my attention

As long as one of them is installed, the others wouldn't appear, so no, you wouldn't see the warning on the menu all the time.
I suggest you give it a try and see how it works.

> All I want to do is to spellcheck, not install libraries!

All I want to do is write my manuscript, not install applications to help me do it. :-P seriously though, you do it once (or never since it should be an auto installed dependency once this gets merged) then that's it.

kakaroto:
Stupid phone posted the comment before i was done. Anyways, the other stuff you talked about I think show you misunderstood how the functionality actually works so I'd suggest you just test out the spellchecker and see if you have comments then.

I don't know if the git repo has packaging info (a setup.py file maybe?) but the new libs would probably need to be added to the dep list.

worstje:
@kakaroto I should probably have held off commenting until I got to test the whole thing, sorry! (I unfortunately suffer that Windows-only pip problem where installing a single package can take several minutes or upto an hour due to some certificate-related stuff that hits a timeout condition for every downloaded file. I figured I'd just comment regarding the parts I could actually see and comment on...)

Regarding spellchecking after I installed the symspellpy and pyspellchecker, I see several things that caught my attention. (Again, caveat here being I haven't really played with the spellchecker before, so it is perhaps not purely about these changes.)

### The dictionaries for symspellpy and pyspellchecker are in a different order.

symspellpy: en, de, es, fr, pt
pyspellcheck: de, en, es, fr, pt

I am not sure if symspellpy has some special condition that forces english to be the \#1 option or whether it is a coincidence, but it would be nice to have both lists list the dictionaries in alphabetical order. I realize not all languages will be available for both libraries, but I was trying to figure out whether the different engine would mark different words, and I accidentally selected the German dictionary on a few occasions. Consistency is nice to have.

(I suspect they use the same dictionaries based on the few tests I've done so far with about a dozen texts since I can't spot any differences between the marked words when I toggle them, but then again, if the dictionaries differ that much, I'd probably be worried about the quality of their spellchecking, too...)

### Spellcheck menu item position is weird.

After installing packages, the Spellcheck option is at the top of the Tools menu, whereas the dictionaries are at a submenu at the bottom of the Tools menu. It squeezes the 'Frequency Analyzer' inbetween. I am not sure how the Frequency Analyzer works internally, but for as far I have been able to observe, it is completely separate from these libraries. In its current iteration, it should be either above or below, but not in between.

### Do we even want this 'Spellcheck' in the Tools menu?

Emphasis in the above question being on both 'this' and 'Spellcheck'. Allow me to explain.

The 'Spellcheck' item (SC for short) in its current format toggles whether or not red squiggles are shown. It is very different from the 'Frequency Analyzer' item (FA for short) which pops up a window, and a user who has impeccable language might think this feature is broken because nothing happens. (Unless they pay attention really well and see the icon being turned on/off). Looking at other menus, one can argue that FA should instead be marked as 'Frequency Analyzer...' to emphasize it pops up a dialog box. (And I will argue that, even if it isn't a part of this PR.)

However, I still question the positioning of the SC menu item. At its core, what it does is something I'd expect to see in the 'View' menu with a text like 'show misspelled words' that makes it crystal clear as to what the option does.

But that does not mean I do not want a 'Spellcheck' item in the Tools menu! But rather than what it does now, I would like for a dialog box to pop up (similar to a traditional Find or Replace dialog) which allows me to easily process the entire text I am currently working with. ([Plenty of examples can be found.](https://www.google.com/search?q=spellcheck+dialog&tbm=isch&sa=X))

This would make it feasible to do the spellchecking separately from the creative process. If I want to check my entire text for misspellings (and I'm going to want to do this a lot now that these alternative libraries exist!), such a dialog would make this a lot easier. I don't have to 'hunt' for the red squiggles in my texts, I won't have to right-click and hover over 'Spelling suggestions' to see what the proper spelling might be, etc.

It would not replace an in-depth read for all-around editing and QC to catch missing or incorrect words, but it would allow the latter to focus less on spelling and more on the writing.

(I realize such a dialog might be out-of-scope for this PR, but it sort-of naturally followed after running into the above items regarding the menu items...)


### Menu structure is awkward when right-clicking a word in the editor

This one kind of came up while I was writing out the previous bit. Nothing is wrong with it; it is functional... but is it ideal? I personally feel that sticking the spelling suggestions in a submenu is not productive. ([Look at the way other software tackles this feature by putting these options at the top of the context menu.](https://www.google.com/search?q=spellcheck+context+menu&tbm=isch&sa=X))

Having these options immediately visible is very convenient. Just hit the popup key while writing and you are immediately presented with spelling suggestions for the word your cursor is on. (It even works while I am typing this in Chrome!)

Looking over some of the screenshots I spotted while googling just now, I'd like to suggest the following structure when rightclicking a 'wrod': (sorry if the text is huge; I have no clue why GitHub does this...)

> &1. **word** (in bold because it is the most likely choice that you could ideally just hit 'enter' for to apply)
> &2. wood
> &3. rod
> &4. trod
> &5. prod
> &More suggestions -> (only when more than 5 suggestions exist)
> ----- brod
> ----- wro
> ----- woe
> -
> Add 'WORD' to dictionary
> -
> Undo
> Redo
> -
> Cut
> Copy
> Paste
> Delete
> -
> Select All

Note that I added the word in question to the 'add to dictionary' item to make this clearer. This is probably more personal preference since I myself consider it less confusing, but since it does highlight the word when the menu pops up it is not essential.

I put the numbers in front with accelerators to make it more convenient for the keyboard warriors among us. Suppose I meant to use 'trod', then I can simply bash the '4' key rather than 'down, down, down, down, enter'. Perhaps forcefully using accelerators like this isn't visually pleasing, so maybe it would be more feasible to have them aligned on the right side as shortcuts. (Or you could use images of the numbers to press on the left which would look nice given that all other menu items have images in that column, but it wouldn't be very accessible...)

(While we're at it, we should probably also check what accelerators make sense or are defaults so that all the other non-spellcheck items can be properly decorated... but that's completely out-of-scope for this PR)

## Sorry.

I figured I'd have to apologize with a heading because I realize I've gone a bit beyond just checking if spellchecking works on Windows 10 with all this. Short answer: it works. Long answer: spellchecking could be way more convenient.

Some of the things mentions may be out-of-scope for this PR, and I mostly mention them because you know all this spellcheck-related code at this point. I could write a patch for the menu-related things I mentioned so others don't have to put their effort into it, but I definitely do not know enough about the libraries nor Qt to write that sort of spellcheck dialog. (Which would be the largest out-of-scope item I can imagine of all the things I wrote.)

kakaroto:
Thanks foe the feedback!
That's quite a list of improvements and I think you're right, most of those items you listed would be really good UX improvements foe the spellcheck feature, but as you also mentioned, it's out of scope for this PR (which was to just add alternative library support to the existing feature with minimal changes).

I would suggest these ideas be compiled in their own issue for SC feedback and UX improvements and I think when I have time, I'll try to tackle those if everyone is in agreement that they improve thr user experience. For my part, I like the idea of a spellcheck dialog to look for all spellchecking mistakes and the idea of moving top suggestions to the menu itself, I'm not sure i agree/care about FA being in the middle between SC and dictionaries or that it's in the Tools menu rather than View, and I would disagree with the accelerators for suggestions, it's not really necessary, barely helps and would be an eye sore. 

As for the issue related to this item, the order of the dictionaries, they are actually in the same order, it's just.. Complicated. To be more precise, symspellpy doesn't come with any dictionaries included, so if you hadn't installed pyspellcheck at the same time then the menu would have said "no dictionary available for symspellpy" instead of "symspellpy not installed". What I did to avoid that is that if pyspellcheck is also included then I also return the list from pyspellcheck (exact same list returned as is, so exact same order) and if you select one of those dicts, then it will convert it to symspellpy format and put it in the cache directory. 
What happened in your case is that you probably selected 'de' by mistake the first time which caused it to import the german dictionary into symspellpy cache directory and then subsequently, it showed 'de' first because it will always list the cached dictionaries before showing the ones it can import from pyspellcheck (from which it removed 'de' sinxe you already have it "natively"). 
Note that the pyspellcheck dictionaries are not that great so I'm actually using the enchant dictionary with symspellpy and in one of my comments in the issue related to this PR, i have a set of instructions on how to download and convert the enchant dictionary into the symspellpy format. That will go into the wiki after this is merged, but having that done for you in a more convenient way (spellcheck section in the settings dialog where the user can import dictionaries or delete cached ones, see all words added to the custom dictionary and manage those as well, etc..)  is also something that i want to add and is out of context for this PR. 
You seem good at UI and UX design so if you could chime in on how you think that could be done in the most user friendly way, that would be great. 

Ps: any mistakes in this comment are due to me using my phone... I swear, i don't write this badly usually...

gedakc:
> Re: file header.  
> I'm currently out of the country and won't be home for another 2 weeks and my laptop just died on me yesterday. Would you mind making that change yourself and committing it if you merge before then?

No rush from my end.  I hope you enjoy your time out of the country.  When you're back we can make and apply enhancements.  In the meantime I will do some more testing.

On the topic of spellcheck improvements I agree that we should probably make the bigger UI enhancements as separate issues/PRs.

Cheers,
Curtis

worstje:
### Regarding dictionaries and ordering and stuff...

I don't think I clicked 'de' the first time by mistake, but at this point it is hard to make sure one way or the other. I am not quite sure I follow most of your explanation, but I do feel that the way dictionaries work or are ordered should be more transparent to the user. It is just a tiny visual wart right now, but since one seems to be generated from the other (but only in the case there isn't a dictionary already present?) it feels like the way dictionaries are listed right now is going to bring along a huge amount of confusion. At the very least I consider myself less likely to be confused when using a program like this, especially a pretty standard feature like spellchecking, so wouldn't new users or less technologically-inclined users be even more confused by the way the dictionaries happen to work out right now?

On a positive note, the items do seem to be correctly ordered right now, so I can only assume the ordering issue is one that only happens the first time the application converts those dictionaries.

### A forgotten little detail from my tests yesterday...

Unlike symspellpy, pyspellchecker seems to be optimized for results and not speed. (Admittedly, it does seem to find better suggestions.) I have a word I tested with yesterday that locks up the program for almost five seconds (!!) when eventually not finding any suggestions. It just looked like the program locked up for no reason whatsoever, and this was very annoying from a user perspective.

May I recommend setting the system cursor to a busy/working/hourglass state while calculating the suggestions for the purpose of this PR? So that I can feel as if it is actually doing work for me while I wait?

Ideally we'd optimize away the wait for spelling suggestions by sneaking that into a background process that populates a cache with suggestions which is kicked off by the detection of an until-then unknown misspelled word. That way users always get an instant popup menu, and because it is in the background most people won't notice or be bothered by the CPU usage. (Such an optimization would probably need a toggle so that it can be disabled for slow systems or systems running on a battery.)

### Regarding menu stuff and other UI ideas...

Since I have already brought up accessibility adn menu-related issues on two occasions now (or was it three?), maybe I can take the lead in making a PR which tackles and refurbishes all the different menus in the application. This is something technically simple that is within my capacity to do given my almost completely non-existent knowledge of Qt and also give me a chance to learn the codebase a little bit more.

Regarding the way a spellcheck dialog (or spellcheck settings screen) would look... I could definitely sketch some rough ideas once I know exactly what sort of settings / features we'd like to have.

For example, many word processors have the different concepts of 'ignoring a word' and 'adding it to the dictionary'. Similarly, I imagine I wouldn't want terms specific to one particular novel to end up in a generic dictionary.

Alternatively (or 'in addition'? Food for thought), it would be very nice to have names of characters or world concepts automagically whitelisted, or at least compiled to an auxiliary dictionary of sorts. Of course, since you could define multi-word strings like 'Blargh McBaggles' as a character name, you'd have to separate such things up into different words for it to work with the spellcheck. Same with world items. If you decide to change the name of a particular place or town, it would be nice for it to suddenly become an 'unknown spelling' so you can easily spot any mentions of outdated terms while editing.

This is once more out-of-spec for this PR as I tend to digress, but the point is: we'll want to have a clear understanding of how we want spellchecking to work exactly and what features would be good for manuskript to have. That way the UI can be optimized for what we need it to be as opposed to just a bland copy of an existing word processor. Manuskript has unique features to pave the way for creativity and we should not forget about those while implementing a feature that has without a doubt been done better in other programs already.

gedakc:
@kakaroto, when you're back will you have a chance to update the file headers as mentioned in an [earlier comment](https://github.com/olivierkes/manuskript/pull/507#issuecomment-483336191) in this PR?

When this is done then I can merge this PR with the **develop** branch

kakaroto:
Hey @gedakc, I got back home yesterday. I've done the small change you requested and pushed it. I hope it's all good.

@worstje I don't think the dictionaries order is much of an issue, I would expect 99% of the users to select their dictionary once and never go into that menu ever again. As for not understanding the details of how it works, that's fine, that's the implementation details, as a user, you wouldn't need to understand it.
pyspellchecker is slow to do suggestions, really slow and that's the whole reason why I added the support for symspellpy and switched to it. 
Currently, the symspellpy suggestions are set to be within a distance of 2 characters (so between "helo" and "help" or "hello", the suggestion distance is 1, but between "helo" and "helicopter" it would be a distance of 6 characters). I'm not sure why you say pyspellchecker gives better suggestions because it also uses an edit distance of 2 by default. Being able to configure symspellpy/pyspellchecker to decide what distance you want would be part of the new spellchecker settings dialog. Higher distance in pyspellchecker means it will be even slower to find suggestions, but for symspellpy, it would be the same speed, but the dictionary cache size would be bigger (26MB for a distance of 2, 54MB for a distance of 5 for en_US, for example)

For the menu/setting design, feel free to do that and send mockups, I'm not good with UI so I welcome anyone who wants to handle that, and I can later implement the design if you'd like. Having custom dictionaries per project should be easy too and I already suggested one way to do it in https://github.com/olivierkes/manuskript/issues/505#issuecomment-466632839 but having it done by manuskript itself would be great, being able to add a word to a dictionary but choose if we add it to the language dictionary or to the project custom dictionary, being able to see both lists in the settings window and add/remove/move words from one dict to another, and be able to switch from one language to another but keep our custom words, etc..

gedakc:
@kakaroto, 'hope you enjoyed your time away.

I confirmed that additition of the file header and will merge this PR with the **develop** branch.

PR #508, opened 2019-02-22
Fix corkView background image on Windows

kakaroto:
Windows path to the image has '\\' path separator instead of '/' which makes
the stylesheet fail. Background images don't appear and console gets spammed with :
Could not parse stylesheet of object corkView(0x27248eb6900, name = "corkView")

I grew tired of that message on the console and hadn't realized the corkview was supposed to have a background image. It looks much nicer now.

Note: I have not tested if this causes issues in case the path is supposed to have some escaped characters in it. I don't know if we're supposed to urlencode the path or something like that, but for now, this works for me.

### DIFF
--- a/manuskript/ui/views/corkView.py
+++ b/manuskript/ui/views/corkView.py
@@ -36,7 +36,7 @@ class corkView(QListView, dndView, outlineBasics):
             background-attachment: fixed;
             }}""".format(
                 color=settings.corkBackground["color"],
-                url=img
+                url=img.replace("\\", "/")
         ))
 
     def dragMoveEvent(self, event):


gedakc:
Thank you @kakaroto for creating a patch for the missing background image in the Editor pane on Windows.

I tested on both Kubuntu 16.04 and Windows XP and confirmed that the corkview image is now displayed in the editor pane when I have selected a chapter.

Thanks,
Curtis

kakaroto:
Humm.. I didn't realize the bug would be reproducible on Kubuntu. Either way, thanks for the merge! :)

gedakc:
> I didn't realize the bug would be reproducible on Kubuntu.

Perhaps I could have used better wording regarding my testing.  The bug did not occur on Kubuntu.  My testing confirmed that the patch did not break things on GNU/Linux.

Thanks again for the patch.

kakaroto:
Ah of course. That makes more sense :)

Issue #234, opened 2017-11-27
Making Fullscreen Mode Great Again


hugo-id:
Fullscreen is where I spend most of my time in the app, as I think most writers probably would, and feel like this view is a bit half-baked in comparison to the other options.

Plume Creator:
![image](https://user-images.githubusercontent.com/30918735/33254061-171d1838-d39b-11e7-98c5-ad0ffbaa154b.png)
In the bottom-right you can see the super-simple interface that makes plume-creator one of the best to simply be productive.
The small green arrow, when pressed, shows all scenes/chapters/books, allowing easy navigation.
The left / right arrows allow you to simply change between the adjacent scenes (allowing distraction-free proof-reading)
The button to the right of those, creates a new scene after the current one. (Cannot believe that plume creator is the ONLY writing app to allow this function)
Then:
Words in current scene,
Time,
Goal progress.
The three buttons underneath are basically broken (notes, settings, close)

Focuswriter:
![b_focuswriter](https://user-images.githubusercontent.com/30918735/33254232-d19b784e-d39b-11e7-9d43-3fea23a92936.png)
In the bottom-right is an incredibly basic implementation of a simple status indicator.

Focuswriter's general workflow and interface is awesome, but perhaps his focus on focus has made it less functional. It is infuriating that to check the time I need to alt+tab, or that the only way I can get a nice update on how much I've written is to check in with an alarm (which is what i posted a screenshot of).

Overall, it feels like fullscreen mode in Manuskript has been "implemented" but not fleshed-out. I feel that between Plume and Focus, there is a perfect fullscreen mode that writers are all waiting for.

Things I feel are invaluable:
The time
Goal Progression
Current words
New Scene
Type-following (either typewriter, or via virtual pages)
The whole "lock screen" concept is bizarre. If I don't want to type, I'll just walk away. That whole interface may be better served by turning it into a basic goal setting widget. (And once you've started it, there is no feedback or way to passively monitor progress)

In case I wasn't clear, the "distraction-free" mode is incredibly personal to the individual in question. Please keep in mind that any option, no matter how widely-requested will be kryptonite to someone, and that as many options as possible in fullscreen mode should be customisable.

Manuskript is genuinely amazing, and I genuinely prefer it over scrivener 1.9. Hopefully you can keep up the good work, and linux will have yet another best-in-class app!

olivierkes:
Thanks for this detailed feedback.

Have you seen that you can right-click on a panel, and **disable "auto-hide"**?
[...]

Issue #199, opened 2017-11-06
Enhancement: disable `ALT+TAB` when in locked fullscreen


olivierkes:
Right now, user can lock fullscreen (either time or word count), but easily bypass that with `ALT+TAB`.

gedakc:
@olivierkes I think it is important to always have an easy and obvious way to exit from manuskript.

Consider the following scenario:

* A user has invoked a manuskript lock for a time period or word count.
* Someone encounters a health emergency.
* Can the computer be used to access help via the Internet or another application?

olivierkes:
Good point. Isn't too easy right now, though?

gedakc:
I have to plead ignorance here.  I thought this issue was discussing a future feature.

How does one lock manuskript in full screen mode now?

gedakc:
Thanks for pointing out how to use lock fullscreen mode.  I would never have guessed this, as I suspect that our users wouldn't guess this feature existed either.  Unless they normally move the mouse back and forth and up and down over the entire screen.  ;-)

We've got our work cut out for us documenting the many features of manuskript.  On the documentation topic, great work describing the [import/export capabilities](https://github.com/olivierkes/manuskript/wiki/Import-and-Export-capabilities).  :+1: 

> Isn't too easy right now, though?

I agree that it is not very easy to exit from lock screen mode, especially for a user not familiar with manuskript lock screen mode.

When limiting user actions, I try to think of potential adverse side effects.  For instance if a smart phone app locked the screen so that phone calls could not be made.  Then the phone owner experiences a health issue.   Now how can a bystander use the phone to call emergency services?

hugo-id:
You could throw a warning that alt + tab will add 10% to goal requirements.

If you're being slack, you've now given yourself extra work, and in an emergency nobody cares

PR #509, opened 2019-02-22
Fullscreen panels improvements

kakaroto:
Experimental feature, if you merge it great, if not, that's fine, the purpose here is really to start discussing the feature and whether I should continue on this same path. The Pull request is for everyone to be able to test the changes and see if they like them.

The main feature here is that the panels can now have "displays" which is a bad name for a group of widgets that go together to display some sort of info to the user. The context menu of the panel can allow you to enable/disable these displays, and the settings get saved so it stays persistent.
The top panel now has the scene title in the center (finally!) and both the title and the spellcheck button can now be shown/hidden, the bottom panel can show/hide the theme selector, the word count, progress and a new clock widget I added. I also added a setting to decide if the clock should show seconds or not. 
This is really more of a proof of concept (especially the "clock: show seconds" which I'll remove, unless people actually like it), I'll let people who are better at UI design decide what to display and where.

This is of course a first step for fixing #234 and my plan is to have the top panel show a : 
`<Left arrow> <Folder name> - <Subfolder name> - <etc..> - <Text/Scene title> <Right arrow>`
The Left/Right arrows obviously would allow us to move to the previous/next text entry sequentially (don't know yet how the model is defined, I feel like that's going to be the difficult thing to figure out), and clicking on the Text name would pop a combobox with all the texts in that folder, same with clicking on the folder names, so you could jump to another chapter more easily (would show the first text entry in that folder). 

I'm not sure where to put the "New scene" button, maybe as one of the entries in the text combobox, or right next to the Scene title name, and have spacings on the left and right for the arrow. If someone has a better idea, let me know!

P.s: I'm not going to actively work on this for now until I get some feedback on how to proceed. I've opened the pull request so the work can be tested and people can decide if they like it or not. In the meantime, I'll get back to actually writing my manuscript :)

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -122,6 +122,7 @@ class abstractItem():
     def row(self):
         if self.parent():
             return self.parent().childItems.index(self)
+        return None
 
     def appendChild(self, child):
         self.insertChild(self.childCount(), child)
@@ -176,6 +177,11 @@ class abstractItem():
         item = self.__class__(xml=self.toXML())
         item.setData(self.enum.ID, None)
         return item
+
+    def siblings(self):
+        if self.parent():
+            return self.parent().children()
+        return []
 
     ###############################################################################
     # IDS
--- a/manuskript/models/outlineItem.py
+++ b/manuskript/models/outlineItem.py
@@ -80,6 +80,16 @@ class outlineItem(abstractItem):
     def wordCount(self):
         return self._data.get(self.enum.wordCount, 0)
 
+    def __str__(self):
+        return "{id}: {folder}{title}{children}".format(
+            id=self.ID(),
+            folder="*" if self.isFolder() else "",
+            title=self.data(self.enum.title),
+            children="" if self.isText() else "({})".format(self.childCount())
+            )
+
+    __repr__ = __str__
+
     #######################################################################
     # Data
     #######################################################################
--- a/manuskript/settings.py
+++ b/manuskript/settings.py
@@ -34,6 +34,12 @@ viewSettings = {
         },
     }
 
+fullscreenSettings = {
+    "autohide-top": True,
+    "autohide-bottom": True,
+    "autohide-left": True,
+    }
+
 # Application
 spellcheck = False
 dict = None
@@ -119,10 +125,11 @@ def save(filename=None, protocol=None):
     global spellcheck, dict, corkSliderFactor, viewSettings, corkSizeFactor, folderView, lastTab, openIndexes, \
            autoSave, autoSaveDelay, saveOnQuit, autoSaveNoChanges, autoSaveNoChangesDelay, outlineViewColumns, \
            corkBackground, corkStyle, fullScreenTheme, defaultTextType, textEditor, revisions, frequencyAnalyzer, viewMode, \
-           saveToZip, dontShowDeleteWarning
+           saveToZip, dontShowDeleteWarning, fullscreenSettings
 
     allSettings = {
         "viewSettings": viewSettings,
+        "fullscreenSettings": fullscreenSettings,
         "dict": dict,
         "spellcheck": spellcheck,
         "corkSizeFactor": corkSizeFactor,
@@ -198,6 +205,10 @@ def load(string, fromString=False, protocol=None):
             ]:
             if not name in viewSettings[cat]:
                 viewSettings[cat][name] = default
+
+    if "fullscreenSettings" in allSettings:
+        global fullscreenSettings
+        fullscreenSettings = allSettings["fullscreenSettings"]
 
     if "dict" in allSettings:
         global dict
--- a/manuskript/ui/editors/fullScreenEditor.py
+++ b/manuskript/ui/editors/fullScreenEditor.py
@@ -65,6 +65,11 @@ class fullScreenEditor(QWidget):
             self.btnSpellCheck = None
 
         self.topPanel.layout().addStretch(1)
+        item = self._index.internalPointer()
+        title = item.data(Outline.title)
+        self.lblTitle = QLabel(title, self)
+        self.topPanel.layout().addWidget(self.lblTitle)
+        self.topPanel.layout().addStretch(1)
 
         # Close
         self.btnClose = QPushButton(self)
@@ -120,6 +125,7 @@ class fullScreenEditor(QWidget):
 
         if self.btnSpellCheck:
             self.topPanel.addDisplay(self.tr("Spellcheck"), 'top-spellcheck', (self.btnSpellCheck, ))
+        self.topPanel.addDisplay(self.tr("Title"), 'top-title', (self.lblTitle, ))
         self.bottomPanel.addDisplay(self.tr("Theme selector"), 'bottom-theme', (self.lstThemes, themeLabel))
         self.bottomPanel.addDisplay(self.tr("Word count"), 'bottom-wc', (self.lblWC, ))
         self.bottomPanel.addDisplay(self.tr("Progress"), 'bottom-progress', (self.lblProgress, ))
--- a/manuskript/ui/editors/fullScreenEditor.py
+++ b/manuskript/ui/editors/fullScreenEditor.py
@@ -73,14 +73,13 @@ class fullScreenEditor(QWidget):
         self.btnNext.setFlat(True)
         self.btnNext.setIcon(QIcon.fromTheme("arrow-right"))
         self.btnNext.clicked.connect(self.switchNextItem)
-
-        # Title/Path and New document Buttons
-        self.lblTitle = myTitle(self)
-        self.wPath = myPath(self)
         self.btnNew = QPushButton(self)
         self.btnNew.setFlat(True)
         self.btnNew.setIcon(QIcon.fromTheme("document-new"))
         self.btnNew.clicked.connect(self.createNewText)
+
+        # Path and New Text Buttons
+        self.wPath = myPath(self)
 
         # Close
         self.btnClose = QPushButton(self)
@@ -97,7 +96,6 @@ class fullScreenEditor(QWidget):
         self.topPanel.layout().addWidget(self.btnNew)
 
         self.topPanel.layout().addStretch(1)
-        self.topPanel.layout().addWidget(self.lblTitle)
         self.topPanel.layout().addWidget(self.wPath)
         self.topPanel.layout().addStretch(1)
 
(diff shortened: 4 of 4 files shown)


kakaroto:
I just added navigation support to the fullscreen editor and it makes it so so much more useful. This was the number one feature that I wanted and I'm very happy with the results (especially with the idea of adding hotkey support and the myPath popup navigation)! Now someone needs to come in and decide on a (potential) better UI design, or at least on a pixmap to use as separator for the myPath, and a stylesheet to make the path's flat buttons more visible.

Copy/pasting info from commit message : 
```
Top panel now has left/right arrows to navigate through the scenes. It will
automatically find the next/previous text item and display it, navigating through
the outline tree.

There's also a "New document" icon which will create a new text entry immediately
after the current one and switch to it.

Navigation can also be done using Alt+Page-Up, Alt+Page-Down or Alt+Left and Alt+right
shortcuts (Fixing #444).

There's now also the option between Title or Path for the top panel, if Path is chosen
then the full path of the scene is displayed and clicking on the scene or parent items
opens a menu to quickly switch to the selected chapter/scene. Selecting a folder will
automatically display the first text entry available in that folder.

```

tildagail:
I don't know if this is the place to leave feedback or not. Sorry if it's wrong.

This is fantastic! I'd love to see it as part of Manuskript. I have a few feedback items:

* The next/back buttons change position depending on the length of the path (or title), so I can't just automatically go to the same spot on the screen every time. Fixing them in a single place seems better for usability.

* I saw in the issues section someone wanted the summary to appear somewhere in the fullscreen panels (#527). How about as a third option in the title/path section of the top panel?

* Weird things seem to happen to the locker if you change scenes. E.g. set goal of 500, lock scene, go to the next scene. Goal is now 366. Go to the next scene. Goal is now 417. Go to the next scene. Lock deactivates. I'd prefer the lock to either disable the forward/back shortcuts or else have the same value persist between scenes so you can add words to any scene to reach the goal, but not leave fullscreen mode.

* I saw keyboard shortcuts for back/forward, but not one for creating a new scene... although now that I think about it, Manuskript doesn't have a 'new scene' shortcut even in the main editor, does it?

* Total wishlist, but showing the chapter/folder/whole project wordcount would be nice...

kakaroto:
@tildagail thanks for the feedback, I'm glad you love it!
And this might be the right place to do it, or maybe the right place would be #234 and keep comments here about actually reviewing the code in the various commits. I'm not entirely sure to be honest, but most your comments were specifically about issues/bugs in this implementation rather than about the feature so I think this is the right place.

To answer your feedback : 

1. Yes, the arrows changing position is very annoying which is why I just use the hotkeys instead. I'm not a UI person so I was hoping someone would find a better layout arrangement to get this fixed. 
2. Humm.. maybe the one line summary would work, yes. I think in #527 they were talking about the full summary text though. I'll add "One line summary" as one of the options besides path/title. That's a good idea.
3. Yes, I noticed it unlocks when we change scenes, I thought I mentioned it when I sent the PR but I guess I forgot. I tested again and it seems it doesn't just unlock, it actually just changes the target to the scene's target (if you have 400 words written with a goal of 500, then your lock goal becomes 100) and if you switch to a scene where the total words is superior to the goal, that's when it unlocks it. I wouldn't want lock to disable the forward/back because I think if you want to lock for 15 minutes and want to write in multiple scenes during that time, you should be able to. I have no idea how it should react when the lock target is in words instead of time based. It should technically retain the same value and not unlock if you stumble on a scene that already has its goal reached, but do people use the "word goal" lock as a way to "finish this scene" or really about "write X words before I'm done"... Since fullscreen can now work on multiple scenes, I think it becomes the later rather than the former.
4. I don't see a point for such a shortcut, the hotkeys for previous/next are useful since you can spam them to quickly move/navigate the scenes, but creating a new scene would be a relatively rare occurrence I think
5. Where? Bottom panel as an option to display "Parent folder's total wordcount".. that could work, I'd have it disabled by default though I think. I don't know where in that line it would best fit... Btw @olivierkes or @gedakc would the right term really be "folder" or "chapter"?

gedakc:
Following are responses to points made by @kakaroto.  Note that I have not had time to review the code yet.

1.  Normally the top bar in a UI is reserved for Window controls (minimize, maximize, close) and an informational title.  I think it would be preferable to move the folder/text navigation arrows down to a fixed location on the bottom bar.  At the same time I suggest moving the spell check toggle down to the bottom bar as well.

2. The summary title might be long which might cause display issues placing it in any of the bars.  I guess if it is optional then the user could turn it on or off as appropriate.  If the summary title was line-wrapped then perhaps it would work in a wider side bar.  That's just a thought.  The fullscreen feature is after all for **distraction free** writing.  We are not trying to re-implement all of the features contained in the main Manuskript screens.  :-)

3. I don't think we should disable scene navigation when lock is enabled.  See also comments in issue #199 regarding disabling **Alt+Tab**.

4. My thoughts are that in fullscreen mode, where one is trying to work distraction free, does one create new scenes frequently enough to require a keyboard shortcut?  I'm thinking not.

5. Because manuskript supports multi-folder structures with user editable names I think it would be better to use **folder** and **text** as terms.  For example a novel structure might be Book/Part/Chapter/Scene, whereas a research project might be Topic/Section.  Further in the right-click menu in the **Editor** pane we currently use the terms **New Folder** and **New Text**.

worstje:
@kakaroto : my apology for responding to your message in the other issue and this entire chain rather late. Life got to me, and on top of that my Python installation was properly knackered so it took me far longer today than I'd like to admit to actually get to try out your build.

I have barely managed to play with it so far, but I didn't want to go to bed without actually responding to you first. 😄 

I see that in the meanwhile tildagail already pointed out how the buttons at the top move about when you switch scenes. That is definitely awkward. Personally, I would probably prefer to have the buttons next to eachother, so that if you click too far ahead or back, you can easily switch back without moving your mouse a huge distance to try and hit a tiny button. (I am using a 2160p screen here.)

The new-text button is not something I see any need for. Frankly, I find the icon to be a beacon of glaring white in my dark layout, and I'd much rather want it gone. (Besides, I am just liable to press it by accident, and as it is I just found out that I have somehow already corrupted my project into having some files that share the same ID, a matter that I'll have to dig into tomorrow. Even more texts with more IDs accidentally popping up are the last thing I need at this moment in time.)

Regarding the one-line-summary and full-summary things I mentioned elsewhere and I think were also mentioned here: I understand we don't want to replicate all of the normal Manuskript UI in the fullscreen editor. That would be weird.

However, my current approach is to either keep quitting Manuskript, or to copy the summary into the actual text while writing.. and then I end up having to scroll up and down a bunch to find where I pasted it while I go through my iterative 'adjust every previous sentence a dozen times' approach of writing a new paragraph.

I don't necessarily need it to be there all the time; for the normal summary I would definitely be completely in agreement with a panel similar to the lock on the left that only shows when you hover around that edge. Combined with the existing autohide functionality and adjustable text margins I could pretty much achieve the goal of having a simple overview with my aims for that text visible to me even when my mouse is not on the edge.

The one-line summary is a bit awkward. I think it can be an excellent source of inspiration, just like the more in-depth summary. Rather than getting stuck in the minutiae of the plot points I need to address with that one, I find my one-line-summaries tend to evoke the mood and feeling I wanted a text to have when I came up with the direction I was writing in.

With your current build, I feel the one-liner would probably be best fit in the bottom-bar, while the Theme Selector could be moved to either the top, or to be put on a panel like the Lock screen. (Honestly, I feel that control especially is an awkward one to have. It isn't informational and I personally don't know why anyone would be so eager to change the way their editor looks mid-writing. If the ability to keep the skin changing facility is indeed desired, I'd look for a way to stick it into a popup menu of some kind. That selectable list of themes is perfect to be squeezed into a sub-popup, after all. The only question is where: if it is stuck in a popup context menu, what do you right click? It doesn't fit on the bars at that point, and I personally have no interest of seeing it in my right click menu of the editable area where you get cut/copy/paste/spellcheck/whatever. I guess the non-text area (padding? margins? I am not sure what the correct term is here) would be perfect if it could be made to pop up a menu.

This ended up way longer than intended after all. Whoops. 😬

gedakc:
@worstje if you copied a folder or text (not simply the contents) then that can lead to duplicate id's.  Unfortunately the fix involves manually changing the copied id's, or copy/pasting the content to a newly created text and removing the text with the duplicate id.

To avoid this bug only copy content, not whole folders and texts.    

See existing issue 290 and issue 324.  Note that these are not hyperlinks because the copy bug is not related to this PR and I did not wish to link these together.

worstje:
Yes, those issues you linked seem to be related. I am quite sure I never used Copy-Paste in the outline, but I did drag files to a different folder on a few occasions, once even causing a crash I think.. and I just noticed I somehow seem to have duplicated one of the entries in two places.

Since two issues for highly similar bugs are already open, and I am not quite sure how my corruption came to be (it has been at least two weeks) while suspecting the same code is to blame, I'm not going to file a new issue unless it happens again and it is sufficiently different in cause.

kakaroto:
@worstje no worries about delays. It's called life!

Thanks for the feedback on the buttons. I had also wondered about putting them close to each other but I didn't know where (on the left? on the right? how far away from each other ? etc..) so I'm leaving that task to those who have a better sense of UI design than me. 

I agree with the new-text button, I don't like having it either, but it was requested by others so I added it. If you don't like it, just remove it, that's the whole point of having the panels customizable. The fact though that you make the comment (and other similar ones) lends me to think you never tried to right click on the panel. Give it a try and see how you can customize the panels.

Note: you can ALT-Tab instead of quitting full screen, I often do that when I need to look at metadata.

The theme selector is definitely awkward but I didn't want to change it because that would be changing existing behavior/functionality. But you can just remove it with the right-click config menu, so it's a non-issue.

gedakc:
@kakaroto, I think this PR would be a good improvement that I would like to include in a new release soon.

Would you be open to left aligning the navigation to butt up against the spellcheck indicator?

We should also consider placing the left navigation arrow, right navigation arrow, and plus icon before the title.  That way the navigation icons should not "bounce around" based on the length of the title.

kakaroto:
@gedakc yes of course, I'm open to suggestions. Like I said before, I suck at UI design and this started more as a proof of concept of where I could take this, and I'm still waiting on someone to tell me how to better design the UI.
Left aligning the navigation sounds good. Should it be : `spellcheck - Left - New - Right <span> - Title <span> Close`  ? 
Or maybe swap the left/new/right into left/right/new ?
Also, I'm confused, in your last sentence you say "also consider placing the navigation arrows before the title" and I thought that was what you meant by the second sentence which said to have them butt up against the spellchecker icon... 
Could you maybe have a mockup/drawing to make it clearer ? Or should I just make the change the way I understood it then you can tell me if I got it right or not ?

gedakc:
> Left aligning the navigation sounds good. Should it be : `spellcheck - Left - New - Right <span> - Title <span> Close `?
> Or maybe swap the left/new/right into left/right/new ?
>
> Could you maybe have a mockup/drawing to make it clearer ? Or should I just make the change the way I understood it then you can tell me if I got it right or not ?

'Sorry for the confusion.  I think it should be:  
`spellcheck - Left - Right - New  <span> - Title <span> Close `

Note that I'm not a GUI designer either.  With that said I think we agree thought that having buttons bounce around is not a good design.

I think your enhancements will make for a better writing experience for full screen users.

kakaroto:
Done!
I don't like the navigation buttons on the side instead of surrounding the title, but I don't see how that could be done without having the arrows moving around when title length changes.

I also found an issue with the Progress disappearing if we navigate to a scene without a goal, and realized it won't appear if we navigate to one with a goal. I added a "Progress: Auto Show/Hide" setting to take care of that use case.

gedakc:
Thank you @kakaroto for updating the navigation in this PR.  I think the new navigation button locations are a good compromise.  I especially like that a person can enable or disable each individual feature so that they can add only those features they think are necessary.

While testing, the following questions arose:

1.  Should the fullscreen top bar implement "radio button" controls?

    With the cursor at the top of the screen, if I enable "title" then this disabled "path" and only the title is shown.  The reverse is true.  These two controls appear to be operating like a radio control button (only one choice may be active).

    For simplicity I would be okay with choosing only one of these options, such as "Path.

2.  What does the "Progress: Auto Show/Hide" entry in the fullscreen bottom bar implement?

    I tried enabling and disabling this feature, but did not observe any difference in the bottom bar.

3.  Would you like to update some of the commit messages?

    A. In the first commit, it mentions that "_This is an experimental idea._", but I think these improvements are more than just an experiment.  I'm okay if you decide to keep this wording.

    B. In the last commit, the first word is mispelled  "_fulscreen: Add a auto-show/hide progress setting._"  
    I know that "fullscreen" isn't the proper spelling either, but it is what the original author used when describing the feature.

kakaroto:
1. It technically should but I didn't want to make it too complicated. With the purpose of keeping the menu simple I kept them that way, I figured any user would realize it's either Title or Path after a couple of clicks. In order to do radiobuttons we'd need to add a separator then an action group, then the items and `addDisplay` is not the API for that. I thought of maybe changing the text into "Center: Title, or" and "Center: Path" to make it clearer but I don't think that's a good idea either and it's not a big problem for people to figure out.
2. Ah, for that, you should try to navigate between a scene that has a goal set vs a scene that doesn't. The auto-show/hide will auto-show the progress if the scene has a goal and will auto-hide it if the scene has no goal set (previous code auto-hid the progress when fullscreen on a scene without a goal which overrode the setting of showing the progress or not). Read the commit message of 1f0eccd66 where I explained the issue.
3. That specific commit was experimental, it sort of got out of the experimental stage with further commits, so I don't think it's an issue to keep the term in that first commit. Also note that the fullscreen features aren't what is experimental, the experimental feature/commit is about adding items to the context menu of a panel. I'm still not convinced of calling them "displays", I don't know why I chose that name but I don't really have a better option. I think the `addSetting` API should also get a function to call back when the option changes. I'm also wondering if that whole thing should be removed and replaced with a more static/proper menu design instead of a generic API like that. It would allow us to have the radio buttons discussed in (1).
Sorry about the typo in the last commit, will fix it asap :)

gedakc:
1.  I'm on board to keep it simple so we can ignore radio buttons.  

    A. Another way to approach this is to make the "Path" option simply prefix the path at the beginning of the "Title".  That way a radio button would not be required.  It would; however, let a person show just the "Path" without a "Title", but I think that would be easier for a user to figure out.

    What do you think of such an approach?

    B. Another thing that might help make things clearer would be to place the option list in the same order that the buttons are shown on the screen.  More specifically I'm thinking of the following order:

    TOP BAR  

    Manuskript-Fullscreen-TopBar

    - **A**uto-hide
    - Spellcheck
    - Navigation
    - New **Text**
    - Path
    - Title

    C. Note that I changed "_New Document_" to be "_New Text_" to be consistent with other areas of the UI that refer to _Folder_ and _Text_.  I'm referring to the project terms in the **Project tree** pane right-click menu.  
    I also made the first letter capitalized in "_**A**uto-hide_" to be consistent with the bottom bar.

2.  Thank's for the explanation.  I can see the difference now by using a sample project that contains texts both with and without word count goals.  :-)

kakaroto:
> That solution should work and be less confusing.

It's done! And yeah, it's much nicer now! I'm glad I found a way to do it in a less confusing way! :D

I also reworked the API, replaced `addDisplay` with `addWidgetsetting` and I added the `setSettingCallback` API. I'm much happier now with the panel code too.

> Your are 100% correct. I could have sworn that the top bar had lowercase "auto-hide", but upon checking again it is definitely an uppercase "Auto-hide".

You probably saw it non-capitalized when reviewing the code and thought it was in the UI. I had the same impression today when I saw this line :
`        if hasattr(widget, "_autoHide") and not widget._autoHide:`

There's the small issue of how the locking screen behaves when it's a word-count lock and we navigate out of the scene, and the suggestion to add a summary/metadata view in the left panel, but I think those can wait after this release.

gedakc:
My first impressions are that I like these most recent enhancements to fullscreen.  I'll do some more testing with an eye towards merging these changes as early as tomorrow.

> There's the small issue of how the locking screen behaves when it's a word-count lock and we navigate out of the scene, and the suggestion to add a summary/metadata view in the left panel, but I think those can wait after this release.

I agree that these can wait for a future release.

gedakc:
This enhancement looks good to me.  In my testing each feature worked, and no regressions were observed.  As such I will rebase and merge this PR.

@kakaroto would you be able to update the wiki for [Full screen mode](https://github.com/olivierkes/manuskript/wiki/Full-screen-mode)?

kakaroto:
Thanks for the merge! I'll update the wiki in a few days.

kakaroto:
Wiki page updated, feel free to edit/clarify/improve it if you feel it's necessary.
Thanks

gedakc:
Awesome work on the [Fullscreen Mode](https://github.com/olivierkes/manuskript/wiki/Full-screen-mode) wiki documentation @kakaroto!  The document includes text and picture descriptions, covers all features, and is easy to follow.  Great job!

Yesterday we released [Manuskript 0.9.0](https://www.theologeek.ch/manuskript/2019/04/04/manuskript-0-9-0/).  I like to leave some downtime before committing new changes to the **develop** branchjust in case a serious problem arises with the new release.  As such I plan to look at the new spellcheck PR later next week.

PR #510, opened 2019-02-22
Add ability to add new background images through UI.

### LINKED ISSUE #399: [Feature Request] Add Ability to Add Image When Creating Fullscreen Theme
When creating fullscreen theme, ada ability to add user's image as background. For now, it only can add space and table image and color.
screenshot from 2018-10-13 17-20-41

kakaroto:
The Cork background and fullscreen theme backgrounds images can now be added
by using the "+" icon from the combobox. Once a file is added, the combobox
is repopulated and the new image is selected.

Note on line 871, in updateThemeBackground, there was a bug where it was using
self.cmbCorkImage instead of self.cmbThemeBackgroundImage

Fixes #399.

### DIFF
--- a/manuskript/settingsWindow.py
+++ b/manuskript/settingsWindow.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # --!-- coding: utf8 --!--
 import os
+import shutil
 from collections import OrderedDict
 
 from PyQt5.QtCore import QSize, QSettings, QRegExp, QTranslator, QObject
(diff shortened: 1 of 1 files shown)


gedakc:
Thank you @kakaroto for this enhancement to add new background images to user created **Fullscreen** themes.

From the **Edit -> Settings -> Fullscreen -> {_my New theme_} - Edit** screen, I confirmed that I was able to add a new background image using the "+" icon in the **Image** drop down menu list.

Is there a way through manuskript to remove the newly added background image from the **Image** drop down menu list?

Note:  On my Kubuntu 16.04 GNU/Linux I noticed that the new background image is copied to the `~/.local/share/manuskript/manuskript/resources/backgrounds/` folder.  If I delete the background image from this directory then the image is also removed from the **Image** drop down menu list.

kakaroto:
You're welcome (you don't need to formally thank me in every PR, after the 3rd one, I think we're good 😄 )! 

There is no way at the moment to delete the backgrounds other than manually removing the files. I figured I'd add that later as a separate PR.
I'm not sure how to do that though... I wanted to use the '-' icon for that (select an image, then select the '-' icon which will ask for delete confirmation if it's a custom bg) but it would be a behavior change for that icon, also would need a new icon for 'no background' (maybe a red circle with a diagonal bar?)

Note: this change also affects the cork view background, in **Edit -> Settings -> Views -> Index Cards -> Background**

gedakc:
> There is no way at the moment to delete the backgrounds other than manually removing the files. I figured I'd add that later as a separate PR.

Okay, we can look at that as a future enhancement.

PR #512, opened 2019-02-23
Fix background of popup menus that were transparent (black)

kakaroto:
In the properties view, the context menu on the title line would be black
making its content unreadable. Same in the filter line of the "Set Custom icon"
window on the outline's context menu.

Note: This is similar to #504, not sure if you'd prefer the commits to be squished and sent as a single PR.

The propertiesView_ui.py file, I modified manually instead of auto-generating it, I consider the change to be small enough that it's not needed and it's risk free, besides I don't have pyuic installed and I don't know if we'd have the same version and whether the result would be exactly the same and I wanted to limit the amount of changes the diff would have.

Note that in outlineBasics on line 223, there's a `QListWidget` that also uses a similar stylesheet : 
``` self.lstIcons.setStyleSheet("background: transparent; background: none;")```
I don't know if the 'background:none' is overriding the 'background:transparent' and I don't know if a transparent background is supposed to be there for children since `QListWidgetItem`s might need it, so I just left it as is. Either way, there is no context menu on that widget, so I think it's fine, but you might know better so I thought best to give you the FYI.

### DIFF
--- a/manuskript/ui/views/outlineBasics.py
+++ b/manuskript/ui/views/outlineBasics.py
@@ -203,7 +203,7 @@ class outlineBasics(QAbstractItemView):
             txt = QLineEdit()
             txt.textChanged.connect(self.filterLstIcons)
             txt.setPlaceholderText("Filter icons")
-            txt.setStyleSheet("background: transparent; border: none;")
+            txt.setStyleSheet("QLineEdit { background: transparent; border: none; }")
             act = QWidgetAction(self.menuCustomIcons)
             act.setDefaultWidget(txt)
             self.menuCustomIcons.addAction(act)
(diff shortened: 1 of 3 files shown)


gedakc:
Thank you @kakaroto for identifying this issue and providing a fix.

I confirmed the readability / viewability issue with Manuskript 0.8.0 on Kubuntu 16.04 and Windows XP (see screen shot) by enabling the **Metadata** pane, highlighting the **Properties** title line and right-clicking to display the popup menu.

manuskript-properties-popup-menu-not-readable

This PR makes the popup menu readable on both Kubuntu 16.04 and Windows XP (see screen shot).

manuskript-properties-popup-menu-readable

> Note that in outlineBasics on line 223, there's a QListWidget that also uses a similar stylesheet...

I am also unsure of the implications of changing this section of code.  My thoughts are to leave this section as is because it appears to work.  If we encounter problems later then we can see about fixing such problems at that time.

This PR looks good to me so I will rebase and merge it with the **develop** branch.

PR #514, opened 2019-02-25
Fix crash when right-clicking twice on fullscreen panel

kakaroto:
If you right click once on the fullscreen panel and the context menu pop up
then you right click again somewhere else on the panel *while the previous
context menu is still visible* then it will cause a crash with :
"Windows fatal exception: access violation"
It seems to be caused by a crash in the QT event loop, trying to delete the
existing QMenu within an event handler.

### DIFF
--- a/manuskript/ui/editors/fullScreenEditor.py
+++ b/manuskript/ui/editors/fullScreenEditor.py
@@ -359,6 +359,7 @@ class myPanel(QWidget):
         self.show()
         self.setAttribute(Qt.WA_TranslucentBackground)
         self._autoHide = True
+        self._m = None
 
         if not vertical:
             self.setLayout(QHBoxLayout())
@@ -379,6 +380,8 @@ class myPanel(QWidget):
 
     def mouseReleaseEvent(self, event):
         if event.button() == Qt.RightButton:
+            if self._m:
+                self._m.deleteLater()
             m = QMenu()
             a = QAction(self.tr("Auto-hide"), m)
             a.setCheckable(True)


gedakc:
@kakaroto in which operating system environment and PyQt version did you encounter this crash?

So far I've been unable to reproduce the issue on either Kubuntu 16.04 or Windows XP using the current **develop** branch.

The steps I used were:

1.  Activate **Fullscreen** mode.
2.  Right click in the text area to display the edit popup menu.
3.  With the edit popup menu still displayed, right click in a different text area.

In my testing there were no crashes.

kakaroto:
Interesting! I just tested on Fedora 28 with python3-qt5 version 5.10 and qt5 5.11.3 installed as well as on Windows with the official 0.8.0 release (not sure which pyqt5 version but Qt5 is 5.5.1 according to the DLL metadata) and I couldn't reproduce the issue on those versions. It only seems to happen on my development tree on **Windows 10** (using remote develop branch of course). I have **PyQt5 version 5.12** (with Qt 5.12.1) installed. I also confirmed it crashes with both the 32 bit and 64 bit versions of my python installation.
Also note, the steps to reproduce were wrong, it's not about right clicking on the text area, it's about right clicking on the panel itself (where the popup menu says "Auto-Hide") then while the auto-hide popup is open, right-click again somewhere else on the panel.

gedakc:
Thanks for clarifying the test steps.  I retested and could not reproduce the crash.

Further I tested on Windows 7 and could not reproduce the crash.

Software Versions in Use for Windows 7:
- Python 3.7.2
- PyQt 5.12
- Qt 5.12.1

I'm using the current git **develop** branch at commit 3f217257c47

Are you using the same, most recent **develop** branch commit?

kakaroto:
I had tested on develop branch before you merged my recent commits, I just re-fetched now and can confirm it happens with 3f21725 as well. It's very easy to reproduce for me, 100% crash rate as soon as I right click the second time and I'm using the same versions of python/qt/pyqt as you. This is what appears in the console :

```
Windows fatal exception: access violation

Current thread 0x0000315c (most recent call first):
  File "v:\Projects\book\manuskript-git\bin\..\manuskript\main.py", line 101 in launch
  File "v:\Projects\book\manuskript-git\bin\..\manuskript\main.py", line 114 in run
  File "manuskript", line 13 in <module>
```

I don't know why it's not reproducible for you. I doubt it's a windows 10 only bug though it's possible.

gedakc:
@kakaroto I was able to reproduce the crash on Windows 10.

To install I used the steps in [Appendix A: Install Required Software Packages on Windows 7 and Higher](https://github.com/olivierkes/manuskript/wiki/Run-Manuskript-from-Source-Code-on-Windows#appendix-a--install-required-software-packages-on-windows-7-and-higher).  It looks like this bug affects Windows 10 only.

I will need to perform regression tests in other environments to ensure that the proposed fix does not adversely affect other platforms.

gedakc:
Thank you @kakaroto for discovering this Windows 10 only bug, helping to debug the issue, and providing a code fix.

I have confirmed that this PR does indeed fix the crash on Windows 10.

Further I have tested this PR on Kubuntu 16.04 and Windows XP and confirmed that these continue to work as expected.

kakaroto:
Thanks for the merge and for taking the time to test/confirm the bug on Windows 10! Good to know it's not just my system that's acting weird :)

PR #530, opened 2019-03-07
Providing a suitable icon for consumption by Windows operating systems

worstje:
As of version 0.8.0, the only time a user sees the Manuskript icon / logo is when:

* they download it,
* the application is loaded and it is tucked away in the top-left at a 16px size, or
* when looking in the about box.

In practice, this means the user is utterly unaware of the Manuskript branding. Stumbling across `icons/Manuskript` was a cute little reminder to me personally that a Manuskript logo does in fact exist... but that it is very well-hidden from users on Windows to the point of being unusable!

To solve this, we now have a `manuskript.ico` file, which is the standard format used by Windows for all sorts of icons. Additionally, the PyInstaller `manuskript.spec` has been adjusted to assign this new icon to the generated `manuskript.exe` binary.

This allows for the following usecases / behaviours:

* Viewing the extracted folder now allows new users to immediately discern which file to run due to the distinctive icon.
* Shortcuts created to the `manuskript.exe` file will automatically get the Manuskript icon.
* Developers can create their own shortcuts and have the icon set to the `manuskript.ico` file.
* If we supply a proper installer package in the future, it too can rely on this icon to set itself apart.


gedakc:
Thank you @worstje for creating a way to associate the Manuskript icon with the Windows `.exe` file.

My testing has gone well. With this PR applied, running PyInstaller on Windows XP results in the Manuskript icon being shown in Windows Explorer for the `manuskript.exe` file.  Further there were no regressions in building PyInstaller files on Debian 8 and Mac OS X (Travis CI).

As an aside, it may be next week before I can fully review the other PR you created.

Thanks,
Curtis

Issue #529, opened 2019-03-07
Cannot open a project.


chopin2712:
Dear !

I recently discover this tool, it is very helpful but  I cannot re-open a recent project (on ubuntu 18.10) 

My project is this one : http://www.mediafire.com/file/5dhgch7wyhagj18/WALL-E+est+un+chef-d%27oeuvre.msk

But this project works : 

http://www.mediafire.com/file/6m6zwf3e26pfwiw/d.msk/file

Can you help me ?

gedakc:
When I try to open the project in the first link I see the following in the console log:

```
$ bin/manuskript 
Debug: Web rendering engine used: QWebView
Running manuskript version 0.8.0.
Found translation in settings: 
Note: No translator found or loaded from system locale for locale en_CA.
Last accessed directory "/home/gedakc/tmp" loaded.
Loading: /home/gedakc/tmp/WALL-E est un chef-doeuvre.msk
Detected file format version: 1. Zip: True.
Traceback (most recent call last):
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/welcome.py", line 163, in openFile
    self.mw.loadProject(filename)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/mainWindow.py", line 565, in loadProject
    self.loadDatas(project)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/mainWindow.py", line 792, in loadDatas
    errors = loadSave.loadProject(project)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/loadSave.py", line 66, in loadProject
    v1.loadProject(project, zip=isZip)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 883, in loadProject
    appendRevisions(mdl, root)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 964, in appendRevisions
    appendRevisions(mdl, child)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 964, in appendRevisions
    appendRevisions(mdl, child)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 969, in appendRevisions
    ID = root.attrib["ID"]
[...]

PR #531, opened 2019-03-07
Directory entries in ZIP break loading code

worstje:
While tackling issue #529, I stumbled across the odd behaviour that re-compressing the archive with 7-Zip broke what should be a valid Manuskript project.

After investigation it turned out that the code that loads the texts sensibly expects there to only be files tracked in the files dictionary.

It is completely valid for a zip file to contain entries describing the contained directories. The logical fix is to simply avoid adding these directory entries to our files dictionary in the first place.

I have verified the fix against the project in issue #529 by having Manuskript manually save the unzipped version as a single file, and comparing the files it tries to load against a 7-Zip compressed project. They load exactly the same files with this patch applied.

Throwing it at my own project worked completely fine as well, so I am confident this is fixed and won't cause any data to be lost.

### DIFF
--- a/manuskript/load_save/version_0.py
+++ b/manuskript/load_save/version_0.py
@@ -178,7 +178,10 @@ def loadFilesFromZip(zipname):
     zf = zipfile.ZipFile(zipname)
     files = {}
     for f in zf.namelist():
-        files[os.path.normpath(f)] = zf.read(f)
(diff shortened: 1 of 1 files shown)


gedakc:
Thank you @worstje for this enhancment.

I have successfully tested this PR on Debian 8, Kubuntu 16.04, and Window XP.  In each test Manuskript loaded and saved projects with no issue and no regressions were observed.

Issue #527, opened 2019-03-04
Fullscreen editor suggestions


worstje:
I finally got around to writing a couple of chapters in Manuskript in the past few days while also figuring out what works best for me. And based on that I have a few requests / suggestions which will cause me to spam some issues. Feel free to shut any of them down; they are just what I think is handy or convenient.

This particular issue concerns my experience with the fullscreen editor mode.

## Markup colors

I already mentioned a related topic in one of the other issues I opened today, but I cannot adjust the color of bracketed spans to match my editor colors, making some of them hard to read unless I change my editor to look different from what I would find most convenient.

## Lock Screen panel on left is unreadable

No doubt this is because of my custom color theme as I can read it when I switch to another one, but I am not sure which colors it uses. It uses a combination of colors that don't end up on top of eachother in the editor, at least.

(I don't quite know how this featureworks, either... the 'lock' name scares me a bit, haha! 🤣 I'll experiment with that another day!)

## Top hover bar only has the spell check... what a waste of space.

I feel like more could be done with the top hover bar. For example, have the remainder of the empty space show the text name. Or have it show the one-line summary for reminding purposes. Or maybe both. (Or just make it configurable in the Settings -> Views screen somewhere like is done for so many other views!)

## No easy way to check text summary while writing.

I would not mind a button in the top bar that would toggle the usage of the free space on the right side to list the multiline summary of that text (which I use to briefly describe the plot points I need to touch on while actually writing a text). It would be a very useful thing indeed.

Right now, I find myself leaving the full screen mode far too much to make this happen. (And in the process I get slammed in the face by all the brightness of the whitish interface colors. 🤣 )

gedakc:
@worstje, were you aware that @katafrakt was actively working in this area?

See Issue #234 and Pull Request #509.

Your testing and comments could go a long way to helping @kakaroto implement fullscreen improvements.

worstje:
No @gedakc , I wasn't aware at all. I have spent almost all my time the last few weeks at importing all my old chapters into Manuskript, setting up a way to publish them in a format I can use and only just now got to finally use the program for the actually productive parts of it. 😁 

My primary aim for posting all these issues was something you said several weeks ago regarding the importing functionality in particular, which was to essentially share my experiences. I figured that no matter what part of the program it is, whatever thoughts I have along the way could give the developers some insights as to my usage patterns and perhaps help identify bits that need love, even if it is not in the way I suggest.

[...]

PR #539, opened 2019-03-27
Fix color scheme of fullscreen editor

kakaroto:
Color scheme was inverted if foreground was transparent, bug and fix provided
by @tildagail [1] and the text on the left side panel was ignoring text color
settings, making it unreadable in dark themes, as reported by @worstje in #527

[1] https://github.com/olivierkes/manuskript/issues/527#issuecomment-469578130


gedakc:
Thank you @kakaroto for implementing the patch suggested by @tildagail.  The patch looks good to me so I will rebase and merge this for inclusion in the next release of Manuskript.

PR #540, opened 2019-03-28
Todo

Temmon:
I was wanting an inline Todo marker, since I do that all the time. I had been using <> because that highlit the text, but with the latest version, it turns out that marks links and it was pretty disruptive. So I added my own flags. I didn't see any official markdown flags to use that weren't stretching their use too far, so I created my own. You can denote any block of text that isn't interrupted by a line break with %t SOMETHING TO REMEMBER % and it'll highlight it in gray. Further, I added a new dialog, like the frequency analyzer, that'll mark out all your todos for you along with the file it's found in. I'm going to be using it myself, so I thought I'd offer if anyone else was interested. I'm fine with any changes.


gedakc:
Were you aware of HTML comments as shown in the screen shot below?

Manuskript-Editor-Comments

Do these HTML comments work for your purposes?

I am hesitant to add new tags especially when tags such as % are used to encode other characters.  See [Percent-encoding](https://en.wikipedia.org/wiki/Percent-encoding).  Additionally each new regex we add to Manuskript slows it down.

Temmon:
No problem.

gedakc:
The closing of this PR with no discussion surprised me.

Does this mean that HTML comments work for you?

The concept of a TODO list is a good idea.  My hesitation arises from introducing a new tag that may cause compatibility issues with HTML and MarkDown (and any future markup text that Manuskript may support such as ASCIIDOC).

I know of one other Integrated Development Environment called Eclipse that implements a TODO list, and it implements the list using the existing programming language commenting mechanism.  For example see [Eclipse Task Managment with TODO in comment](https://www.tutorialspoint.com/eclipse/eclipse_task_management.htm).  Such a feature might be implemented in Manuskript similar to the existing **Frequency Analyzer**.

Temmon:
Sorry. My brain's apparently not working this morning. I just tried the HTML comments and it seems like it'll work fine. I had staged it in two separate comments to make it pretty easy to separate the inline notation changes from the TODO list feature and I definitely understand why you'd hesitate to add more custom markup. I'd forgotten about percent encoding.

Do you think adding an additional flag is valuable beyond the HTML comment for marking which will be displayed in the dialog? `<!-- I need to do this -->` or `<!-- TODO I need to do this -->` I feel like the latter is getting into very wordy territory, but it does make things obvious.

gedakc:
If a person types additional text marker, such as `TODO` followed by the task description and then **Ctrl+D**, then one can already get a list of the texts/scenes that contain this marker using existing Manuskript **Search** functionality.

Manuskript-TODO-Search

Note that there is an open issue for **Search** to highlight the word marker and not just list the text/scene.

Does that help?

Temmon:
I don't love that, because you then have to search through the document for all todos. Since I like to note out little things like words to lookup as well as for large items like dialog, that can get cumbersome in a larger file. I'd already developed this dialog, based off the frequency analyzer that's almost independent of what tokens are used to denote todos (you just need to specify a regex to match) that shouldn't add significantly to bloat because it's only run when requested. 

tododialog

gedakc:
> Do you think adding an additional flag is valuable beyond the HTML comment for marking which will be displayed in the dialog? `<!-- I need to do this -->` or `<!-- TODO I need to do this -->`

If listing all comments is sufficient then I'd say the `TODO` is not required.  In that case, perhaps the dialog and menu items should say "Comment List" and/or "List Comments" instead of "Todo".

PR #558, opened 2019-04-21
Fixed #549 and refactored the image tooltip code

### LINKED ISSUE #549: Image crash: When using tooltip on an incomplete image filename
If i insert an image file as a markdown link, and the image is empty, program crashes, example:

`image`

When using the mouse over the link, the program closes.

worstje:
Issue #549 was caused because the request and response objects are not garantueed to be the same. Redirects are the most common cause, but a malformed URL apparently also qualifies. We now make sure to look at the original request.

The code confused me a bit while I was working on it, and I decided to refactor and document it so I could understand what was going on. I am glad I did: I stumbled across another bug involving the rapid-firing of tooltip requests, and the processing dict never had its entries removed either, leading to a (very slow) memory leak.

### DIFF
--- a/manuskript/ui/views/MDEditView.py
+++ b/manuskript/ui/views/MDEditView.py
@@ -527,7 +527,7 @@ class MDEditView(textEditView):
                   +"<p><img src='data:image/png;base64,{}'></p>")
             tooltip = None
             pos = event.pos() + QPoint(0, ct.rect.height())
-            imageTooltiper.fromUrl(ct.texts[2], pos, self)
+            ImageTooltip.fromUrl(ct.texts[2], pos, self)
 
         elif ct.regex == self.inlineLinkRegex:
             tooltip = ct.texts[1] or ct.texts[2]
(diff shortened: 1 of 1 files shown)


gedakc:
Thank you @worstje for fixing the issue #549 crash.  In my testing your code enhancement did indeed prevent the issue #549 crash.

Good catch on lower case letter starting the name of a class ("**i**mageTooltiper" versus "**I**mageTooltip").  I also appreciate the addition of comments.

When reviewing the code I came across one spelling mistake  in the commit message ("garant**u**eed" instead of "guaranteed").

Would you like to fix the spelling mistake before I commit this PR?

worstje:
That word gets me every single time. No matter how many times I check and re-check, I always end up forgetting and doubting which version I need. And the one time I think I do know... I get it wrong again.

Fixed. Sorry about that.

gedakc:
Thanks for the update.  I'll do a quick retest, and after the Travis CI processes completes I will plan to merge this PR.

PR #570, opened 2019-05-02
App Bundle as the packaging target for macOS

### LINKED ISSUE #567: Package as App Bundle for macOS
It would be nice to package the app as an App Bundle so that it can live in the Applications folder and be added to the dock.
I did some research on how to achieve this with PyInstaller. The changes in [this](https://github.com/dschanoeh/manuskript/commit/2574772cfd06baca4fcb6a95a69102a4f371b057) commit seem to do the trick. A Manuskript.app is created and zipped ready for deployment.

I noticed that the icon doesn't display in Finder but before investigating further: Is that something that can be added and is the integration going in the right direction?

dschanoeh:
This modifies the packaging process such that for macOS, an App Bundle is generated instead of a regular folder.
Additionally, an .icns icon set is added so that the App Bundle can use it.

Fixes #567.


manuskript.spec
```diff
@@ -38,3 +49,11 @@ coll = COLLECT(exe,
                strip=False,
                upx=True,
                name='manuskript')
+
+# also build an app bundle on macOS
+if platform == 'mac':
```
worstje: None of the existing build scripts in this file do any platform detection. Better yet, the different actions taken appear to be platform agnostic despite targeting the creation of a Windows Exe, among other things. Can the platform check be removed entirely, allowing for the bundle be generated on other machines? Or is there something preventing this?

I do not believe any of the maintainers have a Mac, so if building the app bundle requires a mac, it would be less than ideal when they want to release a new version.

manuskript.spec
```diff
@@ -38,3 +49,11 @@ coll = COLLECT(exe,
...
+    app = BUNDLE(coll,
+        name='Manuskript.app',
+        icon=os.path.join(SPECPATH, 'icons/Manuskript/manuskript.ico'),
```
worstje: You probably want the Mac-specific icon here. (But you took care of that in another commit.)

I know @gedakc is a fan of having a single commit with the final product before merging the pull request, so you may want to look into consolidating these commits.

gedakc: For a change such as this a single commit is preferred.  As it currently stands the set of commits add a `manuskript-macos.spec` file and then effectively delete the file and merge with the existing `manuskript.spec` file.  This makes it harder to troubleshoot issues later on.

manuskript.spec
gedakc: Correct, none of the current maintainers have a Mac.

@dschanoeh does this PR change the file name of the macOS X package?  
If so then we will need to handle the change in Travis CI.

In addition to other things, Travis CI is used to build the macOS X package which is controlled by the [.travis.yml](https://github.com/olivierkes/manuskript/blob/develop/.travis.yml) file.  Note that the .travis.yml file uploads the `dist/manuskript-osx-develop.zip` file to our web server because all files in Travis CI disappear after Travis CI has run.

The original Manuskript author @olivierkes also set up Semaphore CI, which I believe uploads to [Github Releases](https://github.com/olivierkes/manuskript/releases) but unfortunately we do not have access to change how that works.  All we can do there is make a request to @olivierkes, or create work-arounds.

manuskript.spec
```diff
@@ -1,4 +1,15 @@
 # -*- mode: python -*-
+import distutils.util
+
+COMPILING_PLATFORM = distutils.util.get_platform()
+if COMPILING_PLATFORM == 'win-amd64':
+    platform = 'win'
+    STRIP = False
+elif COMPILING_PLATFORM == 'linux-x86_64':
+    platform = 'nix64'
+    STRIP = True
```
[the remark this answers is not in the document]
dschanoeh: I found some generic code to do platform detection and thought it would be useful to add this if anyone wants to use it in the future. I'm happy to remove the other platforms that aren't needed.

manuskript.spec
dschanoeh: As far as I understand, the Mac package gets built in a macOS environment by Travis CI, right? In the end, there will be Mac specific binaries (e.g. the QT library) in the bundle. If we remove the platform check here, you might be able to build an App Bundle (which really is just a folder with defined structure) on e.g. Windows but I'm assuming the resulting bundle wouldn't run on the Mac.
I understand that it would be cleaner to not have a check but I think you'd always be building an additional folder on Windows/Linux that wouldn't be useful to anybody.

Regarding the capital "M" in the bundle name: It is common on the Mac to use title case for app file names. For example Wireshark's app is called 'Wireshark.app' even though the command line binary is 'wireshark'. I've seen capital 'Manuskript' used everywhere on the home page and therefore thought that would be consistent. Let me know if that needs to change.

manuskript.spec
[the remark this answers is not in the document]
gedakc: In my opinion code that isn't used will unnecessarily add to complexity.  Further since the check for macOS seems to only be used to build an app bundle on macOS, it makes sense to place this check at that section of the code.

manuskript.spec
gedakc: Correct.  The mac package gets built in a macOS environment by the Travis CI.  Note that the Travis CI also runs for GNU/Linux so code in the `.travis.yml` file must take this into account.

I am okay either with or without the capital "M" in "Manuskript.app".  The key point is that a change in filename must also be reflected in the code that runs in Travis CI.  We also have to make corresponding changes to the [Manuskript Download](https://www.theologeek.ch/manuskript/download/) page.

gedakc:
I've been thinking about this PR and would like to share my thoughts.

First, thank you @dschanoeh for working on this enhanced package for macOS X users.

Second, thank you @worstje for reviewing this PR and providing your comments.

Regarding the upper or lowercase letter "M" in "Manuskript.app", I think it is best to stay consistent with how "manuskript" is invoked from the command line in all of the supported operating systems.  Currently we use lowercase "m", as in `manuskript`, when invoking Manuskript from the command line.  If the uppercase "M" in "Manuskript.app" only affects the name of a menu in macOS X, then it is probably okay to be uppercase.  Because I am not familiar with macOS X I really can't provide a definitive answer.

Regarding changes to the macOS X package, currently none of the developers of Manuskript have access to macOS X.  The existing PyInstaller package has been tested and proven to work when @katafrakt had access to a Mac.  In order for us to maintain the macOS X package, any change to the macOS X package must include not only the PyInstaller build, but also the Travis CI build steps, and the file transfer to a non-TravisCI location.  The package also needs to also be downloaded from the non-TravisCI location and tested to be sure it works.

Because none of the developers of Manuskript have access to macOS X, this PR will have to wait for someone with time and resources to see this PR through the above steps.

TheJackiMonster:
I think this can be closed now that #1056 is merged.

Issue #559, opened 2019-04-21
Crash in Windows 10 when drag and drop


Phieldar:
In Windows 10, latest version of Manuskript. I try dragging and dropping one or many folder (including scenes) in another folder and it crashs. I tried with one folder... and many folders and still it crash. In fact. When I drag and drop individual folder. The first transfer don't crash, but the second crash.

worstje:
Before I dig into it and chase my own tail trying to figure out how to reproduce it, can you offer any more information?

Does it work with an empty folder?
Are there nested folders?
How many files are inside?
What types of files?

Phieldar:
In "plan" mode, I have 4 folders (chapters) including 4 scenes. I created a new folder and wanted to move all four chapter in that new folder that I wanted to name Book 1, (inspired by the trilogy preset). I selected first the 4 folders at once and dragged them in the new folder (5th in the list) and it crashed. I  tried the same but on folder (chapter) at a time... and after the 2nd, it crashed again.

worstje:
I have noticed a crash once in the past when I was organising way back when, but I wasn't particularly interesting in figuring out the cause back then.

Anyway, I have tried moving around some folders in a test project just now and I can't seem to reproduce it though, even when following your instructions. (I assume you mean the outline screen when you talk about 'plan' mode).

Two more questions that come to mind:

1. Are you saving to a single file? (See settings menu.)

2. Do you have auto-saving turned on? If so, do you have revisions turned on? The latter slows down saving considerably and may make it more likely to trigger.

My guess is that you might be noticing some sort of race condition / inconsistency bug that is caused by the auto-save feature saving while the UI is trying to move files about when other parts of the code aren't expecting those bits to change. I've noticed things like those pop up once or twice in the past.

Phieldar:
I'm sorry to be so unprecise. I try to be clearer in the future. 
1. Indeed. I'm saving in a single file
2. I do have auto-saving turned on and also revisions (+ intelligent supression)

I disabled revisions and now the drag and drop is working just fine. Thanks a lot! I believe that I can reactivate revision now.

worstje:
I am currently digging into the saving and outline drag-n-drop code, and I am not quite sure how those two interact as of yet. I did manage to cause a crash while messing around so far but it does not seem to involve saving functionality.

[...]

Issue #479, opened 2019-02-08
Crash when deleting folder with files in tree view


madeonearth:
**Manuskript 0.8.0 Win 10 64 4GB**

Manuskript crashes when you try to delete a folder in treeview that contains files.

> C:\Manuskript\080>manuskript.exe
> Debug: Web rendering engine used: QWebView
> Running manuskript version 0.8.0.
> Found translation in settings: C:\Manuskript\080\i18n\manuskript_de.qm
> Loaded translation: C:\Manuskript\080\i18n\manuskript_de.qm.
> Loading: C:/Manuskript/080/test.msk
> Detected file format version: 1. Zip: False.
> Project C:/Manuskript/080/test.msk loaded.
> Project test.msk saved.
> Project test.msk saved.
> Project test.msk saved.
> Traceback (most recent call last):
>   File "manuskript\models\abstractModel.py", line 125, in parent
>   File "manuskript\models\abstractItem.py", line 124, in row
> ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05D7B870> is not in list

gedakc:
Would you be able to provide a list of steps to recreate the problem?

So far I have been unable to recreate the crash.

I have tried the following:

1. Create a new Manuskript project using the **Novel** template
2. Select the **Outline** pane
3. Select **Chapter 2**
4. Click the red minus sign icon near the bottom of the pane
5. Click **Yes** when asked to confirm "You're about to delete 1 item(s)"

In my case I noted that **Chapter 2** and the scenes within were all deleted.  There was no crash.

madeonearth:
Following your steps at first I wasn't able to recreate myself.
I was allready starting to get nervous but finally after I changed the view from 'index cards' to 'text' with the bottom cmdButtons I was able to reproduce the crashes.
I noticed different values for the Error:


ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05D50C30> is not in list
ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05D42C90> is not in list
ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05D81AD0> is not in list
ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05D43B90> is not in list
ValueError: <manuskript.models.outlineItem.outlineItem object at 0x05C8B610> is not in list

While I am at it what is the tab button 'open selected items in that view' next to the 'split' button for? In whatever state of Manuskript I use it I can not note any difference!?

[...]

Issue #516, opened 2019-02-27
editor/cork options wrong after deleting a text


kakaroto:
Minor bug:
To reproduce, select a text scene in the editor, and click the '-' button to remove it. The main editor window will switch to its parent (the folder) but the editor options beneath it don't change, so we can't switch from text/corkview/outline modes, and we have the option to go fullscreen on the folder even though we shouldn't.
![image](https://user-images.githubusercontent.com/27990/53460630-c115e680-3a0b-11e9-96f9-fb253bd948c6.png)

gedakc:
This issue has been resolved in the **develop** branch and will be included in the next release of Manuskript.

PR #571, opened 2019-05-02
Fix occasional crashes when (re)moving items

worstje:
Describing all the rabbitholes that I and kakaroto have gone through
while debugging this one until dawn can frankly not do enough justice to
the crazy amount of rubberducking that went on while trying to fix this.

This bug would be triggered whenever you had a document open in the
editor and then moved an ancestor object downwards (visually) in the tree.
Or when you simply deleted the ancestor. Depending on the exact method
that caused the opened item to be removed from the internal model, the
exact nature of the bug would vary, which means this commit fixes a few
different bits of code that lead to what appears to be the same bug.

In order of appearance, the bugs that ruined our sleep were:

1) The editor widget was trying to handle the removed item at too late a
stage.

2) The editor widget tried to fix its view after a move by searching for
the new item with the same ID, but in the case of moving an object down
it came across its own old item, ruining the attempt.

3) The editor widget did not properly account for the hierarchical
nature of the model.

Upon fixing these the next day, it was revealed that:

4) The outlineItem.updateWordCount(emit=False) flag is broken. This
function would call setData() in several spots which would still cause
emits to bubble through the system despite emit=False, and we simply got
lucky that it stopped enough of them until now.

This last one was caused by a small mistake in the fixes for the first
three bugs, but it has led to a couple of extra changes to make any
future bug hunts slightly less arduous and frustrating:

a) When calling item.removeChild(c), it now resets the associated parent
and model to mirror item.insertChild(c). This has also led to an extra
check in model.parent() to check for its validity.

b) The outlineItem.updateWordCount(emit=) flag has been removed entirely
and it now emits away with reckless abandon. I have been unable to
reproduce the crashes the code warned about, so I consider this a code
quality fix to prevent mysterious future issues where things sometimes
do not properly update right.

Worthy of note is that the original code clearly showed the intention to
close tabs for items that were removed. Reworking the editor to support
closing a tab is unfortunately way out of scope, so this intention was
left in and the new fix was structured to make it trivial to implement
such a change when the time comes. An existing FIXME regarding unrelated
buggy editor behaviour was left in, too.

Many thanks to Kakaroto for burning the midnight oil with me to get to
the bottom of this. (I learned a lot that night!)

Issues #479 and #559 are fixed by this commit. And maybe some others,
too.

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -141,6 +141,9 @@ class abstractItem():
         @return: the removed abstractItem
         """
         r = self.childItems.pop(row)
+        # Disassociate the child from its parent and the model.
+        r._parent = None
+        r.setModel(None)
         return r
 
     def parent(self):
--- a/manuskript/models/abstractModel.py
+++ b/manuskript/models/abstractModel.py
@@ -92,9 +92,15 @@ class abstractModel(QAbstractItemModel):
         """
         return self.rootItem.findItemsContaining(text, columns, mainWindow(), caseSensitive)
 
-    def getItemByID(self, ID):
+    def getItemByID(self, ID, ignore=None):
+        """Returns the item whose ID is `ID`, unless this item matches `ignore`."""
+
         def search(item):
             if item.ID() == ID:
+                if item == ignore:
+                    # The item we really want won't be found in the children of this
+                    # particular item anymore; stop searching this branch entirely.
+                    return None
                 return item
             for c in item.children():
                 r = search(c)
@@ -104,9 +110,12 @@ class abstractModel(QAbstractItemModel):
         item = search(self.rootItem)
         return item
 
-    def getIndexByID(self, ID, column=0):
-        "Returns the index of item whose ID is `ID`. If none, returns QModelIndex()."
-        item = self.getItemByID(ID)
+    def getIndexByID(self, ID, column=0, ignore=None):
+        """Returns the index of item whose ID is `ID`. If none, returns QModelIndex().
+
+        If `ignore` is set, it will not return that item if found as valid match for the ID"""
+
+        item = self.getItemByID(ID, ignore=ignore)
         if not item:
             return QModelIndex()
         else:
@@ -119,7 +128,10 @@ class abstractModel(QAbstractItemModel):
         childItem = index.internalPointer()
         parentItem = childItem.parent()
 
-        if parentItem == self.rootItem:
+        # Check whether the parent is the root, or is otherwise invalid.
+        # That is to say: no parent or the parent lacks a parent.
+        if (parentItem == self.rootItem) or \
+           (parentItem is None) or (parentItem.parent() is None):
             return QModelIndex()
 
         return self.createIndex(parentItem.row(), 0, parentItem)
--- a/manuskript/ui/editors/editorWidget.py
+++ b/manuskript/ui/editors/editorWidget.py
@@ -293,16 +293,14 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
 
         try:
             self._model.dataChanged.connect(self.modelDataChanged, AUC)
-            self._model.rowsInserted.connect(self.updateIndexFromID, AUC)
-            self._model.rowsRemoved.connect(self.updateIndexFromID, AUC)
-            #self.mw.mdlOutline.rowsAboutToBeRemoved.connect(self.rowsAboutToBeRemoved, AUC)
+            self._model.rowsAboutToBeRemoved.connect(self.rowsAboutToBeRemoved, AUC)
         except TypeError:
             pass
 
         self.updateStatusBar()
 
     def setCurrentModelIndex(self, index=None):
-        if index.isValid():
+        if index and index.isValid():
             self.currentIndex = index
             self._model = index.model()
             self.currentID = self._model.ID(index)
@@ -313,17 +311,26 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
         if self._model:
             self.setView()
 
-    def updateIndexFromID(self):
+    def updateIndexFromID(self, fallback=None, ignore=None):
         """
         Index might have changed (through drag an drop), so we keep current
         item's ID and update index. Item might have been deleted too.
-        """
-        idx = self._model.getIndexByID(self.currentID)
 
-        # If we have an ID but the ID does not exist, it has been deleted
+        It will ignore the passed model item to avoid ambiguity during times
+        of inconsistent state.
+        """
+        idx = self._model.getIndexByID(self.currentID, ignore=ignore)
+
+        # If we have an ID but the ID does not exist, it has been deleted.
         if self.currentID and idx == QModelIndex():
-            # Item has been deleted, we open the parent instead
-            self.setCurrentModelIndex(self.currentIndex.parent())
+            # If we are given a fallback item to display, do so.
+            if fallback:
+                self.setCurrentModelIndex(fallback)
+            else:
+                # After tab closing is implemented, any calls to `updateIndexFromID`
+                # should be re-evaluated to match the desired behaviour.
+                raise NotImplementedError("implement tab closing")
+
             # FIXME: selection in self.mw.treeRedacOutline is not updated
             #        but we cannot simply setCurrentIndex through treeRedacOutline
             #        because this might be a tab in the background / out of focus
@@ -337,19 +344,39 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
             self.setView()
 
     def modelDataChanged(self, topLeft, bottomRight):
-        # if self.currentID:
-        # self.updateIndexFromID()
-        if not self.currentIndex:
-            return
+        if not self.currentIndex.isValid():
+            return  # Just to be safe.
+
+        # We are only concerned with minor changes to the current index,
+        # so there is no need to call updateIndexFromID() nor setView().
         if topLeft.row() <= self.currentIndex.row() <= bottomRight.row():
+            self.updateTabTitle()
             self.updateStatusBar()
 
-    #def rowsAboutToBeRemoved(self, parent, first, last):
-        #if self.currentIndex:
-            #if self.currentIndex.parent() == parent and \
-                                    #first <= self.currentIndex.row() <= last:
-                ## Item deleted, close tab
-                #self.mw.mainEditor.tab.removeTab(self.mw.mainEditor.tab.indexOf(self))
+    def rowsAboutToBeRemoved(self, parent, first, last):
+        if not self.currentIndex.isValid():
+            return  # Just to be safe.
+
+        # Look for a common ancestor to verify whether the deleted rows include our index in their hierarchy.
+        childItem = self.currentIndex
+        ancestorCandidate = childItem.parent()  # start at folder above current item
+        while (ancestorCandidate != parent):
+            childItem = ancestorCandidate
+            ancestorCandidate = childItem.parent()
+
+            if not ancestorCandidate.isValid():
+                return  # we ran out of ancestors without finding the matching QModelIndex
+
+        # My sanity advocates a healthy dose of paranoia. (Just to be safe.)
+        if ancestorCandidate != parent:
+            return  # we did not find our shared ancestor
+
+        # Verify our origins come from the relevant first..last range.
+        if first <= childItem.row() <= last:
+            # If the row in question was actually moved, there is a duplicate item
+            # already inserted elsewhere in the tree. Try to update this tab view,
+            # but make sure we exclude ourselves from the search for a replacement.
+            self.updateIndexFromID(fallback=parent, ignore=self.currentIndex.internalPointer())
 
     def updateStatusBar(self):
         # Update progress
@@ -359,7 +386,7 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
         if not mw:
             return
 
-        mw.mainEditor.updateStats()
+        mw.mainEditor.tabChanged()
 
     def toggleSpellcheck(self, v):
         self.spellcheck = v
--- a/manuskript/models/outlineItem.py
+++ b/manuskript/models/outlineItem.py
@@ -178,13 +178,11 @@ class outlineItem(abstractItem):
 
     def removeChild(self, row):
         r = abstractItem.removeChild(self, row)
-        # Might be causing segfault when updateWordCount emits dataChanged
-        self.updateWordCount(emit=False)
+        self.updateWordCount()
         return r
 
-    def updateWordCount(self, emit=True):
-        """Update word count for item and parents.
-        If emit is False, no signal is emitted (sometimes cause segfault)"""
+    def updateWordCount(self):
+        """Update word count for item and parents."""
         if not self.isFolder():
             setGoal = F.toInt(self.data(self.enum.setGoal))
             goal = F.toInt(self.data(self.enum.goal))
@@ -219,12 +217,11 @@ class outlineItem(abstractItem):
             else:
                 self.setData(self.enum.goalPercentage, "")
 
-        if emit:
-            self.emitDataChanged([self.enum.goal, self.enum.setGoal,
-                                  self.enum.wordCount, self.enum.goalPercentage])
+        self.emitDataChanged([self.enum.goal, self.enum.setGoal,
+                              self.enum.wordCount, self.enum.goalPercentage])
 
         if self.parent():
-            self.parent().updateWordCount(emit)
+            self.parent().updateWordCount()
 
     def stats(self):
         wc = self.data(enums.Outline.wordCount)
--- a/manuskript/ui/views/outlineBasics.py
+++ b/manuskript/ui/views/outlineBasics.py
@@ -383,7 +383,7 @@ class outlineBasics(QAbstractItemView):
 
         parentItem.childItems.insert(index.row() + delta,
                                      parentItem.childItems.pop(index.row()))
-        parentItem.updateWordCount(emit=False)
+        parentItem.updateWordCount()
 
     def moveUp(self): self.move(-1)
     def moveDown(self): self.move(+1)


manuskript/models/abstractItem.py
```diff
@@ -141,6 +141,9 @@ def removeChild(self, row):
         @return: the removed abstractItem
         """
         r = self.childItems.pop(row)
+        # Disassociate the child from its parent and the model.
```
kakaroto: Dissociate instead of disassociate ?

manuskript/models/abstractModel.py
```diff
@@ -92,9 +92,15 @@ def findItemsContaining(self, text, columns, caseSensitive=False):
...
-    def getItemByID(self, ID):
+    def getItemByID(self, ID, ignore=None):
+        """Returns the item whose ID is `ID`, unless this item matches `ignore`."""
+
         def search(item):
             if item.ID() == ID:
+                if item == ignore:
+                    # The item we really want won't be found in the children of this
+                    # particular item anymore; stop searching this branch entirely.
+                    return None
```
kakaroto: The comment here is misleading, the 'ignore' only ignores if the item with the right ID matches the 'ignore' value, the comment says it skips any items that are children of the item to ignore.

manuskript/models/abstractModel.py
```diff
@@ -104,9 +110,12 @@ def search(item):
         item = search(self.rootItem)
         return item
 
-    def getIndexByID(self, ID, column=0):
-        "Returns the index of item whose ID is `ID`. If none, returns QModelIndex()."
-        item = self.getItemByID(ID)
+    def getIndexByID(self, ID, column=0, ignore=None):
+        """Returns the index of item whose ID is `ID`. If none, returns QModelIndex().
+
+        It will ignore the given item as valid match for the ID."""
```
kakaroto: I'd say "If `ignore` is set, will not return that item if found as valid match for the ID"

manuskript/models/abstractModel.py
```diff
@@ -122,6 +131,9 @@ def parent(self, index=QModelIndex()):
         if parentItem == self.rootItem:
```
kakaroto: `if parentItem is None or parentItem == self.rootItem:` instead of duplicating the 'return qModelIndex'

manuskript/ui/editors/editorWidget.py
```diff
@@ -293,9 +293,8 @@ def addSpacer():
 
         try:
             self._model.dataChanged.connect(self.modelDataChanged, AUC)
-            self._model.rowsInserted.connect(self.updateIndexFromID, AUC)
-            self._model.rowsRemoved.connect(self.updateIndexFromID, AUC)
-            #self.mw.mdlOutline.rowsAboutToBeRemoved.connect(self.rowsAboutToBeRemoved, AUC)
+            self._model.rowsInserted.connect(self.rowsInserted, AUC)
```
kakaroto: Did you check to see if the index needs to change when a row is inserted or if the old index remains valid since the outlineItem itself doesn't change?

manuskript/ui/editors/editorWidget.py
```diff
@@ -313,17 +312,26 @@ def setCurrentModelIndex(self, index=None):
...
+        idx = self._model.getIndexByID(self.currentID, ignore=ignore)
+
+        # If we have an ID but the ID does not exist, it has been deleted.
         if self.currentID and idx == QModelIndex():
-            # Item has been deleted, we open the parent instead
-            self.setCurrentModelIndex(self.currentIndex.parent())
+            # If we are given a fallback item to display, do so.
+            if fallback:
+                self.setCurrentModelIndex(fallback)
+            else:
```
kakaroto: You don't need the 'if fallback' or the else part, because fallback will always be set to the parent when rows are being removed, and in the insertion case, your current ID will never end up deleted. I'd just put 'fallback' as a non optional argument in this function too.

manuskript/ui/editors/editorWidget.py
```diff
@@ -313,17 +312,26 @@ def setCurrentModelIndex(self, index=None):
...
+                #self.mw.mainEditor.tabSplitter.tab.removeTab(self.mw.mainEditor.tab.indexOf(self))
+                self.setCurrentModelIndex(None) # TODO: change this to close the tab.
```
kakaroto: You'll be removing this since it's unused code, but this would cause a crash too since the first thing setCurrentModelIndex does is "index.isValid()" so passing None to it would crash.

manuskript/ui/editors/editorWidget.py
```diff
@@ -344,12 +352,33 @@ def modelDataChanged(self, topLeft, bottomRight):
...
-                                    #first <= self.currentIndex.row() <= last:
-                ## Item deleted, close tab
-                #self.mw.mainEditor.tab.removeTab(self.mw.mainEditor.tab.indexOf(self))
+    def rowsInserted(self, parent, first, last):
+        if self.currentIndex:
+            self.updateIndexFromID(fallback=self.currentIndex.parent())
+
+    def rowsAboutToBeRemoved(self, parent, first, last):
+        if not self.currentIndex:
+            return
```
kakaroto: What's the use case where currentIndex is None? That doesn't happen, even a 'invalid index' is set in setCurrentModelIndex with `           self.currentIndex = QModelIndex()`
Same for rowsInserted

manuskript/ui/editors/editorWidget.py
```diff
@@ -344,12 +352,33 @@ def modelDataChanged(self, topLeft, bottomRight):
...
+        commonality = crumb.parent() # start at folder above current text
+        while (commonality != parent):
+            crumb = commonality
+            commonality = commonality.parent()
+
+            if commonality == None:
+                return  # we ran out of parents without finding the matching QModelIndex
+
+            if commonality == parent:
+                break  # we found our shared parent
```
kakaroto: Did you confirm if `self._model.rootItem` == a non valid QModelIndex ? Test with moving/deleting scenes that are in the root directly.

manuskript/ui/editors/editorWidget.py
```diff
@@ -344,12 +352,33 @@ def modelDataChanged(self, topLeft, bottomRight):
...
-                #self.mw.mainEditor.tab.removeTab(self.mw.mainEditor.tab.indexOf(self))
+    def rowsInserted(self, parent, first, last):
+        if self.currentIndex:
+            self.updateIndexFromID(fallback=self.currentIndex.parent())
+
+    def rowsAboutToBeRemoved(self, parent, first, last):
+        if not self.currentIndex:
+            return
+
+        # Check parents until we find a parent that matches the parent we are removing.
```
kakaroto: Comment is confusing, maybe rewrite into : 
`# Look for a common parent to verify if the deleted rows include our index in their hierarchy`

manuskript/ui/editors/editorWidget.py
```diff
@@ -344,12 +352,33 @@ def modelDataChanged(self, topLeft, bottomRight):
...
+        crumb = self.currentIndex
+        commonality = crumb.parent() # start at folder above current text
```
kakaroto: I'm not sure about `crumb` and `commonality` as variable names, I'd suggest `childItem` and `commonParent` 
respectively.. I don't like the names I found either but they are less confusing, if you have a better proposal, feel free to suggest it!

manuskript/models/abstractModel.py
worstje: It will skip the children of the item currently being compared against. Which makes sense because the IDs are supposed to be unique... except in the exact situation we are dealing with here. The code will only ever loop through the child items when the ID doesn't match in the first place. The new item cannot be a child of the duplicate item to begin with unless we have a bug elsewhere.

The comment is there exactly because the code itself made me second-guess whether the logic was right. (And your later confusion thinking it was a flat loop as opposed to the recursion it is only added to that.)

If this is not what you mean, can you suggest a better comment?

manuskript/models/abstractModel.py
worstje: I'll change it to that with a small change for better flow: "If `ignore` is set, **_it_** will not return that item if found as valid match for the ID."

manuskript/models/abstractModel.py
worstje: Will change. I used to have debug code inbetween and I didn't realize I could combine them when cleaning up. Also good catch on the 'is None'; I always make that mistake.

manuskript/ui/editors/editorWidget.py
worstje: I left it in because I do not have a 'proper' fix for that bit of code without digging all the way into the editor code. This function can't start trying to dig into the existing item and obtain a parent because we'll end up where we were, and removing it entirely will leave that tab in an inconsistent (and likely crashy) state.

The code as it is currently implemented cannot trigger the else, which is why I kind of left it like that. I can make the else `raise NotImplementedError("implement tab closing")` if you'd prefer that approach? It is probably neater.

(The neatest is to just implement it right now, but scope creep...)

manuskript/ui/editors/editorWidget.py
worstje: Defensive coding against inconsistent state. That check was in the original code for one of the handlers and it made sense to me to have it so I made sure the signal handlers were equal in their check.

Which does not mean the check made sense to begin with if the base state is a QModelIndex.

Question though: the index paramter for setCurrentModelIndex() is by default set to None. I am not quite sure how its check if 'index.isValid()' could work out? I believe that should trigger an AttributeError... but assuming that does not crash, currentIndex can (in the current shape of things) definitely be set to None. So if not that, we may also want to fix that code while we're about this topic.

manuskript/ui/editors/editorWidget.py
worstje: I tried the code with items in the highest root level and I didn't see any problems while cutting or moving them. That said, I have not specifically found a way to see if rootItem can pop up in that situation.

rootItem is defined as a special Outline item with ID 0, by the way. See abstractModel.py. I do not believe we should be able to come across it as a parent because the only way for it to end up that way is if the rest of the code somehow set the root item as a parent... but all the model code checks for rootItem, so that should not happen.

manuskript/ui/editors/editorWidget.py
worstje: I have rewritten it into '_Look for a common ancestor to verify whether the deleted rows include our index in their hierarchy._' to avoid the entire 'parent' word entirely.

manuskript/ui/editors/editorWidget.py
worstje: I took a while to come up with these.

Commonality is because I wanted to stay away from the word parent and child because of the loaded meaning with both `QModelIndex` and our own model heavily using those terms. It made the code easier to read and understand for me to avoid those words. Besides: it is a perfectly fine english word. If you think the word isn't specific enough given that it may not actually be a commonality yet, how about `commonalityCandidate`? xD 

Crumb comes from the idea of a breadcrumb trail; we want to be able to look back at the item we originally came from.

manuskript/models/abstractItem.py
worstje: They are both valid for as far I am able to tell based on [this link](https://grammarist.com/usage/dissociate-disassociate/). To be honest, I didn't even know 'dissociate' was valid english myself until you asked this.

Can change or can keep, I don't really care. For me they are both english, although the form I used obviously makes more sense to me.

manuskript/ui/editors/editorWidget.py
worstje: The intent for the fallback parameter is to specify an alternative view in the case you do not want the default action of the tab being closed. As such the design for that parameter is optional, and changing the function signature so that it is required communicates that the desired behaviour for this function is to always show the fallback when updating the ID. It isn't: existing code clearly shows that closing a tab is the preferred action.

From our discussion last night I realize we are of different minds here; you are an advocate of 'show something else instead' and not very much into the entire 'close the tab' business. :-) However, the original design for the code is to close the tab if the item is removed; the 'show the parent' logic is something this PR adds which definitely has its uses... but beyond the 'existing code says X' issue, I do not believe it is the desired default solution.

For example, take a folder with five texts in it. These five texts are open in the editor on separate tabs because I am working on them. Then I decide I don't want the folder there anymore, so I 'cut' it. The result with the current code will be to have five tabs that all show the root. To me as a user, all those root windows are utterly useless and have nothing to do with what I was doing in them anymore, meaning I will have to close those tabs manually afterwards. IMHO, the fallback option should only be used when you really want to keep the tab open.

My goal here is not to change the default behaviour or intended operating principles of Manuskript, but to fix a bug. The original code (that doesn't quite work) shows that we want to close tabs here, so I structure the code with that intent in mind. The best way to do that right now is by showing the parent instead in all situations where a fallback is needed. But that is not a long term solution. Truly fixing all this is out of scope for this PR, but it will eventually get fixed.

Note that the code in its current form should not trigger the else condition because a fallback is always specified, so I do not see any functional problems with the code as it is. I am fine with changing some implementation details to make this PR neater, but I am quite stubborn on preserving the current function signature and meaning of the fallback parameter.

manuskript/models/abstractItem.py
kakaroto: Interesting read! Never heard disassociate, sounded weird to me. If it's not a typo, then all good! 👍

manuskript/models/abstractModel.py
kakaroto: Humm.. I hadn't thought of deletion of a folder when we have a folder opened in corkview, I was only thinking of a Text open, and a Text obviously can't have children. +1

manuskript/models/abstractModel.py
kakaroto: Good! I like the optimization, and I hadn't realized "is None" acted different from "== None", I just am used to using "is None". Good that you caught the difference :)

manuskript/ui/editors/editorWidget.py
kakaroto: Alright, that sounds good to me. It's a good explanation too, and I agreed with you in our last discussion that it's a difference of opinion, nothing inherently wrong either behaviors, but I don't think it's "important enough" to have that behavior be a settings option for the user. That's why I suggested to open the discussion here, see what @gedakc or @olivierkes thinks. As for "original intent" or "existing code", I disagree. That was the old commented out code, it was maybe the initial implementation that was later removed for a better implementation (which was to not close the tab), but maybe it was a different opinion from a different developer, or maybe it was removed because it kept crashing and they changed the index instead as a temporary fix, or maybe closing a tab was there as an early implementation but then they changed it to something they thought was a better behavior.. I don't know, which is why discussing it would be good, but I just don't accept your argument of "existing code shows that closing a tab is the preferred action."

Either way, while I prefer the tab not to close but switch to the parent, I have no issues with whatever choice is decided. In my case anyway, I usually work with a single tab.

manuskript/ui/editors/editorWidget.py
kakaroto: Yeah, I think if you want to leave it, put it in a comment and raise a NotImplementedError, it's better than leaving it like that because if the else ends up being used a year from now, and it causes a crash, nobody will understand why that happens.

manuskript/ui/editors/editorWidget.py
kakaroto: yeah, that should be `if index and index.isValid()` and the 'if self.currentIndex' is probably older code when it could be set to None.. at least, that's what I'd assume.

manuskript/ui/editors/editorWidget.py
kakaroto: Good idea!

manuskript/ui/editors/editorWidget.py
kakaroto: Yeah, I understood the commonality and crumb meaning, and yes they are valid english words, they are just confusing to me and not very clear on what they are for in this context, that's why I suggested changing them. If you prefer them to remain as is, that's your choice, feel free to ignore my comment :)

manuskript/ui/editors/editorWidget.py
gedakc: When writing I usually work with a single tab too, so I don't have a strong opinion on which way this decision lands.  We can change it later if it becomes a problem.

Regarding original intentions, I don't think we will always know what these were/are.  @olivierkes has not been active in the project for over a year so I don't think we should hold up this PR for @olivierkes' input.

manuskript/ui/editors/editorWidget.py
gedakc: Thank you both @worstje and @kakaroto for tackling code challenges such as the ones addressed by this PR.   When you are comfortable with the code in this PR then please let me know so that I can take another look at it with an eye towards merging it with the **develop** branch.

worstje: @kakaroto Fact is my terms do indeed diverge from the terminology in the rest of the project, so you have a very valid point indeed. 😄 Others will have to understand the code later, so why should I stay stubborn?

But because I want to keep the code understandable and avoid the way 'parent' would be overly represented in that snippet, I have opted for a third option that uses the best of both worlds: `childItem` and `ancestorCandidate`. These should best describe the intent of the code in question. (I have also polished up the comments in this snippet of code to match the same language.)

manuskript/ui/editors/editorWidget.py
worstje: My suspicion is that it remains valid. I cannot reproduce buggy behaviour. But do we want to count on that?

From what I can tell, `model.createIndex()` relies on a row, column and item to create said index. Looking at the manual, it exposes `row()` and `column()` functions for accessing this data. The class also appears as very lightweight in its design and does not receive any updates. Hell, I just spotted this in the manual:

> **Note**: Model indexes should be used immediately and then discarded. You should not rely on indexes to remain valid after calling model functions that change the structure of the model or delete items. If you need to keep a model index over time use a QPersistentModelIndex.

This leads me to believe that inserting an item before a currently stored index would indeed cause the QModelIndex to be invalid. Which makes me wonder: should we not be using `QPersistentModelIndex` instead? About that class, the manual tells us:

> A QPersistentModelIndex is a model index that can be stored by an application, and later used to access information in a model. Unlike the QModelIndex class, it is safe to store a QPersistentModelIndex since the model will ensure that references to items will continue to be valid as long as they can be accessed by the model.

I am not going to make any such change in this patch. The current infrastructure that receives signals is adequate to monitor such changes and apply them where necessary.

That said, I did just notice it fails to update the tab title when we rename an open item. I guess that fix now gets to piggyback along with this patch. 😁

manuskript/ui/editors/editorWidget.py
worstje: Agreed. 👍 

Combined with your suggested change, I have removed all checks for `self.currentIndex` and its validity in the signal handlers given that the validity of a current index is already determined when originally setting the index.

worstje:
@gedakc This PR is all ready for you now with all code review comments from @kakaroto taken care of.

Rebase & squash have been done in the assumption all is good now, but if it isn't, just yell and I'll adjust whatever needs adjusting.

manuskript/ui/editors/editorWidget.py
kakaroto: Well, the non-None-ity of a current index is already determined, but its validity is not. If the current Index refers to the root item, then the index is an invalid QModelIndex as far as I understand it.

manuskript/ui/editors/editorWidget.py
kakaroto: Ah, that's really interesting. I guess we would want to switch to a QPersistentModelIndex instead of a QModelIndex (not in this patch though). And thanks, that does answer my question about the need for rowsInserted.

manuskript/ui/editors/editorWidget.py
worstje: Right, but I also [read the manual](https://doc.qt.io/qt-5/qmodelindex.html#isValid), which states:

> bool `QModelIndex::isValid()` const
>
> Returns true if this model index is valid; otherwise returns false.
>
> A valid index belongs to a model, and has non-negative row and column numbers.

Row nor column will become negative out of nowhere and it will still be a part of the same model. It does not check upper boundaries. Thus, the check is completely pointless to have, which is why I removed it.

Note that this is official Qt documentation, and not just an implementation detail dug up from a source file. If it were an implementation detail, I would have erred on the side of caution, but it literally describes the conditions the function is supposed to check.

kakaroto: Yeah, but the same manual (at the top also says : 
> New QModelIndex objects are created by the model using the QAbstractItemModel::createIndex() function. An invalid model index can be constructed with the QModelIndex constructor. Invalid indexes are often used as parent indexes when referring to top-level items in a model.

In this case, if the editor is set to displaying the "Root", then the currentIndex.isValid() would return False, because to set it to displaying the Root item, we do :

            self.currentIndex = QModelIndex()

And I was replying to your comment that said "I have removed all checks for self.currentIndex and its validity in the signal handlers given that the validity of a current index is already determined when originally setting the index." and my point was that the validity of a current index is not determined because we are originally setting the index to `QModelIndex()` which by definition is invalid. As for whether or not we need to check for the validity in the signal handlers, I'm not sure.

worstje: Good points, I had not spotted that paragraph.

That said, I tried deleting all folders and texts from my model, adding new texts, removing texts again and repeated a few more times in slightly different ways.. and I do not trigger any problems. Even when the root folder displays when performing these actions I see absolutely no errors thrown. So from testing, the current code seems stable.

(I also managed to open up the root as a text document using the context menu... the editor really needs a thorough overhaul... ☹️ )

kakaroto: Yeah, it wouldn't be a problem, you just wouldn't find a common ancestor if you delete something and current index is root. (since root itself can never be deleted) and yes, there are issues like that which might happen with deleting, see #516 for example.

worstje: Right, but that is something I can't do anything with here. To summarize: the signal handlers, with their checks removed, seem to withstand all the tests I can throw at them involving displaying the root. So while I agree that an invalid QModelIndex could be assigned, it does not seem to be any issue. (Or if it actually is an issue, it is indistinguishable from the myriad other editor oddities we could already observe while messing around in the editor. 😬 )

Based purely on my tests, I say the code is fine like this. If we want those checks put back in, just give a shout and I'll put them back in.

worstje: I should really not comment on issues when I am tired, because I clearly didn't understand. 😭 Re-reading while awake, you are clearly in agreement that removing these checks isn't a problem.

As for #516, that was an easy fix to implement while chasing up my sleepy misunderstandings, so I'll incorporate it into this PR. And because I am changing this commit anyways, I might as well implement the stupid check and be done with the matter for safeties sake. 😈

worstje:
All good now (includes a fix for #516 too), unless @kakaroto has even more concerns... 😈

kakaroto:
Lol, yeah, I was worrying about that check not being there, re-reading while awake was a good idea :p I didn't re-review the latest patches but if you fixed all the stuff we discussed, then it should be all good from my side of the review process.

gedakc:
Wow, lots of work done on this PR!

I will start reviewing, which is mostly testing the known crash cases to see if these are now fixed.

gedakc:
Unfortunately I have been unable to confirm that this PR improves the situation for either issue #479 or issue #559.

Any help you can provide regarding how to test this PR would be appreciated.

### Issue #479 Testing

Using Kubuntu 16.04 and Windows 10 with this PR applied, the crash still occurs in my testing.

See [issue 479 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/479#issuecomment-493137783).

### Issue #559 Testing

So far I am still unable to reproduce the crash mentioned in this issue.

kakaroto:
Just confirmed #479 isn't fixed, though the traceback is different this time.

```
  File "v:\Projects\book\manuskript-git\bin\..\manuskript\models\abstractModel.py", line 134, in parent
    return self.createIndex(parentItem.row(), 0, parentItem)
TypeError: createIndex(self, int, int, object: object = 0): argument 1 has unexpected type 'NoneType'
```

gedakc:
### Issue #559 Testing - Take 2

Using Kubuntu 16.04 and Windows 10 with this PR applied, the crash no longer occurs.  :-)

See [issue 559 - steps to reproduce the crash](https://github.com/olivierkes/manuskript/issues/559#issuecomment-493180519).

worstje:
@gedakc Feel free to test the testcase that still remained broken. Hopefully for the last time, as it should now be fixed! 😄 

(I haven't squashed the commit yet because I figure everyone with issues that I reference in my squashed commit must be quite sick of all the notification emails I keep generating for their issues, and I might as well wait for the all-good signal this time around.)

gedakc:
Thanks @worstje for the code enhancement to fix issue 479.

On Kubuntu 16.04 I retested both issue 479 and issue 559 and neither crash occurred.  :-)

I did also try moving various collections of folders up and down and into and outside of other folders and was able to get one of my CPU's to run 100% with the Manuskript GUI becoming unresponsive.  Unfortunately I do not have a series of steps to reliably reproduce this situation.

I eventually used CTRL+C in the terminal window to halt execution.  When I did this I received the following console log:

```
$ bin/manuskript  
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
^CTraceback (most recent call last):
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/editors/editorWidget.py", line 367, in rowsAboutToBeRemoved
    if ancestorCandidate == None:
KeyboardInterrupt
Project d3.msk saved.
Project d3.msk saved.
$
```

It is entirely possible that this issue exists in 0.9.0 as well.  I will try to come up with a reproducible set of steps.

Either way I think that the code in this PR is an improvement.  If you would like to squash the commits, or otherwise edit the commit messages as the first commit does not fix issue 479, then I will merge with the **develop** branch.

As an aside:  On the topic of commits, my personal goal is to ensure that each individual commit compiles, runs, and implements an improvement in the code.  This helps with troubleshooting because it makes using `git bisect` work well to identify regressions.  Also if a set of commits adds some code and then later removes the same code, then it is much cleaner to not add the code in the first place.  Simple fixes should only require one commit.  More complex fixes might require multiple commits that make a logical sequence of changes to the code. These are just general guidelines I try to follow.

gedakc:
Finally I was able to figure out a series of steps to reproduce the GUI hang.

### Test 4 - Multi Chapter Drag and Drop Chapter beside and in other Chapters

Open an terminal window and start manuskript.

1. Create a new Manuskript project using the **Novel** template
2. Select **Editor** pane
3. Select **Chapter 7**
4. Select the **Outline** pane
5. Select **Chapter 7** through **Chapter 9** with shift-click
6. Drag and drop between **Chapter 12** and **Chapter 13**
7. Select **Chapter 7** through **Chapter 9** with shift-click
8. Drag and drop into **Chapter 3**
9. Expand **Chapter 3**
10. Select **Chapter 7** through **Chapter 9** with shift-click
11. Drag and drop between **Chapter 12** and **Chapter 13**

At this time the Manuskript GUI becomes unresponsive.

If I use `top` in another terminal window I discover that one python3 process is using 100% of a single CPU.

Use **Ctrl+C** in the terminal window to regain Manuskript GUI responsiveness.

The console log is as follows:

```
$ bin/manuskript 
Project d3.msk saved.
Project d3.msk saved.
Project d3.msk saved.
^CTraceback (most recent call last):
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/editors/editorWidget.py", line 365, in rowsAboutToBeRemoved
    ancestorCandidate = ancestorCandidate.parent()
KeyboardInterrupt
Project d3.msk saved.
Project d3.msk saved.
```

worstje:
How do you keep managing to break this? I swear I spent ten minutes just dragging folders and texts around after that last commit... I may not be amused, but I am definitely in awe of your testing efforts! 😭 

The loop you got stuck in is one I added and whose result I was paranoid about in the latest commit. Unfortunately that paranoia was based on the assumption it does, in fact, manage to exit.

I think I might know the problem. [QModelIndex.parent()](https://doc.qt.io/qt-5/qmodelindex.html#parent) does not return null (None), but an invalid model index object. You must have somehow managed to cause an update to be processed on an object which was kind of removed from the tree? I'm not sure.

I'll fix it paranoid-style and combine it with a squash. You'll have it shortly. 😄

gedakc:
Thank you @worstje for continuing to investigate and fix these issues.  In my opinion the code is relatively complex and with the asynchronous nature of GUI's and signals, it can be very difficult to debug.

I appreciate both you and @kakaroto delving deep into this area to fix it for all users.  :+1:

worstje:
@gedakc No problem.

Your latest testcase has been reproduced and also fixed with the fix I created before I even saw your testcase. 👍

With that I assume issue 479 is indeed fixed, so I'll leave that mention in the commit message.

(I'm living on the razors edge here by assuming the rabbithole only goes so deep... but I like to live dangerously!)

gedakc:
I've been on both sides of coding and testing so I know how frustrating it can be when the testers keep finding problems.  ;-)

I will re-test using the known steps for producing errors.  When that passes I will try some random folder/text moves to determine if we've finally arrived at a robust solution.

worstje:
I usually try to do the same, although there is substantial evidence that I have failed in that regard where it comes to testing these changes. 

Here is to hoping this is going to be my final comment in this PR. 😄

gedakc:
All of my following tests with the most recent PR 571 applied have been **successful**.  :-)

### Issue 479 Testing

See [issue 479 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/479#issuecomment-493137783).

- Kubuntu 16.04:  SUCCESS.  No crash.
- Windows 10:  SUCCESS.  No crash.

### Issue 516 Testing

See [issue 516 - editor/cork options wrong after deleting a text](https://github.com/olivierkes/manuskript/issues/516) for steps to reproduce problem.

- Kubuntu 16.04:  SUCCESS.  Correct bottom of pane options now displayed (Go to parent item, Text, Index Cards, Outline).
- Windows 10:  SUCCESS.  Correct bottom of pane options now displayed (Go to parent item, Text, Index Cards, Outline).

### Issue 559 Testing

See [issue 559 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/559#issuecomment-493180519).

- Kubuntu 16.04:  SUCCESS.  No crash.
- Windows 10:  SUCCESS.  No crash.

### PR 571 GUI Hang

See [steps to reproduce GUI hang](https://github.com/olivierkes/manuskript/pull/571#issuecomment-493691449).

- Kubuntu 16.04:  SUCCESS.  No GUI hang.
- Windows 10:  SUCCESS.  No GUI hang.

**Great work @worstje and @kararotu!**

After that I will move on to reviewing the backlog of new PR's.

gedakc:
I will merge this PR with the **develop** branch for inclusion with the next release of Manuskript.

Issue #481, opened 2019-02-08
Inconsistent and/or undesirable window placements


worstje:
_A small thing turned kinda big in issue 470, so as requested a separate issue for the observed behaviour ( https://github.com/olivierkes/manuskript/issues/470#issuecomment-460823331 )_

As it turns out, there are a lot of ways to do window placement. The way the QT5 library appears to behave differs considerably across Linux/Windows, and my particular observations were made on Windows 10.

Based on observations using Windows 10, the various windows can be put into a few different groups:

**F6 - Compile:** No respect for screen edges.
**F7 - Import:** No respect for screen edges.
**F8 - Settings:** Centered on top of main window.
**Help -> About:** No respect for screen edges.
**Organize -> Split**: Centered on top of main window.
**Tools -> Frequency Analyser:** Centered on the primary monitor(?)

_These classifications may not necessarily be correct; I basically drew the screen and window diagonals on my screenshots to reach these conclusions._

According to @gedakc, window placement is considerably different on Kubuntu 16.04.

**_This behaviour is not desirable; ideally all the modal / dialog windows act identically in a manner that agrees with the standard behaviours of the operating system in question._**

> 
> 
> **Appendix: Screenshots** _(Windows 10)_
> 
> Of particular note on the Windows 10 screenshots supplied by me in this issue is that I have two monitors: my primary monitor (1) is the larger 2560x1440 one situated on the **right** side, whereas my secondary monitor (2) is the 1920x1080 on the **left** side.
> 
> Also note that the following screenshots were based off the -test2 build at https://github.com/olivierkes/manuskript/issues/470#issuecomment-461125961 ; the fix in it was not appropriate but because the screenshots themselves still demonstrate the various 'approaches' of window placement on Windows 10, I haven't repeated the effort.
> 
> **F6 - Compile:** Cut off at the top.
> ![manuskript-test2-f6](https://user-images.githubusercontent.com/399950/52414811-98be5a80-2ae5-11e9-9b7f-efed36855e4a.png)
> 
> **F7 - Import:** Cut off at the top.
> ![manuskript-test2-f7](https://user-images.githubusercontent.com/399950/52414916-e1761380-2ae5-11e9-9620-fa608b24da82.png)
> 
> **F8 - Settings:** Appears to be centered on top of main window.
> ![manuskript-test2-f8](https://user-images.githubusercontent.com/399950/52414927-ed61d580-2ae5-11e9-8133-92a18695c841.png)
> 
[...]

PR #576, opened 2019-05-09
Fix dialog windows being created outside the desktop area

kakaroto:
The About/Settings/Import/Export/ExportManager windows were all created
in odd places, usually to the left of the main window, which meant outside the
desktop area with little overlap if the main window is maximized. The logic in
centering the window on its parent was wrong. This fixes it.

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -1379,6 +1379,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
     def frequencyAnalyzer(self):
         self.fw = frequencyAnalyzer(self)
         self.fw.show()
+        self.centerChildWindow(self.fw)
 
     ###############################################################################
     # VIEW MENU
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -3,7 +3,7 @@
 import imp
 import os
 
-from PyQt5.QtCore import (pyqtSignal, QSignalMapper, QTimer, QSettings, Qt,
+from PyQt5.QtCore import (pyqtSignal, QSignalMapper, QTimer, QSettings, Qt, QPoint,
                           QRegExp, QUrl, QSize, QModelIndex)
 from PyQt5.QtGui import QStandardItemModel, QIcon, QColor
 from PyQt5.QtWidgets import QMainWindow, QHeaderView, qApp, QMenu, QActionGroup, QAction, QStyle, QListWidgetItem, \
@@ -1068,14 +1068,17 @@ class MainWindow(QMainWindow, Ui_MainWindow):
     # HELP
     ###############################################################################
 
+    def centerChildWindow(self, win):
+        r = win.geometry()
+        r2 = self.geometry()
+        win.move(r2.center() - QPoint(r.width()/2, r.height()/2))
+
     def about(self):
         self.dialog = aboutDialog(mw=self)
         self.dialog.setFixedSize(self.dialog.size())
         self.dialog.show()
         # Center about dialog
-        r = self.dialog.geometry()
-        r2 = self.geometry()
-        self.dialog.move(r2.center() - r.center())
+        self.centerChildWindow(self.dialog)
 
     ###############################################################################
     # GENERAL AKA UNSORTED
@@ -1364,9 +1367,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         self.sw.hide()
         self.sw.setWindowModality(Qt.ApplicationModal)
         self.sw.setWindowFlags(Qt.Dialog)
-        r = self.sw.geometry()
-        r2 = self.geometry()
-        self.sw.move(r2.center() - r.center())
+        self.centerChildWindow(self.sw)
         if tab:
             self.sw.setTab(tab)
         self.sw.show()
@@ -1508,15 +1509,10 @@ class MainWindow(QMainWindow, Ui_MainWindow):
     def doImport(self):
         self.dialog = importerDialog(mw=self)
         self.dialog.show()
+        self.centerChildWindow(self.dialog)
 
-        r = self.dialog.geometry()
-        r2 = self.geometry()
-        self.dialog.move(r2.center() - r.center())
 
     def doCompile(self):
         self.dialog = exporterDialog(mw=self)
         self.dialog.show()
-
-        r = self.dialog.geometry()
-        r2 = self.geometry()
-        self.dialog.move(r2.center() - r.center())
+        self.centerChildWindow(self.dialog)
--- a/manuskript/ui/exporters/exporter.py
+++ b/manuskript/ui/exporters/exporter.py
@@ -3,7 +3,7 @@
 import json
 import os
 
-from PyQt5.QtCore import Qt
+from PyQt5.QtCore import Qt, QPoint
 from PyQt5.QtGui import QBrush, QColor, QIcon
 from PyQt5.QtWidgets import QWidget, QStyle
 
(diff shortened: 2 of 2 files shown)


worstje:
@kakaroto Does this address all the window placement issues listed in #481? If it does, I'll find some time today or tomorrow to test all those scenarios all over again.

kakaroto:
Ah, I hadn't seen that. I missed the frequency analyzer and the split window. I fixed it now for the 
frequency analyzer, but the split dialog is created with `.exec()` which is synchronous and we don't control its position. Either way, it seems to be centering itself on the parent, which is what we want anyway.

worstje:
I can confirm the patch itself works on Windows 10. Windows are now consistently placed in the center of the main window.

One thing I did notice in the patch is that for most of the dialogs, you do the centering **after** showing the window. For the settings window however, the centering happens **before** the window is shown.

Is there a reason you do the centering later for those other windows? While I am on a fast system and unable to notice it, I imagine moving the window after already displaying it could cause a flicker effect or other window management side effects.

Admittedly, we're not playing around with window focus on this occasion, so it shouldn't matter much... but past experience has taught me to get my ducks in a row before showing a window rather than messing around with the window attributes in the no-mans-land that is the asynchronous pleasure of window manager interaction.

kakaroto:
Ah, I hadn't noticed that it was moving the window before its shown for the settings window. I just replaced the existing `move` with the proper algorithm. 
I'm not sure this is 100% valid, but I'd assume we need to call `show()` for it to do size calculations which would give us the right width/height we need to be able to center it. Also, I assume the call to `show()` doesn't actually show the window as that happens asynchronously on the QT event loop to which we'd only return after we finish the function. So I think (though not sure) the show will prepare the window for being shown, calculate its size, etc.. then we move it, then we return to the main event loop which will then display it and trigger the 'exposed' event or something like that.
I'm not sure where to check to confirm that...

worstje:
That is assuming a lot, but then again, my comment was similarly assuming things. 😄 

You are likely right in the fact that Qt handles it asynchronously, but in that case, wouldn't it be some sort of race internally as to whether it resizes before it actually shows the window? I am sure they handle whatever outcomes of that race correctly, but why race if we can avoid the race?

I just checked, and I think it is completely safe to set the size and position prior to showing the window.
See [QWidget::move](https://doc.qt.io/qt-5/qwidget.html#pos-prop) which can operate on hidden windows, and no limitations listed anywhere in [QWindow](https://doc.qt.io/qt-5/qwindow.html) with regards to when you can set size or position.

Additionally, I see that the about dialog does set the fixed size right before showing. Thus, based on the documentation, I see no reason why the position cannot be set prior to showing the window.

kakaroto:
It's assuming things based on past experience with the TK toolkit (which could be completely different behavior with QT). Also, we're not "setting the size" but rather "getting the size", I don't know if the correct size gets calculated (based on child element's min sizes) until the window is shown or not.
When I get the time, I'll try to print the dialog's geometry before and after a show, and print whether or not the window is 'exposed' (or received the signal) after calling show, to confirm my assumptions.

kakaroto:
Yeah, I did my tests on the about page, the settings page and the frequency analyzer page. First, I can say that I was right, the size doesn't get known until we call show(). To be more precise, the size for all the windows is correct before and after the call to show() but that's because the _ui.py file has a resize() with hardcoded size values in there, it's different though for the settings window, maybe because the hardcoded size was too small, but yes, the size changes (about 10 pixels more) after the call to show() since it forces it to recalculate its size.
I'm not sure when/whether the window isn't show until AFTER the call to show() is called, but I added a changeEvent handler and I see events "33" (WindowTitleChange) and "21" (ParentChange) when we create the window, then no changeEvent is called when we call `show()` (I was expecting an ExposeEvent but nope) then we get two ActivationEvents which mean (https://doc.qt.io/qt-5/qevent.html) "A widget's top-level window activation state has changed."

So I think that's why, the show() makes the window re-calculate its size, which is why centering should be done after the call to show() even though it's not really needed in this case since all the .ui files have a hardcoded size value in them, and the window doesn't get actually shown until we return to the main loop.

gedakc:
Overall code enhancments in this PR look good to me.  However I did encounter some strange messages in the console log when testing the **Settings** window in Windows 10.

Do you see similar _DirectWrite: CreateFontFaceFromHDC() failed..._ messages on Windows 10?

Test results follow.

### Test Window Positions

| OS | Window | Before (0.9.0) | After (PR 576 applied) |
| --- | ---  | --- | --- |
| Kubuntu 16.04 | About | Centered | Centered |
| Kubuntu 16.04 | Settings | Centered | Centered |
| Kubuntu 16.04 | Import | Centered | Centered |
| Kubuntu 16.04 | Compile / Export | Centered | Centered |
| Kubuntu 16.04 | Manager Exporters | Centered | Centered |
| Kubuntu 16.04 | Frequency Analyser | Bottom left of screen | Centered |
| --- | ---  | --- | --- |
| Windows 10 | About | Upper left of screen | Centered below parent window title bar |
| Windows 10 | Settings | Mostly centered | Centered [1] |
| Windows 10 | Import | Upper left of screen | Centered below parent window title bar |
| Windows 10 | Compile / Export | Upper left of screen | Centered |
| Windows 10 | Manage Exporters | Upper left of screen | Centered below parent window title bar |
| Windows 10 | Frequency Analyser | Mostly centered | Centered below parent window title bar [2] |

[1] Console Log Messages after selecting **Edit -> Settings**:

```
C:\Users\gedak>C:\Python37\python.exe C:\manuskript.gedakc\bin\manuskript
Debug: Web rendering engine used: QTextEdit
Note: No translator found or loaded from system locale for locale en_US.
Loading: C:/Users/gedak/Documents/deleteme-1.msk
Project C:/Users/gedak/Documents/deleteme-1.msk loaded.
DirectWrite: CreateFontFaceFromHDC() failed (Indicates an error in an input file such as a font file.) for QFontDef(Family="Fixedsys", pointsize=12, pixelsize=15, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("Fixedsys", lfWidth=0, lfHeight=-15) dpi=96
DirectWrite: CreateFontFaceFromHDC() failed (Indicates an error in an input file such as a font file.) for QFontDef(Family="Fixedsys", pointsize=12, pixelsize=15, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("Fixedsys", lfWidth=0, lfHeight=-15) dpi=96

```

[2] Example of Frequency Analyser Window centered below parent window title bar on Windows 10.

Manuskript-Frequency-Analyser-Window-on-Windows10

worstje:
@gedakc Those DirectWrite messages are not something you need to worry about in the context of this PR. I have seen those ever since I downloaded the first version, and while I can't be sure those errors are without harm (I never used the affected fonts inside manuskript), they do not seem to affect anything beyond that and the source of the issue is very likely in the Qt core itself.

It shows up whenever it tries to use a bitmap font as opposed to the typical TrueType/OpenType/whatever font. I have another font like that installed on my computer, and this isn't the first time a program isn't completely happy with the fact it is one of those old-fashioned fonts. 😄

kakaroto:
Thanks for testing/confirming. I've never seen those warnings myself, it might have something to do with installed fonts on the system.
For the window under the title bar, I hadn't seen that because I always have manuskript full screen, but it looks like it just centers the window on the parent, but in this case, the parent is smaller, so it appears that way ? Could it be that it was actually centered, it just happens that the sizes make it fit perfectly under the the title bar ? I just tested and it does fit with the minimal window size of the main window...
Is the "Centered below parent window title bar" an issue here or is the PR good as is ?

gedakc:
I do not think that "Centered below parent window title bar" is an issue.  I mentioned it because the behaviour differs from GNU/Linux which centers all windows based on the horizontal and vertical center of the parent window.  Windows 10 behaviour seems different and seems to try to ensure that the parent window title bar is visible if at all possible.

Based on the @worstje comments regarding the _DirectWrite_ messages, I went back an tested with the **master** branch which represents Manuskript 0.9.0 and the same messages occur.  Hence I agree with @worstje that the messages are not caused by this PR and are likely related to fonts and the Qt package(s).

Issue #535, opened 2019-03-12
pandoc export crashes if project title is empty


franapoli:
If the "Title" field is left empty in "Book information", no pandoc export format will work. I am running Manuskript 0.8 on Ubuntu 18.10.

worstje:
Do you get a specific error in the log? Does a window popup? Does it silently fail?

I'll see if I can reproduce it later.

gedakc:
Confirmed crash with Manuskipt 0.9.9 on Ubuntu 16.0 with the following steps:

1. Start manuskript and create new project using **Novel** template.

2. On **General** pane confirm that there is no text beside **Title**.  

3. Select **File -> Compile**.  

4. Set **Export to** to _Pandoc - HTML_.

5. Click **Preview** button.

6. Pandoc crashes.

Console log as follows:
```
$ bin/manuskript 
Debug: Web rendering engine used: QWebView
Running manuskript version 0.9.0.
Found translation in settings: 
Note: No translator found or loaded from system locale for locale en_CA.
Last accessed directory "/home/gedakc/tmp" loaded.
Project d3.msk saved.
Traceback (most recent call last):
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/exporters/exporter.py", line 101, in preview
    F.preview(self.settingsWidget, self.previewWidget)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/HTML.py", line 35, in preview
    html = self.output(settingsWidget)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/abstractPlainText.py", line 41, in output
    return self.exporter.convert(src, args, outputfile)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/__init__.py", line 71, in convert
    args.append("--metadata=title:{}".format(mainWindow().mdlFlatData.item(0, 0).text().strip()))
AttributeError: 'NoneType' object has no attribute 'text'
```

PR #577, opened 2019-05-09
Fix crash if using a custom pandoc installation

### LINKED ISSUE #563: Crash when exporting with pandoc as custom path only
If pandoc is not globally installed and only a custom path is specified, manuskript crashes when it tries to export.
I've looked at the code and I think I see the issue but I can't test/confirm for now.
The Exception is "File not found" on the subprocess.Popen : 
https://github.com/olivierkes/manuskript/blob/2b4943c36b20c2c4f889e2e21c1f7b54d8f53ace/manuskript/exporter/pandoc/__init__.py#L75

The issue I think is that the command is built using `self.cmd` : https://github.com/olivierkes/manuskript/blob/2b4943c36b20c2c4f889e2e21c1f7b54d8f53ace/manuskript/exporter/pandoc/__init__.py#L51

but the `self.cmd` is empty and instead `self.customPath` needs to be used.
The `version()` method uses `self.run([--version])` and the run method in basic.py does : https://github.com/olivierkes/manuskript/blob/2b4943c36b20c2c4f889e2e21c1f7b54d8f53ace/manuskript/exporter/basic.py#L55-L59

kakaroto:
Fixes #563

### DIFF
--- a/manuskript/exporter/basic.py
+++ b/manuskript/exporter/basic.py
@@ -59,7 +59,7 @@ class basicExporter:
             run = self.customPath
         else:
             print("Error: no command for", self.name)
-            return
+            return None
         r = subprocess.check_output([run] + args)  # timeout=.2
         return r.decode("utf-8")
 
--- a/manuskript/exporter/pandoc/__init__.py
+++ b/manuskript/exporter/pandoc/__init__.py
@@ -48,7 +48,14 @@ class pandocExporter(basicExporter):
             return ""
 
     def convert(self, src, args, outputfile=None):
-        args = [self.cmd] + args
+        if self.isValid() == 2:
+            run = self.cmd
+        elif self.isValid() == 1:
+            run = self.customPath
+        else:
+            print("Error: no command for pandoc")
+            return None
+        args = [run] + args
 
         if outputfile:
             args.append("--output={}".format(outputfile))
--- a/manuskript/importer/pandocImporters.py
+++ b/manuskript/importer/pandocImporters.py
@@ -38,6 +38,9 @@ class pandocImporter(abstractImporter):
 
         r = pandocExporter().run(args)
 
+        if r is None:
+            return None
+
         if formatTo == "opml":
             return self.opmlImporter.startImport("", parentItem,
                                             settingsWidget, fromString=r)


gedakc:
When testing on Kubuntu 16.04 with this PR patch applied to current **develop** branch, I still experience a crash with a Pandoc custom path.

See [issue 563 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/563#issuecomment-493781838).

Following is the console log:
```
$ bin/manuskript 
Project d3.msk saved.
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/exporters/exporter.py", line 101, in preview
    F.preview(self.settingsWidget, self.previewWidget)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/HTML.py", line 35, in preview
    html = self.output(settingsWidget)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/abstractPlainText.py", line 41, in output
    return self.exporter.convert(src, args, outputfile)
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/exporter/pandoc/__init__.py", line 78, in convert
    args.append("--metadata=title:{}".format(mainWindow().mdlFlatData.item(0, 0).text().strip()))
AttributeError: 'NoneType' object has no attribute 'text'
```

Can you retest?

gedakc:
Ah, just found that the crash I'm experiencing is due to [Issue 535 - pandoc export crashes if project title is empty](https://github.com/olivierkes/manuskript/issues/535).

I will update the test steps and re-test.

gedakc:
After re-testing this PR with the updated steps, Manuskript no longer crashes when using a Pandoc custom path.

See [issue 563 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/563#issuecomment-493781838).

Using Ubuntu 16.04 I have confirmed that this PR fixes the crash.

If no concerns are raised then I will plan to merge this PR with the **develop** branch in the next day or so.

gedakc:
'Merging this PR with the **develop** branch for inclusion in the next release of Manuskript.

PR #578, opened 2019-05-09
Fix crash if invalid character is inserted into the text.

### LINKED ISSUE #562: Crash on insertion of new page character
Not at home now so limited testing abilities, but I was showing a friend Manuskript and it kept crashing and I managed to find a way to reproduce the crash. There might be an easier way to reproduce it, but here's what I got : 

From a google document, press Ctrl-Enter to create a page break, then select the page break (select text including the new page 'character', the selection should appear longer than on an empty line)
Paste that into a scene in manuskript, once it saves, it will crash with the following error : 
```
  File "manuskript\mainWindow.py", line 765, in saveDatas
  File "manuskript\loadSave.py", line 21, in saveProject
  File "manuskript\load_save\version_1.py", line 260, in saveProject
  File "manuskript\models\abstractModel.py", line 525, in saveToXML
  File "manuskript\models\abstractItem.py", line 281, in toXML
  File "manuskript\models\abstractItem.py", line 281, in toXML
  File "manuskript\models\abstractItem.py", line 272, in toXML
  File "lxml.etree.pyx", line 744, in lxml.etree._Element.set (src\lxml\lxml.etree.c:44335)
  File "apihelpers.pxi", line 554, in lxml.etree._setAttributeValue (src\lxml\lxml.etree.c:19531)
  File "apihelpers.pxi", line 1393, in lxml.etree._utf8 (src\lxml\lxml.etree.c:27125)
ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters

```

Hope you can fix, otherwise will look at it when I get back home next week.

kakaroto:
If an invalid character is inserted into the text, such as a "^L" (ASCII 0x0C)
when copy-pasting from a google document that has a page break in it, a crash
will happen as the character cannot be inserted into XML. This patch removes
those invalid characters from the text so the revisions.xml can be saved.

Fixes #562

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -9,6 +9,7 @@ from PyQt5.QtCore import Qt
 from PyQt5.QtGui import QIcon, QFont
 from PyQt5.QtWidgets import QTextEdit, qApp
 from lxml import etree as ET
+import re
 
 from manuskript import enums
 
@@ -20,6 +21,9 @@ class abstractItem():
 
     # Used for XML export
     name = "abstractItem"
+
+    # Regexp from https://stackoverflow.com/questions/8733233/filtering-out-certain-bytes-in-python
+    valid_xml_re = re.compile(u'[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+')
 
     def __init__(self, model=None, title="", _type="abstract", xml=None, parent=None, ID=None):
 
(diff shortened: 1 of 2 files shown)


BlackXanthus:
Question: wouldn't it be better to convert the character to it's correct manuskirpt counterpart? Perhaps it's not a task for this particular action (as it's a bug-fix, it definately shouldn't crash), but perhaps it should honour the line-break in some way?

kakaroto:
As far as I could find, there is no line-break counterpart in manuskript. Also, the character remains in the text, it only gets removed from the XML file (revisions) so you'd only lose it if you restore previous revisions. Note also that this is how I got to reproduce this crash, but any other non-xml-acceptable character would cause the same issue unfortunately.

gedakc:
Using the instructions from issue #563 I have confirmed the "crash on save" in kubuntu 16.04.  With the patch applied the "crash on save" does not occur.

I believe that the patch is a good compromise because we never want Manuskript to crash, and the likelihood that someone needs to restore text containing an invalid XML character, such as **Ctrl+L**, is slim.

PR #583, opened 2019-05-18
Track dirty state and have the UI behave accordingly

worstje:
Intending to learn more about the way Manuskript goes about saving the project in order to figure out how to tackle some recent issues, I stumbled into learning that Manuskript likes to save data a whole lot. Too much, in fact. When I close the project with unsaved changes, I expected those changes to not be saved... but they were. Worse, when closing the window with 'quit on save' turned off, the application **just closes without any warning of data loss!** All of this completely subverts my expectations of how a program using typical file-based operations involving opening, saving and closing files should work!

There are three more settings that influence when the program saves, and I personally consider them a bit overkill or even detrimental to the stated purpose. What if Manuskript forces a save when nothing was changed and something goes wrong? Saving too much can in fact be dangerous!

For now, I have left existing functionality as-is, but I would prefer to
respect the dirty flag I have introduced in this commit for at least the
'save-on-quit' and 'save every X minutes' features. (The third is
smarter and only triggers after noticing changes, so it is less
important.)

Making sure the dirty flag works as expected is the first step in making such changes in the future. (Assuming we can reach some sort of consensus on what the desired behaviour would be.)

UI-wise, this commit now offers the user the opportunity to save their
changes, discard them, or outright cancel their action entirely when
performing a destructive action on a dirty project. As of this commit, I
have identified two of such scenarios:

1) closing the project,
2) closing the window with save-on-quit turned off.

**At this point, I have been using this commit for the past month without spotting any regressions, both while writing and trying to reproduce bug reports.**

Note that I have not added any translations; I initially started doing so for my native language but after seeing that it and other translations were out of date to varying degrees, I figured that the best translations would probably be made by someone who really embraces updating the various translations, and not me trying to squish it into this PR just to be complete. 😄

### DIFF
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -118,7 +118,14 @@ def saveProject(zip=None):
     # List of files to be moved
     moves = []
 
+    # MainWindow interaction things.
     mw = mainWindow()
+    project = mw.currentProject
+
+    # Sanity check (see PR-583): make sure we actually have a current project.
+    if project is None:
+        print("Error: cannot save project because there is no current project in the UI.")
+        return False
 
     # File format version
     files.append(("MANUSKRIPT", "1"))
@@ -295,10 +302,8 @@ def saveProject(zip=None):
 
     files.append(("settings.txt", settings.save(protocol=0)))
 
-    project = mw.currentProject
-
     # We check if the file exist and we have write access. If the file does
-    # not exists, we check the parent folder, because it might be a new project.
+    # not exist, we check the parent folder, because it might be a new project.
     if os.path.exists(project) and not os.access(project, os.W_OK) or \
        not os.path.exists(project) and not os.access(os.path.dirname(project), os.W_OK):
         print("Error: you don't have write access to save this project there.")
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -695,6 +695,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         # Clear datas
         self.loadEmptyDatas()
         self.saveTimer.stop()
+        self.saveTimerNoChanges.stop()
         loadSave.clearSaveCache()
 
         self.breakConnections()
@@ -817,6 +818,13 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         # to keep trying and continuously hitting the failure condition. Nor do we want to
         # risk a scenario where the timer somehow triggers a new save while saving.
         self.saveTimerNoChanges.stop()
+
+        if self.currentProject is None:
+            # No UI feedback here as this code path indicates a race condition that happens
+            # after the user has already closed the project through some way. But in that
+            # scenario, this code should not be reachable to begin with.
+            print("Bug: there is no current project to save.")
+            return
 
         r = loadSave.saveProject()  # version=0
 
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -7,7 +7,7 @@ from PyQt5.QtCore import (pyqtSignal, QSignalMapper, QTimer, QSettings, Qt,
                           QRegExp, QUrl, QSize, QModelIndex)
 from PyQt5.QtGui import QStandardItemModel, QIcon, QColor
 from PyQt5.QtWidgets import QMainWindow, QHeaderView, qApp, QMenu, QActionGroup, QAction, QStyle, QListWidgetItem, \
-    QLabel, QDockWidget, QWidget
+    QLabel, QDockWidget, QWidget, QMessageBox
 
 from manuskript import settings
 from manuskript.enums import Character, PlotStep, Plot, World, Outline
@@ -57,6 +57,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
 
         # Var
         self.currentProject = None
+        self.projectDirty = None  # has the user made any unsaved changes ?
         self._lastFocus = None
         self._lastMDEditView = None
         self._defaultCursorFlashTime = 1000 # Overriden at startup with system
@@ -272,6 +273,15 @@ class MainWindow(QMainWindow, Ui_MainWindow):
                 self._lastFocus = new
                 break
             new = new.parent()
+
+    def projectName(self):
+        """
+        Returns a user-friendly name for the loaded project.
+        """
+        pName = os.path.split(self.currentProject)[1]
+        if pName.endswith('.msk'):
+            pName=pName[:-4]
+        return pName
 
     ###############################################################################
     # SUMMARY
@@ -618,33 +628,68 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         # We force to emit even if it opens on the current tab
         self.tabMain.currentChanged.emit(settings.lastTab)
 
+        # Make sure we can update the window title later.
+        self.currentProject = project
+        self.projectDirty = False
+        QSettings().setValue("lastProject", project)
+
         # Add project name to Window's name
-        pName = os.path.split(project)[1]
-        if pName.endswith('.msk'):
-            pName=pName[:-4]
-        self.setWindowTitle(pName + " - " + self.tr("Manuskript"))
+        self.setWindowTitle(self.projectName() + " - " + self.tr("Manuskript"))
 
         # Stuff
         # self.checkPersosID()  # Shouldn't be necessary any longer
 
-        self.currentProject = project
-        QSettings().setValue("lastProject", project)
-
         # Show main Window
         self.switchToProject()
+
+    def handleUnsavedChanges(self):
+        """
+        There may be some currently unsaved changes, but the action the user triggered
+        will result in the project or application being closed. To save, or not to save?
+
+        Or just bail out entirely?
+
+        Sometimes it is best to just ask.
+        """
+
+        if not self.projectDirty:
+            return True  # no unsaved changes, all is good
+
+        msg = QMessageBox(QMessageBox.Question,
+            self.tr("Save project?"),
+            "<p><b>" +
+                self.tr("Save changes to project \"{}\" before closing?").format(self.projectName()) +
+            "</b></p>" +
+            "<p>" +
+                self.tr("Your changes will be lost if you don't save them.") +
+            "</p>",
+            QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel)
+
+        ret = msg.exec()
+
+        if ret == QMessageBox.Cancel:
+            return False  # the situation has not been handled, cancel action
+
+        if ret == QMessageBox.Save:
+            self.saveDatas()
+
+        return True  # the situation has been handled
+
 
     def closeProject(self):
 
         if not self.currentProject:
             return
 
+        # Make sure data is saved.
+        if not self.handleUnsavedChanges():
+            return  # user
+
         # Close open tabs in editor
         self.mainEditor.closeAllTabs()
 
-        # Save datas
-        self.saveDatas()
-
         self.currentProject = None
+        self.projectDirty = None
         QSettings().setValue("lastProject", "")
 
         # Clear datas
@@ -711,23 +756,6 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             self._toolbarState = ""
 
     def closeEvent(self, event):
-        # Save State and geometry and other things
-        sttgns = QSettings(qApp.organizationName(), qApp.applicationName())
-        sttgns.setValue("geometry", self.saveGeometry())
-        sttgns.setValue("windowState", self.saveState())
-        sttgns.setValue("metadataState", self.redacMetadata.saveState())
-        sttgns.setValue("revisionsState", self.redacMetadata.revisions.saveState())
-        sttgns.setValue("splitterRedacH", self.splitterRedacH.saveState())
-        sttgns.setValue("splitterRedacV", self.splitterRedacV.saveState())
-        sttgns.setValue("toolbar", self.toolbar.saveState())
-
-        # If we are not in the welcome window, we update the visibility
-        # of the docks widgets
-        if self.stack.currentIndex() == 1:
-            self.updateDockVisibility()
-        # Storing the visibility of docks to restore it on restart
-        sttgns.setValue("docks", self._dckVisibility)
-
         # Specific settings to save before quitting
         settings.lastTab = self.tabMain.currentIndex()
 
@@ -735,14 +763,42 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             # Remembering the current items (stores outlineItem's ID)
             settings.openIndexes = self.mainEditor.tabSplitter.openIndexes()
 
-        # Save data from models
-        if self.currentProject and settings.saveOnQuit:
-            self.saveDatas()
+            # Save data from models
+            if settings.saveOnQuit:
+                self.saveDatas()
+            elif not self.handleUnsavedChanges():
+                event.ignore() # user opted to cancel the close action
 
             # closeEvent
             # QMainWindow.closeEvent(self, event)  # Causing segfaults?
 
+        # User may have canceled close event, so make sure we indeed want to close.
+        # This is necessary because self.updateDockVisibility() hides UI elements.
+        if event.isAccepted():
+            # Save State and geometry and other things
+            appSettings = QSettings(qApp.organizationName(), qApp.applicationName())
+            appSettings.setValue("geometry", self.saveGeometry())
+            appSettings.setValue("windowState", self.saveState())
+            appSettings.setValue("metadataState", self.redacMetadata.saveState())
+            appSettings.setValue("revisionsState", self.redacMetadata.revisions.saveState())
+            appSettings.setValue("splitterRedacH", self.splitterRedacH.saveState())
+            appSettings.setValue("splitterRedacV", self.splitterRedacV.saveState())
+            appSettings.setValue("toolbar", self.toolbar.saveState())
+
+            # If we are not in the welcome window, we update the visibility
+            # of the docks widgets
+            if self.stack.currentIndex() == 1:
+                self.updateDockVisibility()
+
+            # Storing the visibility of docks to restore it on restart
+            appSettings.setValue("docks", self._dckVisibility)
+
     def startTimerNoChanges(self):
+        """
+        Something changed in the project that requires auto-saving.
+        """
+        self.projectDirty = True
+
         if settings.autoSaveNoChanges:
             self.saveTimerNoChanges.start()
 
@@ -757,11 +813,17 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             self.currentProject = projectName
             QSettings().setValue("lastProject", projectName)
 
-        r = loadSave.saveProject()  # version=0
+        # Stop the timer before saving: if auto-saving fails (bugs out?) we don't want it
+        # to keep trying and continuously hitting the failure condition. Nor do we want to
+        # risk a scenario where the timer somehow triggers a new save while saving.
         self.saveTimerNoChanges.stop()
+
+        r = loadSave.saveProject()  # version=0
 
         projectName = os.path.basename(self.currentProject)
         if r:
+            self.projectDirty = False  # successful save, clear dirty flag
+
             feedback = self.tr("Project {} saved.").format(projectName)
             F.statusMessage(feedback, importance=0)
         else:


gedakc:
I have to admit that I was initially surprised by the automatic saving of a project when I started to use manuskript.  As time went by I came to appreciate the automatic saves.

While many applications rely on the user to decide when to save, there is some precedent for automatic saving.  If I recall correctly Intuit's Quicken financial program automatically saves changes.

With that said, when automatic saving is disabled I think it is important to inform the user about the potential to lose unsaved changes when they exit the application.  This PR does implement such an enhancement.

For reference I checked what wording LibreOffice Writer uses for this same situation.

LibreOffice-Write-Quit-with-Changes-Pending

What do you think of changing to this wording (substituting "project" for "document") and layout?

Notice that the key statements are contained within the text box.  I think this is useful because some window managers don't display the title bar in an obvious location (or even a title bar like Ubuntu's Unity desktop).

worstje:
I think I based what I got off of some other project I came across when I was looking up the parameters for the QMessageBox. Since I am on Windows, many Qt-isms feel kind of out-of-place for my platform to begin with, so I just went with the format I came across and assumed it made sense.

If you want, I can adjust the PR to change the message to match?

Regarding the reasons for the saving behaviour... (sorry, this might come across a bit coarse, and I apologize if it does.)

My motivation for this PR was mostly the fact data could and would be lost simply because I didn't want it to be saved when quitting the application. (I was using it as an easy way to have a baseline for reproducing bugs without having to go through a dozen steps. Automagic saving kind of interferes with that aim.)

While I do believe a database-esque model like your example can have a reason for existing, those are also programs with more robust data storage for which the optimalisations and overhead in saving have been turned into a science. Manuskript is not a web app, is quite slow when saving in a number of scenarios and (based on the issues I have seen since attending here) far more likely to have data loss than it should be. The helpful default single-file setting seems to almost guarantees ultimate destruction in that regard.

If a database-esque model is what we are after, then we should not be saving in a shotgun approach. Rather, we should be saving intelligently, and it should not require the user to know how to best tweak the program to work as they'd expect. Blindly saving every X minutes is bad. Save when making changes, and limit the changes on disk to the bits being changed. (ZIP format kind of fails at this. No database would ever be stuck saved inside of a zip file!)

IMHO, if saving is a configurable setting in any way, it would be limited to 'database saving' behaviour or 'file-based' saving like in traditional applications. Let the user determine their expectations for how the program should work, and let that be the end of the dilemma. Three settings for different ways for saving to happen? A revision system with a handful more settings which also come down to even more saving? And somehow, none of these settings actually tackle the creation of backup files in case the project gets corrupted.

I love Manuskript, but the way saving is currently implemented is a joke. It confounds UX common sense, user expectations and programmer common sense. It needs to be simpler, it needs to do more, and it needs to do so both robustly and consistently.

gedakc:
> If you want, I can adjust the PR to change the message to match?

Please do.  I believe that would improve the ability of Manuskript to be used on multiple platforms.  The concern I'm thinking of is desktop/window managers that do not display the window title bar in an obvious place.

For a mock example see the proposed dialog screenshot below.

PR583-Proposed-Message-Box-on-Kubuntu1604

Now think of the screenshot but with the window title bar missing or placed elsewhere on the user's screen.

I know that not having a window title bar for a message dialog might seem ludicrous, but I recall reading about it in the past.  For an example of a message dialog that doesn't appear to include a window title, see [GNOME HIG - Design Patterns - Dialog Windows](https://developer.gnome.org/hig/stable/dialogs.html.en) and scroll down to **Message Dialogs**.

worstje:
No worries, I totally understand the logic you laid out, and it is completely sensible and right.

I simply didn't put that degree of thought into it; I stupidly assumed that whatever that other project did matched the 'platform' standards for such a dialog on Qt's most dominant platforms.

I'll take care of it tomorrow. 😄

worstje:
Apologies about the delay in making this ready. I ran into a problem where dock visibility was getting updated and it took me a while to find a fix I was happy with. (The code in question appeared a bit hacky in nature, so I tried to unhackify it.. but in the end it was just way easier to work around the way that code behaves.)

gedakc:
Thank you @worstje for the updated "Save Project?" dialog.  In my brief testing this PR performs as expected.

Of note is that in my testing I came across one anomaly.  More specifically when testing Manuskript the changes to **Settings** are not saved unless one saves the project.  This anomaly occurs both with and without this PR is applied.  

**Steps to Reproduce Anomaly**

- First I created a project using the **Novel** template, typed some text into a scene, and then exited the project and it was automatically saved with all the default **Settings**.

  Manuskript-Edit-Settings-Enable-Auto-Save

- Next I opened this project and I edited the **Settings** to ensure that the top three automatic **Saving** options were disabled.

  Manuskript-Edit-Settings-Disable-Auto-Save

- Next I typed some text into a scene, then I clicked the main window upper right hand side **X** to close Manuskript.  At the prompt "Save Project?" I chose **Discard**.

- When I next started manuskript and opened the same project, I noticed that the automatic **Saving** settings had reverted to what they were before I changed them.

  Manuskript-Edit-Settings-Enable-Auto-Save

From this observation I surmise that Manuskript does not automatically save setting changes as soon as these are made.


**Question**

@worstje are you able to reproduce this anomaly?

If so then this anomaly may be another opportunity for improving Manuskript operation, but could be handled in a different PR.

worstje:
Every single time I remember you asking me about this, it is around 2am and I am in bed trying to fall asleep. And today it is almost 3am and I still haven't gone to bed. 😄 

Suffice to say, I have not tested yet, but I just want to post. I do know from my research into the saving mechanism that the Settings dialog is an amalgamation of different settings. Half are saved as a part of the project file, the other half are saved elsewhere. I think it is the registry or wherever Qt decides is a good spot. I do recall seeing some separate calls, and I think the one we call covers the settings that are saved in the project itself. Or is the problem here that we are talking about settings that are supposed to be in the project settings file, yet aren't saved when trying to save?

(This is why we need to split up the project and program settings into separate UIs. Some of the behaviour is very counterintuitive. ☹️ )

Hopefully I'll finally remember to do some tests tomorrow. (But I've been forgetting about this for a week at this point so who knows, I might forget again...)

gedakc:
No worries about when/if you get around to testing.  With the pull of warmer weather my focus on computer work is less than when the snow flies.

I took another look at the updated code and noticed that the text strings are not marked for translation.  For translation strings it is best practice to keep the HTML formatting tags separate from the translated string.  See GNOME Human Interface Guidelines for [L10N Guidelines for Developers](https://wiki.gnome.org/TranslationProject/DevGuidelines) more specifically [Avoid markup wherever possible](https://wiki.gnome.org/TranslationProject/DevGuidelines/Avoid%20markup%20wherever%20possible).  I mention GNOME because I also work on [GParted](https://gparted.org).  For an example of this put into practice in Manuskript see:

https://github.com/olivierkes/manuskript/blob/761be5784a9ca44efd25d40fc285f11131835b83/manuskript/ui/about.py#L41-L46

The above isn't a pretty example but does demonstrate separating the formatting from the strings to be translated, and also shows strings that should not by translated like the name **Python**.  I also know there are many existing strings that do combine formatting with the translation so there is more work for some time in the future.

Would you be able to update the PR so that the appropriate strings are translatable?

Thanks,
Curtis

gedakc:
@worstje do you have time to separate the HTML formatting tags from the language strings to be translated?

If not then I'll just merge this PR as is, with the idea that sometime in the future the language strings can be improved.

worstje:
@gedakc It only took me a month, but the change has been made. 😄

gedakc:
'Glad to see the updated PR.  :-)

In testing I found one situation where an error cropped up.  The steps to reproduce are:

1. Create a project in Manuskript using the novel template
2. Disable automatic saving:  
   Select **Edit -> Settings** under **Saving** and disable  
   - **If no changes during 5 seconds**, and  
   - **Save on quit**.
3. Close **Settings** window.
4. Select **Editor** pane, expand **Chapter 1**, and select **Scene 1**.
5. Enter some text into the scene.
6. Choose **File -> Close project**.
7. Choose **Discard**.
   Manuskrip-PR583-Save-Project-Dialog

8. In the console window notice messages similar to the following:

```
$ bin/manuskript 
Loading: /home/gedakc/tmp/test-track-dirty-state-pr583.msk
Project /home/gedakc/tmp/test-track-dirty-state-pr583.msk loaded.
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/mainWindow.py", line 821, in saveDatas
    r = loadSave.saveProject()  # version=0
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/loadSave.py", line 21, in saveProject
    return v1.saveProject()
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 302, in saveProject
    if os.path.exists(project) and not os.access(project, os.W_OK) or \
  File "/usr/lib/python3.5/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: can't specify None for path argument
Project pr483-track-dirty-state.msk saved.
```

Note that if I choose to close the Window by clicking on the **X** in the upper right hand corner of Manuskript and choose **Discard** in the **Saving Project?** dialog then the project closes properly (without saving) and without the error messages.

gedakc:
Following is an update on the problem outlined above.

**IF** I perform the steps slowly enough so that at least one project save is completed, then the error does not crop up, and instead a console log similar to the following is seen:

```
$ bin/manuskript 
Project pr483-track-dirty-state-2.msk saved.
```

Note that I choose **File -> Close project** after the "`Project pr483-track-dirty-state-2.msk saved.`" message had already occurred in the console log.

gedakc:
I have not been able to reproduce the issue.  I will test some more.  If you have any thoughts on this then I am open to suggestions.

worstje:
I have taken a quick look (and am still looking) and I think I see the rough cause. Roughly put, when calling `MainWindow.saveDatas(projectName-None)` to invoke a save on an existing file, it does not verify whether `MainWindow.currentProject` is not None. According to the program structure, if you can save a project, you must have already specified a location for the project. I'll have to check what the save-by-timer does that my manual save does not do.

Yet somehow the new project code does not correctly set the new path in `mw.currentProject`. I haven't quite found where that is located, but I'll go hunt it down. (And probably include a sanity check where we bump into the current error.)

(FYI, this is exactly the sort of issue that originally set me on the crusade that created this branch to begin with. All the automatic saving can hide the details of what is going on, making it immensely more difficult to reproduce bugs. Next on my list (if I ever get unlazy enough) is to just separate the autosaving code into being a proper program setting as opposed to a project setting. Right now, it creates a project with whatever saving behaviour is the default... and that is super unintuitive. But alas, me and my soapbox... 😄 )

worstje:
I think I have some theories on what happened, but I don't really have a good way to test it. It is kind of tricky that you edited your post compared to the original email github sent me (so I am not quite sure what order of events enabled that error), so I am mostly going by the logfile you posted.

In welcome.py, `welcome.createFile()` can be called with `filename=None`, but that will accomplish nothing because the code ensures that there will indeed be a viable filename. As its final step, it calls `MainWindow.loadProject(filename, ...)` which thus ensures that `MainWindow.currentProject` should become not-None. A lot happens in that function, but there is no logic that can get triggered that prevents `MainWindow.currentProject` from being updated. Unless I am missing a particularly well-disguised return statement, that leaves only one option in my mind: a race condition.

It just so happens that we do a lot of bookkeeping with different timers, which relies on signals... so this seems reasonable. So I began writing this post... and then I got even more confused because the ordering doesn't seem to match the source code I see. Let me annotate to clarify what I mean:

> Debug: Web rendering engine used: QWebView
> Running manuskript version 0.9.0.
> Found translation in settings: 
> Note: No translator found or loaded from system locale for locale en_CA.

The program started. Not really useful save as a baseline.

> Loading: /home/gedakc/tmp/test-track-dirty-state-pr583.msk
> Detected file format version: 1. Zip: True.

These lines are loadSave.py in action, `loadProject(project)` to be exact. This is called by `MainWindow.loadDatas()` which is in turn called by `MainWindow.loadProject(..., loadFromFile=True)`, which is the case when loading from a file and not creating a new one. (And `MainWindow.loadProject` is the final stop in regards to loading actions.)

Basically, something started loading a project.

> Project /home/gedakc/tmp/test-track-dirty-state-pr583.msk loaded.

This confirms that we are looking at the `MainWindow.loadDatas()` code path because that is where this message comes from. And this can only be called from `MainWindow.loadProject`, thus confirming that loadFromFile=True.

Note that this message is a bit treacherous: yes, it finished loading the project, but the project settings, timers and UI have yet to be adjusted to reflect this.

> Traceback (most recent call last):
>   File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/mainWindow.py", line 821, in saveDatas
>     r = loadSave.saveProject()  # version=0
>   File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/loadSave.py", line 21, in saveProject
>     return v1.saveProject()
>   File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 302, in saveProject
>     if os.path.exists(project) and not os.access(project, os.W_OK) or \
>   File "/usr/lib/python3.5/genericpath.py", line 19, in exists
>     os.stat(path)
> TypeError: stat: can't specify None for path argument

This one speaks for itself: it starts out in `MainWindow.saveDatas`, implying it is Qt triggered. Either `MainWindow.saveTimer` or `MainWindow.saveTimerNoChanges` caused this.

> Last accessed directory "/home/gedakc/tmp" loaded.

And now we might be in `welcome.py`. I am not sure exactly what code path is happening now, but `openFile`, `saveAsFile` and `createFile` all call the `getLastAccessedDirectory()` that triggers this function. What they have in common is that these are starting points for loading a new project... but only when nothing was loaded before. But we just did... so this makes no sense?

There is also something in the Settings window that calls this as a part of setting a background image. This is hooked up to some update code which can be called by signals, so this could be a completely harmless invocation caused by messing around on the Settings screen.

> Project pr483-track-dirty-state.msk saved.

This is `MainWindow.saveDatas()` after its call to `loadsave.saveProject()` was successful.

> Loading: /home/gedakc/tmp/pr483-track-dirty-state.msk
> Detected file format version: 1. Zip: True.

We have seen this before: `loadsave.loadProject()`.

> Project /home/gedakc/tmp/pr483-track-dirty-state.msk loaded.

And this is `MainWindow.loadDatas()` again.

> Project pr483-track-dirty-state.msk saved.

And finally, `MainWindow.saveDatas()` which could be triggered from either a timer or elsewhere in the code like closing the window and triggering it through the new code.

**Long story short:** my guess is that something in `MainWindow.loadProject` takes long enough for one of the timers it starts up to kick in before it actually finishes loading. Why it takes so long? No clue. (What madness...)

In case it isn't obvious: this got very long since I am basically trying not to fall asleep while debugging this and I am driving myself nuts in the process. 🤣 Before I go through with madly coding to fix this on some other day, may I recommend some more testing to verify some details to prove my many conjectures?

1. If you open and close the settings window along the lines of your previous test methods, can you make the 'last accessed directory' log line show up?

2. Can you confirm the initial state of the program before your tests? The log files unfortunately aren't very clear on whether a project is autoloaded from a previous session, or whether it was manually loaded through some sort of user interaction.

3. (Bonus) Add a quick log line to show that `MainWindow.closeProject` and `MainWindow.closeEvent` are called. This would help immensely clarifying what is going on.

My tentative plans for fixing this are as follows:

* Improve logging to make it clearer what user actions happened as opposed to just internal calls.

* Make it so that the timers are loaded at the very end of `MainWindow.loadProject` as opposed to quite early on.

* And uhhh... oh right. An extra sanity check (or simply an assert?) to the relevant save function to make sure the current project is indeed not-None.

I'm done for now. Peace out and good night! 💤

gedakc:
> It is kind of tricky that you edited your post compared to the original email github sent me (so I am not quite sure what order of events enabled that error), so I am mostly going by the logfile you posted.

My apologies for that.  Unless I am following a pre-defined test script (which I only have for issues that already have a defined set of steps to reproduce the issue) then I try various combinations of things to test for robustness.  If something unusual crops up then I try to remember and document the steps.

After documenting and posting what _I thought_ were the steps (create new novel project), I realized that these did not match the console log.  This lead to changes to the steps to match the log.

> **Long story short:** my guess is that something in MainWindow.loadProject takes long enough for one of the timers it starts up to kick in before it actually finishes loading. Why it takes so long? No clue. (What madness...)

Your tracking of the console log makes sense to me in terms of what happened.  I also think we might be dealing with a race condition, given the event driven nature of GUI programming.

> 1. If you open and close the settings window along the lines of your previous test methods, can you make the 'last accessed directory' log line show up?

By opening and closing the settings window I have been unable to make 'last accessed directory' show up in the console log.

What does work is to start up Manuskript and click on the **Create** button.  This makes 'last accessed directory' show up immediately in the console log.

This must have been the point when I tried to create a project (pr483-track-dirty-state) before noticing that the issue had occurred.

> 2. Can you confirm the initial state of the program before your tests? The log files unfortunately aren't very clear on whether a project is autoloaded from a previous session, or whether it was manually loaded through some sort of user interaction.

To load the already existing project I start Manuskript, click the **Recent** drop down menu, and select the project name (e.g., test-track-dirty-state-pr583).

> 3. (Bonus) Add a quick log line to show that MainWindow.closeProject and MainWindow.closeEvent are called. This would help immensely clarifying what is going on.

This is a `git diff` of the print to console lines I added:

```
diff --git a/manuskript/mainWindow.py b/manuskript/mainWindow.py
index aa74f6e..cf55662 100644
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -678,6 +678,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
 
     def closeProject(self):
 
+        print("PR583 DEBUG: Start of mainWindow.closeProject()")
         if not self.currentProject:
             return
 
@@ -716,6 +717,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
 
         # Show welcome dialog
         self.switchToWelcome()
+        print("PR583 DEBUG: End of mainWindow.closeProject()")
 
     def readSettings(self):
         # Load State and geometry
@@ -756,6 +758,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             self._toolbarState = ""
 
     def closeEvent(self, event):
+        print("PR583 DEBUG: Start of mainWindow.closeEvent()")
         # Specific settings to save before quitting
         settings.lastTab = self.tabMain.currentIndex()
 
@@ -793,6 +796,8 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             # Storing the visibility of docks to restore it on restart
             appSettings.setValue("docks", self._dckVisibility)
 
+        print("PR583 DEBUG: End of mainWindow.closeEvent()")
+
     def startTimerNoChanges(self):
         """
         Something changed in the project that requires auto-saving.
```

## Finally reproduced it!

Steps:

1. Start Manuskript (e.g., bin/manuskript)

2. Click **Recent** and select an existing project (was previously created using the novel template and has default automatic save settings)

3. Select **Editor** pane, expand **Chapter 1**, and select **Scene 1**.

4. Enter some text into the scene.

   NOTE:  The following steps must be done immediately after entering the last character and quickly to avoid the automatic save after 5 seconds of no changes.

5. **QUICKLY** disable automatic saving:  
    Select **Edit -> Settings** under **Saving** and disable  
    - **If no changes during 5 seconds**, and  
    - **Save on quit**.

6. **QUICKLY** close **Settings** window.

7. **QUICKLY** choose **File -> Close** project.

8. **QUICKLY** choose **Discard**.

Issue occurs as shown in following console log:

```
$ bin/manuskript 
PR583 DEBUG: Start of mainWindow.closeProject()
Loading: /home/gedakc/tmp/test-track-dirty-state-pr583.msk
Project /home/gedakc/tmp/test-track-dirty-state-pr583.msk loaded.
PR583 DEBUG: Start of mainWindow.closeProject()
PR583 DEBUG: End of mainWindow.closeProject()
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/mainWindow.py", line 826, in saveDatas
    r = loadSave.saveProject()  # version=0
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/loadSave.py", line 21, in saveProject
    return v1.saveProject()
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/load_save/version_1.py", line 302, in saveProject
    if os.path.exists(project) and not os.access(project, os.W_OK) or \
  File "/usr/lib/python3.5/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: can't specify None for path argument
PR583 DEBUG: Start of mainWindow.closeEvent()
PR583 DEBUG: End of mainWindow.closeEvent()
```

Let me know if there is anything else you would like me to try.

gedakc:
'Just noticed there were **two** invocations of _mainWindow.closeProject()_ in the log before a close of _mainWindow.closeProject()_.  That can't be good.

worstje:
@gedakc Please test whether this commit fixes the issue for you.

I have fixed the issue in several spots for the sake of avoiding future crashes if another race condition involving this particular bit of code happens again. If you want to test those bits out too, comment out the newly added `self.saveTimerNoChanges.stop()` in mainWindow.py to make sure they get triggered with your previous usecase.

Sorry it took so long; apparently taking a month between updates is my thing right now. 😄

I can squash the commit at a later date, but I am not sure it should be. The root cause for this bug also exists outside of this particular PR; it is just that the patch happens to expose it. Since it does some general code health improvements to minimize it happening in the future, I think it warrants being its own commit once you bring it in so it isn't muddled up with the various UI changes this patch introduces as well.

worstje:
As for the two invocations of closeProject, I am not sure if that is desired or an oversight.

With the way `closeProject()` is implemented with a check for `self.currentProject` right at the top and setting it to None a few lines later, I believe that multiple `mw.closeProject()` calls should be a supported operation in nature.

It would not achieve very much except for multiple popups asking you to save if you keep canceling the close-project operation!), but at the very least, it should never crash the program.

Other code wants to try and close the project? Let them. Let `closeProject()` worry about whether or not it is already closed.

If we end up calling it sequentially for some reason, then those might be separate code paths who are not usually called together, but in this instance are. It does not necessarily denote a bug; it might just be a sequence of operations we did not realize exists, yet is fine to exist.

Or it is an oversight. But it should still be a harmless oversight.

(For what it is worth: just glancing at the log I think here it is just a call to close the existing project as a part of the loading process. And since there is nothing to close, it should be a noop, which it appears to be. So I am honestly not too bothered digging into the exact cause since it wasn't the reason the program was crashing.)

gedakc:
Good work @worstje.  Using your most recent updates I re-tested and the crash has gone away.  :-)

I am in agreement with keeping the two commits separate as you suggested.  Ideally commits should make logical sense, and each one should compile so that tools like `git bisect` work.  My concern is if the commits only represent the random path one took to the solution.  Ideally these should be cleaned up so that later developers can logically follow what was done.

Thanks for sharing your thoughts on `closeProject()`.  I appreciate that one line of thinking is to have the method make sure that it does the right thing in each case regardless of how it is invoked.  As such I'm on board with the code improvement you implemented.

Thanks also for responding to some of the other issue reports.  From your responses I believe we are on the same page in terms of what should be included in Manuskript to keep the project manageable and to stay focused on what is important which I tried to capture in the [Mission Statement](https://github.com/olivierkes/manuskript/wiki#mission-statement).

I will do more review and testing with an eye towards merging this PR in the next day or so.  Note that I will be attending the [When Words Collide](https://www.whenwordscollide.org/) festival this weekend so I may not respond promptly to questions until next week.

gedakc:
'Just discovered one very minor issue.

Even though the situations that invoke the new `print()` strings in the new commit should rarely if ever happen, I think we should still make the strings available for translation.

worstje:
I disagree.

`print()` calls serve the purpose of assisting us developers in our debugging by the means of log files. Some parts of the codebase (but not all) wrap it in a `log()` wrapper, but it doesn't change the purpose. Translating the log strings will complicate things for you and me. If I am fed a log output file with Italian or Chinese translations, I'm going to be helpless in figuring out what is going on.

I realize we are a bit more vocal than we perhaps should be, but all around we have a fair bit of coder-stuff regurgitating on the console as it is. But that would be an indication that we need to improve feedback towards the user rather than have them look at a lot of technical output statements that should not be in their power to do anything with.

gedakc:
Because this `print()` output is not intended for the user, I accept your position and will leave these strings untranslated.  Ideally all Manuskript output meant for the user should be translated, which in our case should be everything in the GUI.

I do agree that the console log is "more vocal" that it should be.  With that said it has been very helpful with troubleshooting so I am in no rush to change it.

I still have some more testing to do before I merge this with the **develop** branch.

I would like to get another manuskript release out soon (perhaps in a week or so).  If there are some items you think should be included then let me know.

worstje:
I am currently looking into that report of the non-working import; but because I ran into access violations inside qt I spent most of today trying to get a debug version of python+qt+pyqt working on my development VM that has a lot of MSVC versions on it in the hope I'll be able to squeeze out more information so I can start to actually tackle that problem. (It should not be that hard, and it is theoretically quite easy, but only if you find all the little steps on several different forums telling you what to do to achieve this...)

Ideally it'll be a quick fix, but if you don't see or hear within the next two days on that, you might as well forget it for the next version. But I am hopeful that it might be a quick-ish fix after I get over this speedbump.

Besides that I am currently working on amending the instructions on the wiki regarding compiling from source; at present it makes no mention of the new spellcheck libraries that kakaroto added support for several months ago.

worstje:
@gedakc Sorry for abusing this PR for communication purposes, but can you verify that the adjusted instructions for all the spell checkers on [this page](https://github.com/olivierkes/manuskript/wiki/Run-Manuskript-from-Source-Code-on-Windows) are indeed effective for the platforms mentioned at the top of the document? I doubt the commands are wrong, but I can't actually properly test them.

(I have spent the last three hours waiting for PyQt to compile, so by the time I have that running, I don't want to install a single Python module for a while, nevermind preparing all the VMs and exact version dependencies listed in the document just to make sure the version numbers listed are indeed valid and effective.)

gedakc:
Thank you @worstje for updating the [Run Manuskript from Source Code on Windows](https://github.com/olivierkes/manuskript/wiki/Run-Manuskript-from-Source-Code-on-Windows) instructions. From a read through, your added steps look good to me.  I made one minor change to indicate that PyEnchant is _Recommended for all except 64-bit Python on Windows_.  At the moment I have too many other things on my plate so I won't be testing the various combinations of spell checkers.  Perhaps @kakaroto can test these and also check the instructions?

What I have done is ensure that PyEnchant does work.  If our users encounter difficulties with the other two spell checkers then these can be addressed as issues at a later time.

I will proceed to rebase and merge this request and then move on to other recent PRs.

gedakc:
It appears that we have encountered some problems with this patch set and our automatic testing using Travis CI.  See [Job #329.2 - olivierkes/manuskript](https://travis-ci.org/olivierkes/manuskript/jobs/570930667).  The log portion from the test section is as follows:

```
============================= test session starts ==============================

platform linux -- Python 3.6.7, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /opt/python/3.6.7/bin/python3.6
cachedir: .pytest_cache
rootdir: /home/travis/build/olivierkes/manuskript
collecting ... Debug: Web rendering engine used: QWebEngineView
Note: No translator found or loaded from system locale for locale en_US.
collected 13 items
manuskript/tests/test_functions.py::test_wordCount PASSED
manuskript/tests/test_functions.py::test_convert PASSED
manuskript/tests/test_functions.py::test_several PASSED
manuskript/tests/test_functions.py::test_outlineItemColors PASSED
manuskript/tests/test_functions.py::test_paths PASSED
manuskript/tests/test_functions.py::test_mainWindow Objects: 1708
PASSED
manuskript/tests/test_settingsWindow.py::test_general Loading: /tmp/tmpfbouhmfh.msk
Detected file format version: 1. Zip: False.
Project /tmp/tmpfbouhmfh.msk loaded.
PASSED
manuskript/tests/models/test_outlineItem.py::test_outlineItemsProperties PASSED
manuskript/tests/models/test_outlineItem.py::test_modelStuff Character POV not found: 1
Character POV not found: 1
Character POV not found: 1
Character POV not found: 1
WARNING ! There are some items with same IDs: ['0', '3', '7', '3', '7', '0', '3']
PASSED
manuskript/tests/models/test_references.py::test_references Loading: /tmp/tmp55f5hogy.msk
Detected file format version: 1. Zip: False.
Project /tmp/tmp55f5hogy.msk loaded.
Ref not found
Error: Ref {C:999:} not found
Ref not found
Ref not found
Ref not implemented
PASSED
manuskript/tests/ui/test_welcome.py::test_autoLoad QXcbConnection: XCB error: 8 (BadMatch), sequence: 613, resource id: 2097162, major code: 42 (SetInputFocus), minor code: 0
QXcbConnection: XCB error: 8 (BadMatch), sequence: 619, resource id: 2097168, major code: 42 (SetInputFocus), minor code: 0
QXcbConnection: XCB error: 8 (BadMatch), sequence: 625, resource id: 2097171, major code: 42 (SetInputFocus), minor code: 0
```

I will investigate further to see what might be causing the _QXcbConnection: XCB error: 8 (BadMatch)_ error.

gedakc:
Finally reproduced the error.  The problem is that the new dialog prompt to "Save project?" is popped up, but not automatically cleared by the test in [`manuskript/manuskript/tests/ui/test_welcome.py`](https://github.com/olivierkes/manuskript/blob/develop/manuskript/tests/ui/test_welcome.py).

The resulting screen from running `pytest -vs` looks like the following:

Manuskript-TravisCI-pytest-vs-broken2

There must be a way to code the test so either this prompt is not displayed, or the prompt is automatically cleared.

In the meantime both the macOS X and Linux Travis CI tests are failing.

I will continue investigating.

gedakc:
A fix for the Travis CI tests failing has been posted in PR #615.

Issue #304, opened 2018-01-16
pandoc v2 has deprecated some options and extensions so manuskript is giving error.


rp-karanth:
The newer version of pandoc has deprecated many extensions, so there is need to make manuskript compatible to both version.

This was the error popped up when I ran manuskript with pandoc v2.05
pandoc 2.0.5

```
--normalize has been removed.  Normalization is now automatic.
--smart/-S has been removed.  Use +smart or -smart extension instead.
For example: pandoc -f markdown+smart -t markdown-smart.
Try pandoc --help for more information.
```

olivierkes:
Right, thanks. 
Do you know what else has been deprecated?

rp-karanth:
In his announcement John MacFarlane enlisted them in [pandoc-discuss](https://groups.google.com/d/topic/pandoc-discuss/Q6JSKcEa-0U/discussion) google group. 
These are exerpts from it.

```
 --reference-doc replaces --reference-docx and --reference-odt 
  --smart removed (instead use +smart on reader and/or writer) 
  --normalize removed (normalization is automatic) 
  --latex-engine removed (instead use --pdf-engine) 
  --parse-raw removed (instead use -f latex+raw_tex or -f html+raw_html) 
  --epub-stylesheet removed (instead use --css) 
  --mathml no longer takes an argument 
```
I would also like to add additional information list of newly added options.

```
  --eol (specifies line endings), 
  --log (prints JSON representation of info and warning messages), 
  --request-header (allows specifying a header to be used when 
    pandoc fetches external resources), 
  --lua-filter (for running lua filters, see below), 
  --epub-subdirectory (for changing the directory used in epub 
    containers), 
  --resource-path (for setting the search path for images and 
    other resources), 
  --abbreviations (for specifying a custom abbreviations 
    file so that the Markdown parser can be sensitive to 
    abbreviations), 
  --syntax-definition (allowing XML syntax highlighting definitions 
    to be loaded dynamically). 
```

gedakc:
I've been looking into this issue and don't have a solution yet, but wanted to capture some of my findings.

First, the pandoc options can be found in the GUI when the **File -> Compile F6** window is open and one of the **Export to _Pandoc_** options is chosen.  The list of options is on the left-hand side under the **Pandoc** heading as shown in the following screen shot:

![screenshot_20180305_104325](https://user-images.githubusercontent.com/10405019/36990655-a4f852e4-2062-11e8-93f3-4c4b8db3256b.png)

[...]

PR #585, opened 2019-05-22
Fix pandoc export crashes is project title is empty

gedakc:
This PR prevents the crash mentioned in issue #535.

For test steps  see [issue 535 - steps to reproduce crash](https://github.com/olivierkes/manuskript/issues/535#issuecomment-493785776).

### DIFF
--- a/manuskript/exporter/pandoc/__init__.py
+++ b/manuskript/exporter/pandoc/__init__.py
@@ -74,8 +74,11 @@ class pandocExporter(basicExporter):
             if var and item and item.text().strip():
                 args.append("--variable={}:{}".format(var, item.text().strip()))
 
-        # Add title metatadata required for pandoc >= 2.x
-        args.append("--metadata=title:{}".format(mainWindow().mdlFlatData.item(0, 0).text().strip()))
+        # Add title metadata required for pandoc >= 2.x
+        title = "Untitled"
+        if mainWindow().mdlFlatData.item(0, 0):
+            title = mainWindow().mdlFlatData.item(0, 0).text().strip()
+        args.append("--metadata=title:{}".format(title))
 
         qApp.setOverrideCursor(QCursor(Qt.WaitCursor))
 


gedakc:
While this proposed PR worked on Kubuntu 16.04, the same solution failed on Windows 10.  More work required.

gedakc:
This PR is available for review and comments.

manuskript/exporter/pandoc/__init__.py
```diff
@@ -75,7 +75,10 @@ def convert(self, src, args, outputfile=None):
                 args.append("--variable={}:{}".format(var, item.text().strip()))
 
         # Add title metatadata required for pandoc >= 2.x
-        args.append("--metadata=title:{}".format(mainWindow().mdlFlatData.item(0, 0).text().strip()))
+        title = " "   # blank prevents args.append() crash if no project title
+        if mainWindow().mdlFlatData.item(0, 0):
+            title = mainWindow().mdlFlatData.item(0, 0)
+        args.append("--metadata=title:{}".format(title.text().strip()))
```
kakaroto: `title.text().strip()` would crash if there is no title set, since `title = " "` makes it a string, and python string doesn't have the `.text()` method.
Also, having the title be a space is useless if you call strip() afterwards on it.
I'd suggest this maybe : 
```
       title = "-"   # blank prevents args.append() crash if no project title
        if mainWindow().mdlFlatData.item(0, 0):
            title = mainWindow().mdlFlatData.item(0, 0).text().strip()
        args.append("--metadata=title:{}".format(title))
```

gedakc: Thanks for reviewing my patch.  And your comments and understanding of the code are spot on.

My goal was to set the title to an empty string in the case where a user did not provide a title for the Manuskript project.  Would it help if I added this goal to the comments?

Or do you think it is better to set the title to a minus sign "-" in the case where the user does not provide a title?

kakaroto: If an empty string works, then go for it. I maybe misunderstood the comment "blank prevents args.append crash if no project title". I thought it meant that you put a blank (a space) in the title because otherwise (if it's an empty string) then it would crash, potentially because "--metadata=title:" as argument would take the next argument in the command line as the title, thus breaking the command. But if it actually works with an empty string, then you can use that.

Actually, if the title isn't set, you could maybe not add the metadata argument at all, I don't think pandoc requires it for it to work. So... 
```
        # Only set the title metadata if a title has been set for the project.
        if mainWindow().mdlFlatData.item(0, 0):
            title = mainWindow().mdlFlatData.item(0, 0).text().strip()
            args.append("--metadata=title:{}".format(title))
```

would that work ?

gedakc: Your propsed solution immediately above should work for pandoc v1.x.  The problem arises with pandoc v2.x as it requires either a "title" or "pagetitle" to be set in the metadata.  See near the bottom of this [issue 304 comment](https://github.com/olivierkes/manuskript/issues/304#issuecomment-370505755) and also PR #418.

Interestingly the warning in pandoc v2.x indicates the "title" must be specified and non-empty.  However pandoc v2.x works as long as "title" is specified even if it is non-empty.  This makes me wonder if your minus sign "-" solution might be better as it may work better with future pandoc versions.  What do you think?

gedakc: Or I could set the empty title to "Untitled" as this seems to be what is suggested by the pandoc v2.x warning.

kakaroto: Yes, I was just writing to suggest "Untitled" then saw pandoc suggests the same then you suggested it as well, so I guess we have consensus :)

gedakc: Thanks for your input.  I will update the MR accordingly.

gedakc:
This PR has been successfully tested on Kubuntu 16.04, Fedora 29, and Windows 10.  With the PR applied there were no export crashes when the project title field was empty.

I will rebase and merge this PR with the **develop** branch.

Issue #468, opened 2019-02-04
New level 'unit' is reset


worstje:
While trying to create a project for an existing work-in-progress of mine, I tried to configure it to use 100 Chapters of 1,000 words to simplify importing all of it.

But no matter what I do, when the file is created, I end up with 'Something 1' thru 'Something 100'.

It isn't the end of the world, but it did have me recreate a couple of times trying to figure out if I was the stupid one for not properly adjusting the default 'Something' text. I did note that specifying a word count would reset the unit back to 'Something', but even after putting in 'Chapter' as the last thing before hitting the Create button would not let the text stick around.

I am trying out Manuskript for the very first time on Windows, so I am just using the official 0.8.0 release.

Manuskript looks quite promising so far, so thank you for creating it!

gedakc:
Thanks @worstje for your interest in Manuskript.

I was able to confirm the issue with the name being reset to "Something" upon creation.

For example I tried creating an **Empty fiction**:

![manuskript-before-create-empty-fiction](https://user-images.githubusercontent.com/10405019/52225919-c8acfa00-2868-11e9-8f9c-45ebd91a62e0.png)

And in the project created, "Something Chapters" was reset to "Something":

![manuskript-after-create-empty-fiction](https://user-images.githubusercontent.com/10405019/52225935-debaba80-2868-11e9-9c04-c3185fbf61fa.png)

In Manuskript note that the base level is a "text" or scene in fiction terminology.  If you wish Chapters "folders" containing Scenes "texts", then you would need to create at least two levels.

For more tips, see the [Manuskript Wiki](https://github.com/olivierkes/manuskript/wiki) and a tutorial I wrote called [Using Manuskript to Write a Fiction Novel](http://gedakc.users.sourceforge.net/display-doc.php?name=manuskript-novel-writing).

worstje:
Thank you. I am sure I'll learn the proper terminology in time; I just wanted to make sure to report the bug now before I'd forget about it.

As for the tips and tutorial, thank you for writing them and I will surely look at them at some point. For now I feel a lot more productive trying to input the notes of 100k words of written story just to see what I'll end up with. I have no proper foundation and have approached plotting my own unique way while writing until this point, so doing things for myself and bumping my head seems to be the easier route for me right now. Monkey see, Monkey do, Monkey stumbles, Monkey tries again. That's me. ;-)

gedakc:
[...]

PR #587, opened 2019-05-27
fix issue #468 'unit' is reset

NocturnalFred:
Hello, My apologizes I haven't contributed to anything outside of work before so I'm not sure please let me know if I miss some proper etiquette. 

This issue resolves two problems, the first not mentioned (but related) to issue 468 was caused by code in welcome.py, mainly line 296 

    spin.setProperty("templateIndex", self.template[1].index(d))

The value of d was often the same for each tuple in template[1], usually 10, so every spinner was storing the first location where 10 was found in template[1].

If the spinner values were initially 1,2,10,10,10, and you changed them to be, 1,2,3,4,5, the "index(d)" would only see index 3 for the 10s, and it would run the connect function 3 times, updating index three with the value "5" on its last run resulting in 1,2,5,10,10.

Adding enumeration and using the actual index instead of deriving the index with the index() function resolves the problem completely.  

Line 289 and 296 are all that are needed to resolve that issue.

And the second issue from 468 is that changes to the naming conventions aren't saved.  This is because there simply doesn't seem to have been any intention to do so in the code.  To resolve the issue I followed the same practice that had originally been done with the spinners, though for the QLineEdit values it sometimes resulted in a "None" being returned, which caused an error so I filtered None out with an if statement.

(I had to amend my commit as I had left some stuff in there that wasn't necessary, a print statement and an blank line)

### DIFF
--- a/manuskript/ui/welcome.py
+++ b/manuskript/ui/welcome.py
@@ -286,18 +286,19 @@ class welcome(QWidget, Ui_welcome):
 
         k = 0
         hasWC = False
-        for d in self.template[1]:
+        for templateIndex, d in enumerate(self.template[1]):
             spin = QSpinBox(self)
             spin.setRange(0, 999999)
             spin.setValue(d[0])
             # Storing the level of the template in that spinbox, so we can use
             # it to update the template when valueChanged on that spinbox
             # (we do that in self.updateWordCount for convenience).
-            spin.setProperty("templateIndex", self.template[1].index(d))
+            spin.setProperty("templateIndex", templateIndex)
             spin.valueChanged.connect(self.updateWordCount)
-
             if d[1] != None:
                 txt = QLineEdit(self)
+                txt.setProperty("templateIndex", templateIndex)
+                txt.textEdited.connect(self.updateWordCount)
                 txt.setText(d[1])
 
             else:
@@ -360,11 +361,20 @@ class welcome(QWidget, Ui_welcome):
                                    Qt.FindChildrenRecursively):
             total = total * s.value()
 
-            # Update self.template to reflect the changed values
+            # Update self.template to reflect the changed count values
             templateIndex = s.property("templateIndex")
             self.template[1][templateIndex] = (
                 s.value(),
                 self.template[1][templateIndex][1])
+
+        for t in self.findChildren(QLineEdit, QRegExp(".*"),
+                                   Qt.FindChildrenRecursively):
+            # Update self.template to reflect the changed name values
+            templateIndex = t.property("templateIndex")
+            if templateIndex is not None :
+                self.template[1][templateIndex] = (
+                self.template[1][templateIndex][0],
+                t.text())
 
         if total == 1:
             total = 0


gedakc:
Thank you @NocturnalFred for your help to improve Manuskript.  And congratulations on your first contribution to [Free Software](https://www.gnu.org/philosophy/free-sw.html).  :-)

From my review of the code this PR looks good.  In my testing with this PR applied on Kubuntu 16.04 I confirmed that the new level "_unit_" is indeed saved when creating a project.  Hence this PR does fix issue #468.

If there are no concerns then I plan to merge this PR with the **develop** branch in the next day or so.

Curtis

gedakc:
@NocturnalFred before merging this request I performed one more test that exposed a problem with this PR.

To test I tried creating a project using the **Empty fiction** template and specifying six levels as shown in the screenshot below.

Manuskript-Create-Six-Level-Project

When I choose **Create** and provide a file name, then Manuskript remains stuck on the same screen and the process running Manuskript goes to 100% usage.  This appears to be an infinite loop.

Please investigate to see if you can recreate this problem and if so then resolve the problem.

gedakc:
My bad.  The test I tried to perform takes a long time to complete due to the huge amount of levels being created.  I tried a smaller six level test and it did complete in a ressonable amount of time.

Manuskript-Create-Six-Level-Project-Test-2

I will proceed to merge this PR.

NocturnalFred:
> @NocturnalFred before merging this request I performed one more test that exposed a problem with this PR.
> 
> To test I tried creating a project using the **Empty fiction** template and specifying six levels as shown in the screenshot below.
> 
> Manuskript-Create-Six-Level-Project
> 
> When I choose **Create** and provide a file name, then Manuskript remains stuck on the same screen and the process running Manuskript goes to 100% usage. **EDIT 3**: ~This appears to be an infinite loop.~ Just takes a long time.
> 
> **EDIT 3**: ~Please investigate to see if you can recreate this problem and if so then resolve the problem.~
> 
> **EDIT**: Note this problem seems to appear in the **master** branch as well. The problem does not occur if three levels are specified.
> 
> **EDIT 2**: I think that a use case such as six levels is highly unusual.

Sorry for not seeing this earlier.  I also noticed that many multipliers causes an issue.  The first thing I did when I used manuskript was try to create something with many levels (I didn't know what I was doing) and it froze and I manually quit the program, and then I never tried doing that again.  But there's no legitimate reason to do that afaik, I just wanted to do it because that's what the UI feels like it was telling me to do, and I didn't have any plans for a particular writing project in mind, I was just playing with manuskript.

There may be cause to limit, at least at welcome screen, the maximum possible levels and their multipliers.  I don't know.  I can't imagine, for example, any reason to have 100 levels of 100 somethings.  But then again the user could simply realize after their first time making the mistake like I did not to do it again.  Or perhaps some performance testing could be in order to see what a reasonable multiplier is, add a check and add some kind of warning system of "this many levels has been known to freeze the program" kind of thing.  Maybe that's over kill.

gedakc:
> there's no legitimate reason to do that afaik (_create project with many lavels_)

I agree.  For now I think it's okay as there is no need that I can see to have even need 10s of levels. 

Thanks again for the patch.  :-)

PR #588, opened 2019-05-28
Spelling: Manuscript, could not, process, … No content

comradekingu:


worstje:
Most of that looks okay. The only thing I wonder about is the 'Manuskript' one.

I think the capitalized 'Manuskript' is only used for the application name on Mac in order to fit in on the greater mac ecosystem, but that lowercase is otherwise the default.

What's the policy on this, @gedakc ?

comradekingu:
@worstje That makes very little sense to me.
https://hosted.weblate.org/zen/manuskript/translations/it/?checksum=&offset=&q=manuskript&search=substring&source=on&type=all&ignored=False&date=&only_user=&exclude_user=
10 x Manuskript
3 x manuskript

https://github.com/olivierkes/manuskript/blob/develop/README.md is all Manuskript

It seems that it is just "manuskript" whenever someone forgot, like
https://hosted.weblate.org/translate/manuskript/translations/it/?checksum=c7145ea461e58e1a uses both.

worstje:
I am not saying we shouldn't have consistency. I don't even care which of the consistency options we go with (uppercase Manuskript like in this PR has my preference!); I just want to know exactly what the rule with regards to branding is here.

[This old pull request](https://github.com/olivierkes/manuskript/pull/570) had a brief discussion on what was suitable for the app / binary / package name and how that would make the Mac app differ from the rest of the builds. It is what was on my mind when I saw your PR, so rather than just blindly agreeing with your suggestion, I want to understand exactly what rules we follow in regards to capitalization.

Maybe those rules are already written out somewhere, but maybe they are as-of-yet unwritten... in which case it wouldn't hurt to have them spelled out and questioned, right? 😄

comradekingu:
@worstje Indeed.
It isn't at all clear that some strings belong to the macOS client when translating.
The law of the land then becomes doing what consistent application there is in the first few strings.

It isn't a library, and doing "Manuskript" takes the guesswork out of what to do when a sentence starts with "m/Manuscript".

https://www.theologeek.ch/manuskript/ only has two "manuskript", and one "MANUSCRIPT", otherwise "Manuskript".

gedakc:
> What's the policy on this (Manuskript upper/lowercase usage)?

As far as I know there is no documented or agreed upon policy.  As best I can tell the project name "Manuskript" is a simple play on / misspelling of the English word "manuscript".

In my opinion because "manuskript" is not a English word in a dictionary, I think references to the "Manuskript" project and application should use uppercase "M".  When referring to invoking the executable from the command line I think lowercase "m" should be used as in `manuskript`.

On the topic of this PR, I wonder why the change from "Plain text" to "Plaintext"?

The two English words "plain text" are descriptive and match the definition of plain text when referring to computer terms.  See [Plain Text](https://en.wikipedia.org/wiki/Plain_text).

From my brief research it would seem that the combined word "plaintext" refers to cryptography which is not the intended meaning.  See [Plaintext](https://en.wikipedia.org/wiki/Plaintext).

comradekingu:
@gedakc that is a distinction that has faded. The reason for doing it is to avoid compound word errors in translation. It avoids having to consider the rules of stricter languages. It could be "cleartext" for similar benefit.

gedakc:
@worstje what are your thoughts regarding "plain text" versus "plaintext"?

worstje:
Definitely **plain text**. While Wikipedia is not exactly the best source in existence, I think it is sufficient for this argument, and it has two articles, one both spellings.

[Plain text](https://en.wikipedia.org/wiki/Plain_text) - No images, no markup. Can be unicode, but unicode includes emojis, so IMHO emojis are totally fine for this.

[Plaintext](https://en.wikipedia.org/wiki/Plaintext) - "In cryptography, plaintext usually means unencrypted information pending input into cryptographic algorithms, usually encryption algorithms. Cleartext usually refers to data that is transmitted or stored unencrypted ('in the clear')."

(Edit: Just spotted that @gedakc already linked these as well, whoops. We're thinking along the same lines, I suppose!)

Ergo, **plaintext** is very much not what we want. While the interpretation of **plain text** might leave to be desired in regards to specific details (markup? images?) I think in the context of saving, it would make most sense to accept the most basic version of **plain text**: no formatting whatsoever. Of course, if you open a file as plain text, that means you are still exposed to all the formatting codes that would have absolutely no meaning.

If we want to truly specify formatting in a 'saving' context, then we have a more suitable name for it, and that is the name of the formatting language used. Manuskript uses .md for Markdown, so what prevents us from referring to that? The same goes for things like (la)tex, html, rtf and whatnot.

comradekingu:
@worstje The latter part of your argument is why I prefer "plaintext", markdown is still in the clear, whereas it isn't just plain text. It does use plain text chars though, so just there it could not qualify as plain text, as it relates to rules of how to write text.

For OTOH, "in the clear", It becomes a question of what is even humanly readable.
For an easy way to see it, "plaintext" is the least obfuscated version.

But is it even a thing to use it like that, or does "plaintext" _have_ to be cryptographic, or is there even a problem using cryptographic terms to describe something?

gedakc:
We do have differing opinions.  Because the original author chose "plain text" over "plaintext" and because the majority lean towards the two word "plain text" lets keep it that way.

@comradekingu if you change this PR back to the original two word "plain text" then I can merge it with the **develop** branch.

comradekingu:
@gedack OK :)

gedakc:
Thanks for the updated patch set.  I will squash and merge.

Issue #561, opened 2019-04-22
Crash on adding word goal in outline


DarkRedman:
Each time I'm on the plan tab and edit the goal of one of the act and press enter to valid, Manuskript crash, but alternatively, when I'm on redaction mode and edit the goal on the sidebar and press enter, it doesn't crash.

worstje:
Try disabling auto-saving. I suspect that feature is the culprit of most of the crash bugs while editing things on the outline screen.

gedakc:
@DarkRedman what version of Manusrkipt are you using, and on which operating system?

Manuskript renamed the **Redaction** pane to **Editor** since at least version 0.5.0.

DarkRedman:
Sorry that I didn't replied that fast.
I used the 0.9 version on Windows, now I exclusively use it on GNU/Linux and I don't have this problem of saving even with the option to automatically change if any modification occurred within 5 seconds and it works fine.

gedakc:
After much testing and investigation I finally reproduced this crash.

### Steps to Reproduce:

1. Start manuskript and create new project (no template).

2. Select **Outline** pane.

3. Click "Text Plus" icon to create a text (default name "New")

4. Select **Editor** pane.

5. Click on **New** to display empty text.

6. Select **Outline** pane.

7. Double-click the empty area on **New** line under title **Goal**, type in "300", and press **Enter**.

   Notice manuskript crash on segmentation fault with following console log:

```
$ bin/manuskript 
Debug: Web rendering engine used: QWebView
Running manuskript version 0.9.0.
Found translation in settings: 
Note: No translator found or loaded from system locale for locale en_CA.
Last accessed directory "/home/gedakc/tmp" loaded.
Project issue561.msk saved.
Project issue561.msk saved.
Traceback (most recent call last):
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/editors/editorWidget.py", line 354, in modelDataChanged
    self.updateStatusBar()
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/editors/editorWidget.py", line 389, in updateStatusBar
    mw.mainEditor.tabChanged()
  File "/home/gedakc/workspace/manuskript.gedakc/bin/../manuskript/ui/editors/mainEditor.py", line 138, in tabChanged
    self.updateStats()
[...]

PR #609, opened 2019-07-23
Fix crash when setting word Goal on new Text (scene) in Outline pane

gedakc:
See issue #561.

The problem appears to be a due to a combination of factors, such as:

- Python does not automatically convert an empty/blank variable to the
  integer zero (0)
- Default goal value is empty/blank for a new Text (scene)
- Asynchronous events can occur such that the change in the Outline
  pane of a new Text (scene) goal from empty/blank to a value is not
  saved to the data model prior to the update event in the Editor pane
  accessing the model value for the word count progress display.

Steps to Reproduce:

1. Start manuskript and create new project (no template).

2. Select **Outline** pane.

3. Click "Text Plus" icon to create a text (default name "New")

4. Select **Editor** pane.

5. Click on **New** to display empty text.

6. Select **Outline** pane.

7. Double-click the empty area on **New** line under title **Goal**,
  type in "300", and press **Enter**.

   Note that manuskript crashes with a segmentation fault.

Work around the crash by using the already existing manuskript
function toInt() which handles conversion of empty/blank values to
integer value zero (0).

### DIFF
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -57,7 +57,8 @@ def drawProgress(painter, rect, progress, radius=0):
     painter.setBrush(QBrush(colorFromProgress(progress)))
 
     r2 = QRect(rect)
-    r2.setWidth(r2.width() * min(progress, 1))
+    r2.setWidth(r2.width() * min(toInt(progress), 1))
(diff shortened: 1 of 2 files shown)


gedakc:
This PR is available for review and comments.

PR #613, opened 2019-08-08
Fix exports silently overwriting files (fixes #608) & small fix to dialog logic

### LINKED ISSUE #608: When compiling, it overwrite files without asking
First off, thanks for the great work!

I think I ran into an issue today for version 0.9: when I compile a project and use a file name that already exists, if overwrites the file without asking "Are you sure?". I think that alert should be added to avoid accidentally overwriting the wrong file.

Thanks!

worstje:
**Note: This has only been tested on Windows 10. While I expect no differences in earlier Windows versions, someone should definitely test whether this behaves as expected on other platforms like Linux.**

This fixes issue #608.

The code was adding on a suffix after the user had already settled on a filename through the QFileDialog presented to them. Besides the fact that changing the filename after the fact is sneaky as hell and betrays user expectations, the functionality we wanted already existed in QFileDialog to begin with.

Using the blame functionality, I can't find a reason why it was implemented the way it currently is. My best assumption is that it was believed this functionality did not exist due to QFileDialog.getSaveFileName() helper not exposing the setting, and that the given solution might be based on an early expectation of the dialog showing all sorts of different filters.

It is worth noting that the functionality we originally implemented theoretically supported a default suffix based on the chosen filter which is not the case for the functionality exposed by QFileDialog, but since all of our dialogs only expose a single filter anyways it is a moot point. It is definitely not worth working out a logic loop where the window closes, we ask for an override (perhaps risking bothering the user on the matter twice), and then re-opening the dialog to let them select something else in the case that they had answered negatively.

[This bug report](https://bugreports.qt.io/browse/QTBUG-11352) from 2011 shows someone tackling the exact same issue, and based on what I can see of the QFileDialog documentation (which is for 90% filled with 'TODO' in [the case of PyQt5](https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtwidgets/qfiledialog.html), so I also crosschecked with [the PyQt4 documentation](https://www.riverbankcomputing.com/static/Docs/PyQt4/qfiledialog.html)), nothing of note has changed that affects this scenario.

As such, simply relying on the most basic functionality available seems like the most sensible choice. Thus ripping out code and relying on a basic, existing solution instead.

(Also, the stowaway commit included means we no longer run the conversion process after canceling the dialog now for the majority of filters. I figured that was worth having in the upcoming v0.10.0 version.)

### DIFF
--- a/manuskript/exporter/manuskript/HTML.py
+++ b/manuskript/exporter/manuskript/HTML.py
@@ -21,6 +21,7 @@ class HTML(markdown):
 
     exportVarName = "lastManuskriptHTML"
     exportFilter = "HTML files (*.html);; Any files (*)"
+    exportDefaultSuffix = ".html"
 
     def isValid(self):
         return MD is not None
--- a/manuskript/exporter/manuskript/markdown.py
+++ b/manuskript/exporter/manuskript/markdown.py
@@ -16,6 +16,7 @@ class markdown(plainText):
 
     exportVarName = "lastManuskriptMarkdown"
     exportFilter = "Markdown files (*.md);; Any files (*)"
+    exportDefaultSuffix = ".md"
     icon = "text-x-markdown"
 
     def settingsWidget(self):
--- a/manuskript/exporter/manuskript/plainText.py
+++ b/manuskript/exporter/manuskript/plainText.py
@@ -5,7 +5,7 @@ from PyQt5.QtGui import QFont, QTextCharFormat
 from PyQt5.QtWidgets import QPlainTextEdit, qApp, QFrame, QFileDialog, QMessageBox
 
 from manuskript.exporter.basic import basicFormat
-from manuskript.functions import mainWindow
+from manuskript.functions import mainWindow, getSaveFileNameWithSuffix
 from manuskript.models import outlineItem
 from manuskript.ui.exporters.manuskript.plainTextSettings import exporterSettings
 import codecs
@@ -24,6 +24,7 @@ class plainText(basicFormat):
     # Default settings used in self.getExportFilename. For easy subclassing when exporting plaintext.
     exportVarName = "lastPlainText"
     exportFilter = "Text files (*.txt);; Any files (*)"
+    exportDefaultSuffix = ".txt"  # qt ignores the period, but it is clearer in our code to have it
 
     def __init__(self):
         pass
@@ -64,27 +65,15 @@ class plainText(basicFormat):
         else:
             filename = ""
 
-        filename, filter = QFileDialog.getSaveFileName(settingsWidget.parent(),
-                                                       caption=qApp.translate("Export", "Choose output file…"),
-                                                       filter=filter,
-                                                       directory=filename)
+        filename, filter = getSaveFileNameWithSuffix(settingsWidget.parent(),
+                                                     caption=qApp.translate("Export", "Choose output file…"),
+                                                     filter=filter,
+                                                     directory=filename,
+                                                     defaultSuffix=self.exportDefaultSuffix)
 
         if filename:
             s[varName] = filename
             settingsWidget.settings["Output"] = s
-
-            # Auto adds extension if necessary
-            try:
-                # Extract the extension from "Some name (*.ext)"
-                ext = filter.split("(")[1].split(")")[0]
-                ext = ext.split(".")[1]
-                if " " in ext:  # In case there are multiple extensions: "Images (*.png *.jpg)"
-                    ext = ext.split(" ")[0]
-            except:
-                ext = ""
-
-            if ext and filename[-len(ext)-1:] != ".{}".format(ext):
-                filename += "." + ext
 
         # Save settings
         settingsWidget.writeSettings()
--- a/manuskript/exporter/manuskript/plainText.py
+++ b/manuskript/exporter/manuskript/plainText.py
@@ -75,8 +75,8 @@ class plainText(basicFormat):
             s[varName] = filename
             settingsWidget.settings["Output"] = s
 
-        # Save settings
-        settingsWidget.writeSettings()
+            # Save settings
+            settingsWidget.writeSettings()
 
         return filename
 
@@ -84,14 +84,15 @@ class plainText(basicFormat):
         settings = settingsWidget.getSettings()
 
         filename = self.getExportFilename(settingsWidget)
-        settingsWidget.writeSettings()
-        content = self.output(settingsWidget)
-
-        if not content:
-            print("Error: No content. Nothing saved.")
-            return
 
         if filename:
+            settingsWidget.writeSettings()
+            content = self.output(settingsWidget)
+
+            if not content:
+                print("Error: No content. Nothing saved.")
+                return
+
             with open(filename, "w", encoding='utf8') as f:
                 f.write(content)
 
--- a/manuskript/exporter/pandoc/HTML.py
+++ b/manuskript/exporter/pandoc/HTML.py
@@ -16,6 +16,7 @@ class HTML(abstractPlainText):
     exportVarName = "lastPandocHTML"
     toFormat = "html"
     exportFilter = "HTML files (*.html);; Any files (*)"
+    exportDefaultSuffix = ".html"
     requires = {
         "Settings": True,
         "Preview": True,
--- a/manuskript/exporter/pandoc/PDF.py
+++ b/manuskript/exporter/pandoc/PDF.py
@@ -23,6 +23,7 @@ class PDF(abstractOutput):
     exportVarName = "lastPandocPDF"
     toFormat = "pdf"
     exportFilter = "PDF files (*.pdf);; Any files (*)"
+    exportDefaultSuffix = ".pdf"
     requires = {
         "Settings": True,
         "Preview": True,
--- a/manuskript/exporter/pandoc/abstractOutput.py
+++ b/manuskript/exporter/pandoc/abstractOutput.py
@@ -10,6 +10,7 @@ class abstractOutput(abstractPlainText):
     toFormat = "SUBCLASSME"
     icon = "SUBCLASSME"
     exportFilter = "SUBCLASSME"
+    exportDefaultSuffix = ".SUBCLASSME"
     requires = {
         "Settings": True,
         "Preview": False,
--- a/manuskript/exporter/pandoc/abstractPlainText.py
+++ b/manuskript/exporter/pandoc/abstractPlainText.py
@@ -16,6 +16,7 @@ class abstractPlainText(markdown):
     toFormat = "SUBCLASSME"
     icon = "SUBCLASSME"
     exportFilter = "SUBCLASSME"
+    exportDefaultSuffix = ".SUBCLASSME"
 
     def __init__(self, exporter):
         self.exporter = exporter
--- a/manuskript/exporter/pandoc/outputFormats.py
+++ b/manuskript/exporter/pandoc/outputFormats.py
@@ -13,6 +13,7 @@ class ePub(abstractOutput):
     exportVarName = "lastPandocePub"
     toFormat = "epub"
     exportFilter = "ePub files (*.epub);; Any files (*)"
+    exportDefaultSuffix = ".epub"
 
 
 class OpenDocument(abstractOutput):
@@ -23,6 +24,7 @@ class OpenDocument(abstractOutput):
     toFormat = "odt"
     icon = "application-vnd.oasis.opendocument.text"
     exportFilter = "OpenDocument files (*.odt);; Any files (*)"
+    exportDefaultSuffix = ".odt"
 
 
 class DocX(abstractOutput):
@@ -33,4 +35,5 @@ class DocX(abstractOutput):
     toFormat = "docx"
     icon = "application-vnd.openxmlformats-officedocument.wordprocessingml.document"
     exportFilter = "DocX files (*.docx);; Any files (*)"
+    exportDefaultSuffix = ".docx"
 
--- a/manuskript/exporter/pandoc/plainText.py
+++ b/manuskript/exporter/pandoc/plainText.py
@@ -14,6 +14,7 @@ class markdown(abstractPlainText):
     exportVarName = "lastPandocMarkdown"
     toFormat = "markdown"
     exportFilter = "Markdown files (*.md);; Any files (*)"
+    exportDefaultSuffix = ".md"
 
 
 class reST(abstractPlainText):
@@ -24,6 +25,7 @@ class reST(abstractPlainText):
     toFormat = "rst"
     icon = "text-plain"
     exportFilter = "reST files (*.rst);; Any files (*)"
+    exportDefaultSuffix = ".rst"
 
 
 class latex(abstractPlainText):
@@ -35,6 +37,7 @@ class latex(abstractPlainText):
     toFormat = "latex"
     icon = "text-x-tex"
     exportFilter = "Tex files (*.tex);; Any files (*)"
+    exportDefaultSuffix = ".tex"
 
 
 class OPML(abstractPlainText):
@@ -47,5 +50,5 @@ class OPML(abstractPlainText):
     toFormat = "opml"
     icon = "text-x-opml+xml"
     exportFilter = "OPML files (*.opml);; Any files (*)"
-
+    exportDefaultSuffix = ".opml"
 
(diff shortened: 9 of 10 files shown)


worstje:
Assuming all works fine with the new bugfix, I'll make sure to re-order and squash where appropriate for the final merge.

gedakc:
I just started testing this PR and came across some interesting differences.

On Kubuntu 16.04, the compile/export to the same filename always silently overwrites the file with no warning (I chose Pandoc HTML).
- Python 3.5.2
- PyQt 5.5.1
- Qt 5.5.1

On Debian 9, the compile/export to the same filename does prompt regarding replace / overwrite.
- Python 3.7.3
- PyQt 5.13.0
- Qt 5.13.0

Prompt text (filename was "test-manuskript.html"):
```
A file named "test-manuskript.html"
already exists.  Do you want to replace it?
The file already exists in "user".  Replacing it will overwrite its contents.
```

Based on this it appears that there may be different built-in behaviour depending upon the Qt version.

worstje:
Question: why remove this patch from the milestone? To my knowledge, it fixes the cause of the originally reported issue.

I realize there is some inconsistent behaviour, but to my understanding that is not the bug that was reported here, nor the fix that was written. And even if it is, it is not a shortcoming we can fix in a nice manner: it is Qt and/or the OS that are responsible for the interaction with the user regarding the selection of a file to save to. Anything we add to alleviate the problem could be considered worse than the original problem due to its nature as a bandaid.

To make sure I'm not misremembering, I just checked the manual:

QFileDialog::DontConfirmOverwrite | 0x00000004 | Don't ask for confirmation if an existing file is selected. *By default confirmation is requested.*
-- | -- | --

We do not send that flag, so by all means it should be asking for confirmation. If it does not, that is a bug on the end of Qt, and not us. (Or it might be some sort of configurable setting in the OS, who knows!)

gedakc:
Wow, you've been busy!  :-)

> Question: why remove this patch from the milestone?

This week I was trying to create the 0.10.0 release and had marked this PR for inclusion assuming that my testing would go well.  However after testing I encounted some issues so I pulled it from 0.10.0 with the intention of still creating the 0.10.0 release this week.

In that way that life has of popping up unexpected challenges I was not able to meet my self-imposed deadline of this week.  Hence the 0.10.0 release will have to wait until I have a large enough window to attempt another release (it takes a few days of prep and extra time to handle packages breaking due to external factors like macOS X dropping support for a version).

> To my knowledge, it fixes the cause of the originally reported issue.

In my earlier testing I observed that this PR fixed the issue for Qt 5.13, but not for Qt 5.5.1.  Following are the tests steps I used, which I should have posted when I tested on Aug 29, 2019.

**Test Steps**

1. Start Manuskript and create a new project based on the **Novel** template
2. Enter some text into **Scene 1**
3. Choose **File -> Compile**.
4. Select **Export to**, set to **Pandoc - HTML**, and choose **Export**
5. Enter **_myprojectname_.html**
6. Choose **Save**
7. Note if prompted overwrite?  _Should not be because first file creation_
8. On still open **Export** window, choose **Export**
9. Don't change filename and Choose **Save**
10. Note if prompted overwrite? _Should be prompted_

| Prompted Overwrite? | Qt5    | PyQt5  | Python | OS                  |
| ------------------- | ------ | ------ | ------ | ------------------- |
| **No** - silently overwrites | 5.5.1  | 5.5.1  | 3.5.2  | Kubuntu 16.04       |
| Yes                 | 5.13.0 | 5.13.0 | 3.7.3  | Debian 10 (pip3)    |

This appears to be another Qt anomaly.

Do you see the same behavior with Qt 5.5.1?

worstje:
Unfortunately I do not have a Kubuntu VM for testing. (Nor a XP one, for that matter.) This makes my testing Qt 5.5.1 issues kind of useless, since the only reason we support that OS is for those older versions.

## What do we support, anyways?

That having been said, [Kubuntu 16.04 has left its LTS period](https://kubuntu.org/news/trusty-14-04-lts-end-of-life-and-end-of-kubuntu-support-for-xenial-16-04-lts/). Do we still want to have Kubuntu as a liability?

Mind you, we don't want Windows XP as a liability either. But I think the only reason many programs still deal with that is because it just won't die. [Even now, it still has a 2.84% market share.](https://netmarketshare.com/operating-system-market-share.aspx?options=%7B%22filter%22%3A%7B%22%24and%22%3A%5B%7B%22deviceType%22%3A%7B%22%24in%22%3A%5B%22Desktop%2Flaptop%22%5D%7D%7D%5D%7D%2C%22dateLabel%22%3A%22Trend%22%2C%22attributes%22%3A%22share%22%2C%22group%22%3A%22platformVersion%22%2C%22sort%22%3A%7B%22share%22%3A-1%7D%2C%22id%22%3A%22platformsDesktopVersions%22%2C%22dateInterval%22%3A%22Monthly%22%2C%22dateStart%22%3A%222018-09%22%2C%22dateEnd%22%3A%222019-08%22%2C%22segments%22%3A%22-1000%22%7D)

Do we have a policy regarding the OS / library versions we aim to support? XP left active support over five years ago after even that one was extended for far too many years. I think we should really consider kicking older operating systems / distros to the curb and formulate some clear limitations.

Besides simplifying the code, it will severely lessen the testing burden that you especially keep tackling.

gedakc:
First, thank you @worstje for all of your work on Manuskript.  You are a major reason why Manuskript is moving forward at a quick pace.  Fixing bugs like this will help us attract more users, which can attract more developers, which can lead to much improved software.  :-)

> **What do we support, anyways?**

That is a good question.  We should be clear on what is supported.

My goal with the following comments is to explain my position, not to convince everyone that my position is the only valid one.  After all there are real costs to maintaining support for multiple versions of multiple OSes with multiple libraries.

**Disclaimer - Personal opinion follows**:

I have a personal pet peeve about software that drops support for an OS immediately after it is unsupported by it's producer.  Not everyone will share my opinion, but I make it a habit to maintain software support for older OSes as long as the effort of doing so does not add an unreasonable burden.  This is of course subjective as to what an "unreasonable burden" is, but I know it when I encounter it.

> Unfortunately I do not have a Kubuntu VM for testing.

The problem can be replicated with other VMs, such as **U**buntu 16.04.  With that said I do not expect developers to have all software versions.  Because I currently use Kubuntu 16.04 which uses Qt 5.5.1 I will discover problems that are exposed with Qt 5.5.1.

> That having been said, Kubuntu 16.04 has left its LTS period. Do we still want to have that unsupported Kubuntu version as a liability we target?

I say yes.

While Kubuntu 16.04 (released Apr 2016) has left the support window for KDE components, the kernel and most of the other tools are still supported because they are part of the larger [Ubuntu 16.04 which has support to Apr 2021](https://en.wikipedia.org/wiki/Ubuntu#Releases).  In my opinion it would be unwise to drop support for the *buntu 16.04 series while it is still within it's 5 year support window.

Some other GNU/Linux releases have an even longer support window (about 10 years), such as [CentOS](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule).

> Upgrading is free...

Every upgrade I've done involves at a minimum my time (not free) and usually breaks something which involves more effort to discover a solution.  Personally I try to perform major OS upgrades no more than once every four years.  This enables me, and the many computers I support, to remain within the *buntu 5 year long support window.

> Do we have a policy regarding the OS / library versions we aim to support?

Not yet, but I think we should come up with one.

I think we should **try** to ensure that Manuskript will run on the currently supported versions of many major GNU/Linux distributions (for an example see the support statement on [GParted](https://gparted.org/development.php) about 2/3's down the page), and vendor supported versions of Windows and macOS X.  In some cases we might not be able to do so, but we can look at these on a case by case basis.

> XP left active support over five years ago after even that one was extended for far too many years.  I think we should really consider kicking older operating systems / distros to the curb and formulate some clear limitations.

I have to admit that supporting an old OS, such as Windows XP, has been screaming "_stop supporting me_" for a while now.  With that said, I have some personal reasons why I continue to build packages for Windows XP.
- I own a copy of Windows XP
- I still use an old [Acer Aspire One](https://en.wikipedia.org/wiki/Acer_Aspire_One) Netbook from 2008 that when released came with either GNU/Linux or Windows XP pre-installed.
- While Windows XP can no longer be considered secure for surfing the Internet, it is still useful for writers with old netbooks that run Windows XP.
- I think it is a waste to throw out these netbooks just because they won't run a modern version of Windows.

> Besides simplifying the code, it will severely lessen the testing burden that you especially keep tackling.

I agree that we should try to keep the code as clean and maintainable as is reasonably possible, and to avoid overly complicating the code.

I think in the case of this PR we can use the compromise of avoiding complicating the code by using the existing built-in Qt support for this "don't accidentally overwrite" feature, with the realization that it doesn't work correctly with an older Qt version such as 5.5.1.

With this compromise in mind, I will continue to review this PR.

gedakc:
Following is a summary of my test results using the "Test Steps" posted earlier.  Using these steps I was trying to find the lowest Qt version wherein this PR works as expected.

| Prompted Overwrite? | Qt5    | PyQt5  | Python | OS                  |
| ------------------- | ------ | ------ | ------ | ------------------- |
| **No** - silently overwrites | 5.5.1  | 5.5.1  | 3.5.2  | Kubuntu 16.04       |
| Yes                 | 5.7.0  | 5.7    | 3.5.2  | Fedora 25           |
| Yes                 | 5.7.1  | 5.8.1  | 3.6.1  | Fedora 26           |
| Yes                 | 5.9.1  | 5.9    | 3.6.2  | Fedora 27           |
| Yes                 | 5.13.0 | 5.13.0 | 3.7.3  | Debian 10 (pip3)    |

Note that I tried Fedora 24 (no longer supported) because it included Qt 5.6, but did not investigate further because Manuskript crashed on startup.

Based on the above, this PR fixes the compile/export silent overwrite problem by prompting the user if they wish to overwrite the file for Qt5/PyQt5 versions 5.7 and higher.

If no concerns are raised then I will merge this PR in the next day or so.

On the topic of "_What do we support, anyways?_", I have added a draft support statement to our [Wiki](https://github.com/olivierkes/manuskript/wiki#developing).

worstje:
> **Disclaimer - Personal opinion follows:**

I will bear all risks of reading and now responding with my own. 😉 

> I have a personal pet peeve about software that drops support for an OS immediately after it is unsupported by it's producer. Not everyone will share my opinion, but I make it a habit to maintain software support for older OSes as long as the effort of doing so does not add an unreasonable burden. This is of course subjective as to what an "unreasonable burden" is, but I know it when I encounter it.

I am in complete agreement. I have tried for the longest time to keep most of my software Windows XP compliant; I even have an unfinished branch of Manuskript lying around where I tried (and failed) to improve a particular scenario.

> The problem can be replicated with other VMs, such as Ubuntu 16.04. With that said I do not expect developers to have all software versions. Because I currently use Kubuntu 16.04 which uses Qt 5.5.1 I will discover problems that are exposed with Qt 5.5.1.

I had not looked at any other LTS editions that are still supported, and you make a great point that there is apparently still a supported OS where Qt 5.5.1 is a reality we have to live with. I am glad you look a bit further into this than I did. 😄 

#### But as a developer ...

I have done a couple of minor patches so far. Admittedly, none of it is very big or extensive, and in some cases I haven't quite dug as deeply as I perhaps should, mostly in part because of my unfamiliarity with both the Manuskript codebase and Qt. But the more I do, the more I realize one thing...

**Supporting older Qt versions is like fighting a 🔥 dumpster fire 🔥, and doing so for multiple operating systems while using a glue library like PyQt is like actively adding fuel and oxygen to the mix to make it explosive.**

(Qt developers, if you read this: I appreciate the thankless task you do. But Qt is a very fertile place for bugs; even comparing it to the immature .NET libraries when I coded against them almost a decade ago and their even more flaky performance on XP I find the Qt toolkit to be in a class of its own.)

On several occasions thus far, I have checked out either the Qt 5.13 or Qt 5.0 codebase to see how particular things are implemented, just to hope and try to figure out what the hell is causing some behaviour. I have also dug into the PyQt sourcecode on a few occasions - that one in particular is very lacking when it comes to its documentation.

PyQt 5.13.0 was built with a broken OpenSSL library on Windows x64, by the way. I'm hoping they'll fix it some time soon.
PyQt 5.11 and 5.12 have that random problem where adjusting widgets can cause it to crash.
PyQt 5.5 apparently has this issue with overwriting checks in the file dialog...

And these are just the version-specific issues that come to mind!

Every single issue tackled is another journey into the Qt rabbit hole. I don't mind bugs, but I am here to try to improve Manuskript, not to tussle with Qt for tiny little victories! Yet almost every single commit I do seems to have some sort of workaround in it because some part of Qt breaks in some version or other somehow.

### My thoughts on support...

**Fact**: Manuskript is still in version 0. This effectively means it is still in beta since 1.0 is usually the 'big release of a mature product'.

**Arguments**

1. There is no real support commitment on our part, nor serious expectation from anyone using beta software. Any support offered is on an as-is basis.

2. Before an official release, a program is in an especially high state of flux. Features get added, changed and even removed. It is unrealistic to try and support everything, so if we support something, supporting only the latest released version makes most sense.

   - It is a lot easier to develop if backwards compatibility support is not your number one priority from the very beginning. To cook an omelet, you need to break eggs. Nevermind the testing burden...

**Support proposal**

 - While in beta, breaking changes are allowed. If this impacts user work, we need to make this impact as small as reasonably possible. Users would always be recommended to create backups either way.
   *I know there's some plans floating around to change the way a number of things are stored and handled, and I think some of those do not allow conversion from old to new.*

 - For development purposes, functionally target only non-LTS releases.

   - This allows the scope of Qt releases to work with to be reduced to a more managable level.
     *Supporting 3 years of Qt is already too much, nevermind 10 years...*

   - For LTS releases, commit to a policy of 'no regressions'. A reasonable effort will be given to keep everything working, but newer features may not be implemented completely.

      - Any issues only affecting LTS may be resolved by putting it on a 'Known issues' list when a proper fix is too difficult, take too much effort or impact the codebase too heavily for the impact of the issue itself.

   - Once out of support, it is not supported. We can maintain unsupported downloads for old operating systems for people to use at their own risk.

      - If newer releases break older versions, at least the older version still exists.
        *If that old version worked fine in the past, it will work fine tomorrow.*

 - For support purposes, only support the latest released version of Manuskript.

   - This means the latest official version for the given platform.

   - It also includes source-code installations of the most recent develop branch commits.

**Why so limited, especially compared to something like GParted?**

* We have relatively few users, and those we do are bound to be less tech-savvy.
   - An average unix user might expect a program to work on every configuration imaginable but most of our targets will be satisfied by just having a build.

* The state of our codebase is nowhere near release quality.
    - We have far too much to do to improve Manuskript, and far too few people with the will and time to do it. The best way to get developers interested is by making contributions less of a chore. It is one thing to know you need to fix your patch so it works on another OS, but another thing to do so for something that is 5 years old and is also a huge pain in the ass.

* The scope of intended functionality is less clear compared to GParted.
    - We have fiction writers, technical writers, dabblers and everything inbetween as our users.
    - Some people write Markdown, others prefer tex and yet others have other desires.
    - One person might want to export solitary sections or documents, whereas another might want to compile a whole book together.

By using GParted as a comparison does not mean I want to throw shade on GParted or suggest it is by any means inferior. It is rather the opposite: I think it is probably in a far better state volunteer, development- *and* functionality-wise than Manuskript. The fact it finally went for the 1.0 this year is probably a great example of that.

If I were to compare them, then GParted is a really fancy set of screwdrivers that is great at anything involving screws, while Manuskript aims to be an entire makerspace that allows every bit of writing-related creativity to happen. Neither is inferior, but they are completely different... but when it comes to being fit for its intended purpose, Manuskript is not even a contender in my mind.

gedakc:
Thanks @worstje for the thought and consideration you put into your response.  You make a number of good points.  I think we can agree on some common middle ground.

My intention is neither to hold up development, nor to prevent new features being added.  I think it is reasonable to expect that some new Manuskript features might not work on older LTS releases.

Because I use Kubuntu 16.04 which has Qt 5.5.1, I will continue to test with this combination.  Next year I plan an upgrade to the future 20.04 LTS version.  'Hopefully it will not have Qt 5.11 or Qt 5.12.  ;-)

I will rebase and merge this PR and then move on to the next PR.

worstje:
Since it briefly came up here: I have (finally) narrowed down the problem with the OpenSSL library and adjusted the 'Known Issues' section with the relevant information so that people can work around it.

Issue #518, opened 2019-02-28
Plot Display as Index Cards


siliconserf:
Enhancement Request: Editing allows displaying documents as index cards. This could be very useful in managing plot entries for each sub-plot. I would like to see the cards stacked vertically, like cascaded windows, with each column assigned to an Act, but that would require a lot more code.

TheJackiMonster:
Can you make a sketch for it?

siliconserf:
Here is how Plottr, an outlining program, shows scene cards in a timeline. Note Act and Chapter headings across the upper part of the display, with "cards" associated with a given heading stack vertically and show only their titles unless you mouse over the card, when a description pops up as a bubble. Editing the order is done by dragging a card from one location to another. Click between or below a card and you can insert a new scene card. It's a very useful, even powerful approach to managing scenes. Even if you show an entire summary on an "index" card image, I think it would still constitute a very helpful feature.

![Screenshot 2021-02-22 123430](https://user-images.githubusercontent.com/32206614/108766753-646e3900-750a-11eb-939d-c4a54e202d05.jpg)

Here's Scrivener's approach. Not as good as Plottr, but has some advantages. Note the indentations on the left help orient with the rest of the outline.

![Scrivener Chapter corkboard stack 2021-02-22](https://user-images.githubusercontent.com/32206614/108769872-7356ea80-750e-11eb-93fd-147ec97f35b2.jpg)

PR #615, opened 2019-08-14
Do not prompt "Save project?" when _Save on quit_ setting enabled

gedakc:
The Travis CI tests began failing after merging Pull Request #583.

Log snippet:

```
...
Ref not implemented
PASSED
manuskript/tests/ui/test_welcome.py::test_autoLoad QXcbConnection: XCB error: 8 (BadMatch), sequence: 613, resource id: 2097162, major code: 42 (SetInputFocus), minor code: 0
QXcbConnection: XCB error: 8 (BadMatch), sequence: 619, resource id: 2097168, major code: 42 (SetInputFocus), minor code: 0
QXcbConnection: XCB error: 8 (BadMatch), sequence: 625, resource id: 2097171, major code: 42 (SetInputFocus), minor code: 0
```

When I run the "pytest -vs" locally, a command used in our .travis.yml
file, a dialog to "Save project?" is displayed.  Because the test
scripts use the "saveOnQuit" default setting of *enabled*, the
"Save project?" dialog should not be displayed.  Hence it appears that
we have a logic issue when closing the project in Travis CI.

In other words when a call is made to close the project, we are
incorrectly prompted "Save project?" because the dirtyProject flag is
set, but so too is saveOnQuit set to True.  What should happen is an
automatic save with no prompt.  This PR fixes this logic problem so
that the Travis CI test suite completes successfully.

This PR ready for review

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -682,8 +682,10 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             return
 
         # Make sure data is saved.
-        if not self.handleUnsavedChanges():
-            return  # user
+        if (self.projectDirty and settings.saveOnQuit == True):
+             self.saveDatas()
+        elif not self.handleUnsavedChanges():
+             return  # user cancelled action
 
         # Close open tabs in editor
         self.mainEditor.closeAllTabs()
--- a/manuskript/settings.py
+++ b/manuskript/settings.py
@@ -138,6 +138,7 @@ def save(filename=None, protocol=None):
         "openIndexes": openIndexes,
         "autoSave":autoSave,
         "autoSaveDelay":autoSaveDelay,
+        # TODO: Settings Cleanup Task -- Rename saveOnQuit to saveOnProjectClose -- see PR #615
         "saveOnQuit":saveOnQuit,
         "autoSaveNoChanges":autoSaveNoChanges,
         "autoSaveNoChangesDelay":autoSaveNoChangesDelay,
--- a/manuskript/ui/settings_ui.py
+++ b/manuskript/ui/settings_ui.py
@@ -1854,7 +1854,7 @@ class Ui_Settings(object):
         self.label.setText(_translate("Settings", "minutes."))
         self.chkAutoSaveNoChanges.setText(_translate("Settings", "If no changes during"))
         self.label_14.setText(_translate("Settings", "seconds."))
-        self.chkSaveOnQuit.setText(_translate("Settings", "Save on quit"))
+        self.chkSaveOnQuit.setText(_translate("Settings", "Save on project close"))
         self.chkSaveToZip.setToolTip(_translate("Settings", "<html><head/><body><p>If you check this option, your project will be save as one single file. Easier to copy or backup, but does not allow collaborative editing, or versionning.<br/>If this is unchecked, your project will be save as a folder containing many small files.</p></body></html>"))
         self.chkSaveToZip.setText(_translate("Settings", "Save to one single file"))
         self.lblTitleGeneral_2.setText(_translate("Settings", "Revisions"))
--- a/manuskript/ui/settings_ui.ui
+++ b/manuskript/ui/settings_ui.ui
@@ -407,7 +407,7 @@
              </font>
             </property>
             <property name="text">
-             <string>Save on quit</string>
+             <string>Save on project close</string>
             </property>
             <property name="checked">
              <bool>true</bool>


worstje:
I am very confused right now. I am sure I accounted for that setting being off to prevent asking... and I did in `closeEvent()`:

https://github.com/olivierkes/manuskript/blob/34f84126dfcda2311bb84b4a0c0335701c62d177/manuskript/mainWindow.py#L767-L771

Your change adjusts `closeProject()`, which is not in fact a 'quit of the program', and as such should not be short-circuited like that.

As such, I feel your fix may be in the wrong location. Why are you adjusting the code, as opposed to adjusting the test? What is it the test tests? Does it test closing the project, or does it test closing the window? IMHO the menu option 'Close Project' should never ever auto-save, but that is what your current patch would accomplish.

I am sorry for not having run the test suite before to where you had to write this patch - I honestly had completely forgotten it existed and is actually updated&useful (due to the automatic checks on github always failing) - but based on what little I know and see in this issue, I think I do not approve of the patch. The PR introduced new functionality, yes.. but that means the test case needs to be adjusted to deal with the new functionality.

(If the code is in fact broken, then by all means, adjust it.. but this seems like a test case that needs updating to me.)

gedakc:
Thanks @worstje for reviewing and responding to this PR.  I make my share of mistakes so I appreciate other viewpoints.  With your collaboration we should be able to make this fix better than either one of us could do alone.

Following are my responses.

> Your change adjusts `closeProject()`, which is not in fact a 'quit of the program', and as such should not be short-circuited like that.

I think you've pin pointed where our approach to this fix differs.  It boils down to the confusion caused by different interpretations for the meaning of the Manuskript project setting **Save on quit**.  Essentially, is this a _program_ setting, or is it a _project_ setting?

Save on quit Setting

- One interpretation is that **Quit** means quitting/closing/exiting the Manuskript _program_.

- Another interpretation is that **Quit** means quitting/closing/exiting the Manuskript _project_.

Either interpretation could be correct.  In order to determine if the **Save on quit** setting is _program_ or _project_ specific I did the following test.

1. Start Manuskript, create Novel _ProjectA_, ensure **Save on quit** is _enabled_, save the project, and exit Manuskript.

2. Start Manuskript, create Novel _ProjectB_, set **Save on quit** to _disabled_, save the project, and exit Manuskript.

3. Start Manuskript, open _ProjectA_ and check the value of **Save on quit**.  It is _enabled_.  Exit Manuskript.

4. Start Manuskript, open _ProjectB_ and check the value of **Save on quit**.  It is _disabled_.  Exit Manuskript.

Based on these tests I conclude that the **Save on quit** setting is _project_ specific.  This leads to my interpretation that when **Save on quit** is _enabled_ then the project should be saved whenever quitting/closing/exiting the _project_, not just when quitting/closing/exiting the _program_.


QUESTION 1:  Do you agree that the **Save on quit** setting is _project_ specific?

If you do not agree then we should try to land on a common understanding of the meaning of **Save on quit**.

Otherwise, proceeding onward....

> As such, I feel your fix may be in the wrong location. Why are you adjusting the code, as opposed to adjusting the test?

There might be a better location and I am open to suggestions.

Originally I also thought that the test should be adjusted.  I later changed my mind based on the reasoning explained in later paragraphs.


> What is it the test tests? Does it test closing the project, or does it test closing the window?

My interpretation is that the test suite performs a variety of unit tests.  At least one of these test scenarios involves closing a project by calling `closeProject()` directly.  For example see [`manuskript/tests/ui/test_welcome.py`](https://github.com/olivierkes/manuskript/blob/develop/manuskript/tests/ui/test_welcome.py) which calls `MWNoProject(MW)` in [`manuskript/tests/conftest.py`](https://github.com/olivierkes/manuskript/blob/develop/manuskript/tests/conftest.py).  More specifically:

https://github.com/olivierkes/manuskript/blob/34f84126dfcda2311bb84b4a0c0335701c62d177/manuskript/tests/conftest.py#L19-L28

Notice the direct call to `closeProject()`.

To determine what the behaviour of the test should be I considered the following:

- The existing test suit leaves the **Save on quit** setting at the default value of _enabled_ .  I state this because I did not find any occurrences of the text string `saveOnQuit` in the code in the `manuskript/tests` folder.

- Prior to merging PR #518 "Track dirty state", the test suite would run to completion with no manual intervention and the project was saved.  I tested and confirmed this by reverting back to the [commit](https://github.com/olivierkes/manuskript/commit/60b6f98c21893375a54911d512fd0f88ef375ff7) immediately before PR #518 was applied.

- PR #518 adds a new feature of the "Save project?" dialog and a new way for handling when **Save on quit** is _disabled_.

- None of the tests in the Manuskript test suite have been updated, so **Save on quit** is _enabled_, never _disabled_.

- Because **Save on quit** has not been _disabled_ there should be no situation whereby the "Save project?" dialog is displayed.

- However, with PR #518 applied, the "Save project?" dialog is indeed displayed.

Example `pytest -vs` session in Debian 10 Buster GNU/Linux.

Manuskript-pytest-vs-with-pr583-applied

> IMHO the menu option 'Close Project' should never ever auto-save, but that is what your current patch would accomplish.

This comes down to whether the **Save on quit** setting is _program_ or _project_ specific.

I would agree with the statement _if_ the **Save on quit** setting is _program_ specific.

However the above tests and logic point to the **Save on quit** setting being  _project_ specific.


QUESTION 2:  Does the above make sense?

Please feel free to ask for further clarification or to point out errors in my thought process.

gedakc:
Any more comments?

If not I then I plan to merge this PR with the **develop** branch.  Currently the Travis CI tests and package builds are broken which prevent any future releases of Manuskript.  The current options to fix this are to either merge this PR, or to revert PR #583.

worstje:
Well... crap. I could have sworn I replied here. I have this vivid memory of writing a post and all that stuff... but maybe I never finished writing it. Whoops.

Anyway, to summarize up the stuff I thought a couple of days ago: if this is the solution you want to go with, I am fine with it if you rename 'Save on Quit' to 'Save when closing project' as that is a better description of what the setting does.

Functionality-wise, I would prefer 'save on quit' because it makes more sense to me as a setting and is probably more common.

But since you do not wish to deviate from existing behaviour, changing the way the setting is labeled seems the best option to me.

gedakc:
What do you think of renaming "Save on Quit" to "Save on Close / Quit"?

My intention is for this to indicate a save will occur if the project is closed or if Manuskript is quit.

worstje:
I feel that is too... confusing, honestly. 'Close the program' and 'quit the program' are sentences that mean the exact same thing in typical language. So they still won't understand that it also affects them when they close the project through the File menu.

Quitting the program means you also close the project, so to me my suggestions covers all bases.

But if you prefer to have both scenarios spelled out, then I could agree with 'Save when closing the project or the program', but I feel that is so overly wordy that it might confuse users more instead. 

Setting names are typically kept brief and simple: it makes it easier for users to figure out what they do when scanning over it trying to find what they wish to adjust. Listing several different scenarios on the checkbox control itself doesn't help with such discovery or accessibility IMHO.

I do not think most users require the degree of 'spelling things out' in the Settings screen that you seem to want to offer. If it requires so much detail to succinctly describe every scenario it affects, that should be something to put in the documentation instead, right?

(My apologies if I sound a bit picky in regards to your suggestions in this issue; I feel the relative simplicity of the current Settings screen is one of its few strong points right now. I do appreciate you put in so much attention to all the details!)

gedakc:
I appreciate your feedback.  Wording can be tricky and it is preferable to minimize string changes as these cause extra work for translators.

To keep the string short and to try to match the proposed suggestion "Save when closing project" my preference would be for something like "Save on project close".

Unfortunately either setting name change has potentially large implications.

Both of these proposed changes involve dropping the word "quit".  In my mind this means we  should also consider changing the setting name in the program code.  For example `settings.saveOnQuit` to `settings.saveOnProjectClose`.  Unfortunately this has ripple effects that affect the current values used in the `projectname/settings.txt` file.  This in turn should force a file format version bump from our current version "1" to something higher.

The downside is a file format change would means a loss in backward compatibility -- once a project is opened and saved with the new `settings.saveOnProjectClose` then it can no longer be opened with older Manuskript versions.  This seems a high price to pay for trying to fix a broken Travis CI test/build system when we are keeping the actions of _save on project close_ and _save on manuskript quit_ the same as before PR #583. 

To sum up, my current thought is to leave the wording as currently exists and to merge this PR as is.

worstje:
Honestly, I believe we can skip the rename in the code itself at this point in time. Just making sure it is properly labeled towards the users is good enough. Why do I think that?

At some point or other, someone is going to have to go through the Settings dialog in a thorough manner and all the other places the program saves settings of some kind, and clean things up for the sake of both maintainers and users. And this stage would definitely break with backwards compatibility to some degree, but that would be hard to avoid. Besides, if it ever happens, it is better to break backwards compatibility things in big chunks before reaching 1.0 than a dozen times smaller at any other point, which would draw out the proverbial suffering.

As such I feel it is good enough right now as long as the functionality and labeling is fixed. Breaking changes, refactoring and so on can (and hopefully will) come later as the code health of Manuskript as a whole is improved.

While I definitely value backwards compatibility, I also believe our codebase is not in the prettiest of states right now. Comments are rare and the structure of the code and logic is not immediately apparent at all, while many functions are spread haphazardly over files in places where someone was working at the time. Right now, we are setting our sights on rather immediate things: bug fixes, stability fixes and minor quality of life improvements for users, but that cannot go on forever. Eventually, we want to start to truly add new and exciting things again; we definitely have enough issues to cover.

But if we want to attract more contributors to work on those things in the long term, giving off a signal that places 'backwards compatibility' first might be demotivating to them instead.

I kinda strayed from the original subject here (sorry, I do that a lot!), but all this has been floating around in my mind for a while. 😄 Some of the things I mention here have been things I've been tempted to tackle or put onto mental TODO lists, but being unfamiliar with Qt and the codebase itself as well as the change-conservative tone in code reviews have made me very hesitant on actually taking the plunge on doing things that might not be very welcome to begin with. After all, nobody welcomes someone unaskedly touching almost every single source file to rename variables, move functions around in files and change established behaviour willy nilly. 😁

gedakc:
I think we are mostly on the same page.

I agree that a code cleanup would be welcome, and that making the settings explicitly clear would help our users (for example we should be clear which settings are application specific and which are project specific.)

I also admit that I am conservative when reviewing changes.  With that said I am open to refactoring, and also agree that breaking changes should be grouped together in big chunks to minimize the number of times we adversely impact our users.

I've changed the "Save on quit" setting string, left the actual setting name/variable unchanged, and made a TODO note as a future _Settings Cleanup Task_.

Assuming that the Travis CI test suite completes successfully, do you have any other concerns before this PR is merged?

worstje:
Nope, nothing at all. All is good! I am supremely happy. You are rid of me. I never want to post in this PR again! 😉 

Thank you for all the effort you have put into making my originally simple PR completely worthy of the project. Next time I'll make sure to run the test suite through as well so you don't have to do it again.

gedakc:
There shall be no need to post in this PR again because I will rebase and merge it pronto.  :-)

PR #624, opened 2019-08-31
change markdown to "Markdown" in....

leela52452:
...A universal document converter. Can be used to convert markdown to a wide range of other formats.
see https://en.wikipedia.org/wiki/Markdown

PR #628, opened 2019-09-06
typo ellipsis, added

leela52452:
[source](https://hosted.weblate.org/projects/manuskript/translations/source/review/?type=check%3Aellipsis)

### DIFF
--- a/manuskript/exporter/pandoc/__init__.py
+++ b/manuskript/exporter/pandoc/__init__.py
@@ -17,7 +17,7 @@ from manuskript.functions import mainWindow
 class pandocExporter(basicExporter):
 
     name = "Pandoc"
-    description = qApp.translate("Export", """<p>A universal document converter. Can be used to convert markdown to a wide range of other
+    description = qApp.translate("Export", """<p>A universal document converter. Can be used to convert Markdown to a wide range of other
     formats.</p>
     <p>Website: <a href="http://www.pandoc.org">http://pandoc.org/</a></p>
     """)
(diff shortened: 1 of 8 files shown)


gedakc:
While reviewing this PR I came across several issues:

- This PR seems to also include changes made in PR #624 and hence is a duplicate.  
  To resolve either close one of the PRs, or else remove the duplicated changes from this PR.  Note that PR's can be amended and re-pushed with the --force option.
- Automatic checks can make mistakes.  In this case I think at least one of the sentences contains mention of three periods "..." and should remain as three periods.  
  - `Replace ... with …`  
    _When enabled this automatic translation ability in Manuskript will convert the writer's three periods into an ellipses._

I highly recommend reviewing and testing all changes before submitting.

gedakc:
In my search for Qt standards I came across the following document:

[User Interface Text Guidelines](https://doc-snapshots.qt.io/qtcreator-extending/qtcreator-ui-text.html)

Following is a quote from these guidelines:

> **Punctuation**
> 
> Avoid using punctuation marks or special characters in menu names and items.
>
> Use punctuation marks as follows:
>
>  - Use full stops in messages.
>  - Never use full stops (.) at the end of menu item names.
>  - Place three full stops (...) at the end of menu item names that open a dialog requiring user action.
>  - Use exclamation marks (!) only in text that demands extra attention from the user or carries special weight.
>  - Use quotation marks ("") around variable values. For example, Close Project "qtcreator". For consistency, use double quotes to emphasize or set apart file names, directory names, URLs, and so on, in user visible strings.
>  - Do not use leading, trailing, or multiple spaces to align text in messages, as translation tools might not handle them correctly.

From reading the above I understand that ellipses are to be avoided in menu names and items.

If you have discovered documentation to the contrary then please share your findings.

gedakc:
For the reasons outlined in an earlier post (_E.g, place three full stops (...) at the end of menu item names that open a dialog requiring user action and don't use ellipses_) I am closing this PR and labeling it as **wontfix**.

Issue #593, opened 2019-06-05
Importing images into Manuskript


scribas:
I tried to follow the instructions on the wiki page, for referencing the images inside the images folder: `![My Image One Title](images/myimage1.jpg)`. Firstly, it complained that it does not understand the protocol. I also tried various combinations in linux. I finally had to put the full path in, with a `file://` upfront: `![My Image Title](file:///home/scribas/git_repos/manuskript/images/myimage1.jpg)`. I suspect that the reason for this is, that the present working directory is in the folder where the manuskript binary is kept?

Aside from this, if I put the correct reference into one of the pages within the `Editor` environment and hover over the link with my mouse, the image pops up. However, if I try and do this in the `Characters` environment, the image only flickers up for a second and then disappears again. As I move my mouse around over the link, the image will keep on flickering on and off very rapidly. It would be great if this could be improved?

gedakc:
Following are some high level ideas for how to proceed:

> I suspect that the reason for this is, that the present working directory is in the folder where the manuskript binary is kept?

You might try to confirm your suspicion by placing the image resources in the same folder.  If this works then it might be a case of needing to prefix the currently open project folder with the image path.

You might try tracing the path through the source code to discover the section where images are displayed.  As a hint for a starting place, the Manuskript main window is described in the `[manuskript/mainWindow.py](https://github.com/olivierkes/manuskript/blob/develop/manuskript/mainWindow.py)` python file.

For the second paragraph in the initial issue report, you might try to locate the code that displays an image while in the **Editor** pane, then find how this works in the **Character** pane, and then compare the two.

scribas:
Awesome, that is a good place to start. Will start looking at
this tomorrow.

scribas:
Feedback to date:
- I put a sample image into the same location as manuskript.exe on Windows and referenced it with: `![My Image One Title](file://example.jpg)`. That crashed Manuskript completely, without any error messages. 
- I am still struggling to understand where the image gets called within the tool tip, when the mouse hovers over the image reference. Looks like I may have to do some reading up on how Qt works.
- In general, it would be helpful if there was a way to map out the files and folders within a development project like Manuskript and explain a bit of the developers thinking, to make it easier for newby developers to find their way. Is there such a mapping thing / documentation framework, aside from putting it in special wiki articles in Github?

gedakc:
[...]

PR #629, opened 2019-09-06
Further refinement of image tooltips (issue #593)

worstje:
Last time I touched this code, I went in looking for a specific problem,
and came out with a fix specific to that issue. That fix was not wrong,
yet it hardly covered all the problems present in the code once one took
into account issues like:

- local vs remote resources,
- relative vs absolute paths,
- different operating systems behaving differently, and
- Qt being uniquely buggy on different platforms.

The major part of it was fixed by using `QUrl.fromUserInput()`, which does
the exact kind of auto-detection for the nature of the resource that we
were in need of.

The rest was fixed by creating a whole lot of test cases and fixing
problems as I came across them, switching between Windows & Ubuntu about
a dozen times till I got it right. These will be included in the PR.

An unfortunate wart left behind is that I have not found a satisfying
way to ensure the processing pipeline remains clean. Some code paths
unfortunately cause the original request QUrl() to be changed during
processing, and without that it is very difficult to know what original
url we were fetching to begin with. (The odd QNetworkRequest behaviours
that I have observed make me hesitant about rewriting the processing
code right now when all my test cases finally work on Ubuntu & Windows.)

The test cases: [Image link tooltip test sheet.txt](https://github.com/olivierkes/manuskript/files/3582420/Image.link.test.sheet.txt)

I'll see if I can get rid of that final wart tomorrow when I am properly awake and not sick of this code anymore. I put up the PR so that those involved in the original issue (@scribas & @gedakc ) can verify whether this patch indeed fixes all the issues they have come across when dealing with image tooltips.


worstje:
I decided not to squash these two commits together because having them separate helps set this last one apart from the other Linux-related Qt that was already in the patch. 😄 

[Image link tooltip test sheet.txt (updated to match new patch)](https://github.com/olivierkes/manuskript/files/3585600/Image.link.test.sheet.txt)

This should be the final version of the image tooltip code. (Famous last words...)

gedakc:
In my testing on Kubuntu 16.04 with Qt 5.5.1 this PR is a definite improvement over the crash that would previously occur

I agree with your assessment to let Qt handle all the OS specific stuff so that we can keep the code simpler and more maintainable.  Personally I would like if it could resolve a relative URL like "figures/image.jpg" as if it was a file path in the current working directory, but this sounds like it would add undue complexity.

While reviewing the code I noticed the following sentence which can appear in the GUI:

`The given path or URL is utter bogus.`

In order to remove any sense of judgement from the message I suggest we use something like the following:

`The image path or URL is incomplete or malformed.`

When I'm not sure about a sentence I try to follow the [GNOME Human Interface Guidelines for writing style](https://developer.gnome.org/hig/stable/writing-style.html.en).

>  Let me know your preference on how you'd like this mess to be squashed

Please squash this PR into a single commit.

Thanks,
Curtis

worstje:
@gedakc Your suggestions have been implemented and the squash has been performed, and now this PR is ready for merging. 😄

Issue #388, opened 2018-09-16
[Feature request] Night mode/dark mode


lappet:
A lot of editors and websites these days come with a night mode or a dark mode where the background text is not white in color. Can we get that please?

RiderExMachina:
As a lover of dark themes I agree!

gedakc:
Did you know that Manuskript already supports manually changing item colours?

To search through the existing issues raised for Manuskript you can add and subtract terms.  The steps to search all issues for the word "**dark**" are:

1. Click on **Issues** in the [Manuskript git page](https://github.com/olivierkes/manuskript/issues).
2. Remove "`is:open`" from the **Filters**.
3. Add "`dark`".
4. Press **Enter** key.

This should create the following [list of issues which contain the word "**dark**"](https://github.com/olivierkes/manuskript/issues?utf8=%E2%9C%93&q=is%3Aissue+dark).

See [issue 183 - Index card text almost invisible in dark themes](https://github.com/olivierkes/manuskript/issues/183) for an example of how some users have configured dark themes.

Does that answer your question?

HorsaTheHetrisiniLord:
No actually... I don't see exactly where it says how to configure a dark theme in that issue...

I'd much rather have an official dark theme instead of manually having to configure it.

gedakc:
I believe the dark theme in the issue #183 mentioned above was selected at the desktop level.

For example in Kubuntu 16.04 I can select "**Breeze Dark**" under **K -> Settings -> System Settings -> Workspace Theme -> Look And Feel**.

![kde-workspace-theme](https://user-images.githubusercontent.com/10405019/46682623-473fa680-cbab-11e8-9914-20415cd05f35.png)

Manukript then uses the chosen workspace theme.

You can further refine the text editor colours using Manuskript's **Edit -> Settings-> Views -> Text editor**.

![manuskript-text-editor-settings](https://user-images.githubusercontent.com/10405019/46682717-8d950580-cbab-11e8-9b3f-8f8e7c701e40.png)

gedakc:
Comment from AlucardNoir copied from [Manuskript 0.7.0 web site page](http://www.theologeek.ch/manuskript/2018/08/15/manuskript-0-7-0/):

[...]

PR #630, opened 2019-09-06
Basic dark theme support (Windows 10)

worstje:
Windows 10 has supported a 'dark theme' option for a while, and the fact
Manuskript is like looking into a bastion of bright white while using it
is bothersome to say the least.

Since this is a setting defined as the OS level, I believe this should
be something Manuskript automatically adjusts itself to match, thus the
lack of a configurable setting on the Manuskript end.

Note that it does not adjust the main editor area since that is already user configurable.

For as far I have been able to check, it works with every builtin theme that is selectable in the Settings screen and produces legible output. (I would not go as far as to call it pretty, simply because most of them aren't pretty to begin with. 😉 )

I looked into Linux-land for a little while to not make this a Win10-only commit, but there is no separate 'dark theme' switch on Ubuntu at least. And even then there is the mess between Gnome and Qt themes to worry about, which means dragging in a whole new library of which I have already forgotten the name and compiling against that... suffice to say I'll leave any such challenges to the people who improving the themes-related situation on Linux. 😄 )

### DIFF
--- a/manuskript/main.py
+++ b/manuskript/main.py
@@ -2,14 +2,15 @@
 
 import faulthandler
 import os
+import platform
 import sys
 import re
 
 import manuskript.ui.views.webView
 from PyQt5.Qt import qVersion
-from PyQt5.QtCore import QLocale, QTranslator, QSettings
-from PyQt5.QtGui import QIcon
-from PyQt5.QtWidgets import QApplication, qApp, QMessageBox
+from PyQt5.QtCore import QLocale, QTranslator, QSettings, Qt
+from PyQt5.QtGui import QIcon, QColor, QPalette
+from PyQt5.QtWidgets import QApplication, qApp, QMessageBox, QStyleFactory
 
 from manuskript.functions import appPath, writablePath
 from manuskript.version import getVersion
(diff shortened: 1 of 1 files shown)


worstje:
I had cleaned up my comments before sending in the PR, but had forgotten to add the changed source file when amending the original commit, so they weren't actually registered. Whoops.

Btw, for good measure, have a screenshot of what it looks on my end with the dark theme patch enabled:

manuskript-dark-theme

(The editor area that by default has a white background has been manually adjusted to have a dark background.)

gedakc:
I performed regression testing on this PR and no adverse issues were discovered with the following:
- Kubuntu 16.04
- Kubuntu 19.04
- Windows 7
- Windows XP

Further the K Desktop Environment **Breeze** and **Breeze Dark** themes continued to work as expected.

Manusrkipt on Kubuntu 19.04 with **Breeze** theme
Manuskript-Kubuntu1904-Breeze-Theme

Manuskript on Kubuntu 19.04 with **Breeze Dark** theme  
(Note that Text Editor settings were not modified in Manuskript settings which is a user controllable feature in Manuskript already)
Manuskript-Kubuntu1904-Breeze-Dark-Theme

Hence this PR looks good to me.

worstje:
I always assumed that KDE-based distros would work out of the box because they use Qt under the hood.. and it seems they do. 👍 It is nice to see that the dark theme patch for Windows resembles that *Dark Breeze* theme for KDE to such a high degree!

FYI: This patch has a platform check for Windows, and then checks a registry entry that should only exist on a Windows version that supports the 'dark theme' toggle.. so if we had any sort of regression on any other OS, there would have been some really funky business going on. 🤣 

Having seen your screenshots, I am all happy with it!

gedakc:
Great work @worstje.  I'll bet that this enhancement will make many Windows 10 users happy.  We've had many votes for dark theme as can be seen in issue #388.

I will merge this PR with **develop** for inclusion in the next release of Manuskript.

Would you update the [How to Enable Dark Theme](https://github.com/olivierkes/manuskript/wiki/How-to-Enable-Dark-Theme) Wiki entry for Windows 10?

Thanks,
Curtis

worstje:
Done. 👍

PR #489, opened 2019-02-10
Fix misc. typos

luzpaz:
`codespell -q 3 -S *.ts,./libs,./sample-projects -L searchin`

PR #645, opened 2019-09-17
Fix source typo

luzpaz:
Split off from #489


worstje:
This fix looks good for as far I can tell. Thank you. 👍 

(I can't believe nobody triggered that typo; it looks like it would have been a pretty nasty traceback.)

gedakc:
Thanks @luzpaz for placing this code fix in a separate PR.  I will merge this with the **develop** branch for inclusion in the next release of Manuskript.

PR #648, opened 2019-09-19
Fix typo missed in previous commit

luzpaz:
https://github.com/olivierkes/manuskript/pull/489#issuecomment-533195819


gedakc:
Doh!  Even with this commit I just realized we missed one of the "save" -> "saved" occurences.  It is the second instance in the same string.

@luzpaz would you like to create yet another PR to capture the last missed typo?

gedakc:
@luzpaz, I will make this last typo change and push it directly to @olivierkes repo.

PR #654, opened 2019-09-22
Restore progress bar functionality

worstje:
A previous fix (5f9ea3) inadvertently broke the progress bar by
converting to the wrong data type. (See issue #561 / PR #609).

While checking the code I realized the problem occurred primarily
because we weren't checking the validity of the values closer to the
source. Doing so alleviates the need to check elsewhere.

In the hope of inspiring a more systematic approach, a new uiParse()
utility function has been added to curb the further growth of toXxx()
functions that exist solely to validate user input.

There is no doubt room for improvement, both on the end of the new
uiParse() function as well as the spot where it is used. Ideally, the
data that comes out of the model should already be 'safe', but since
this is a bugfix for a bugfix I want to keep waves to a minimum.

This commit fixes issue #652.

PR #655, opened 2019-09-22
Changes to Revisions UI

worstje:
See PR #401, #653, #654 for discussion regarding these changes.

Note that this was the prettiest I was able to get the warning for now. I wanted to add some margins to the label but alas, they are a finnicky thing. The only workaround I could find involves adding a whole new parent widget and layout, but I couldn't quite figure out how the auto-sizing works.

manuskript-revision-ui-changes

Since I don't want to ruin the accessibility for small screens or weird DPIs by hardcoding a fixed widget size, _and_ because this message will hopefully only be there for one version, I figured this solution would be good enough to get the message across.


gedakc:
This enhancement looks good to me.  From my testing on Kubuntu 16.04 the UI changes do indeed prevent setting the "revisions per X" values to zero, and that was the cause of the crash reported in issue #381.

Issue #659, opened 2019-09-30
0.10.0 crash using Windows dark mode


mxamber:
Just downloaded the new 0.10.0 release to both my machines. While it seems to work on my main rig (Windows 7 64-bit), apparently fixing all issues I had experienced with 0.9.0, I can't get it to run on my laptop with the Windows colour scheme set to dark. Starting it results in the following error:

> Debug: Web rendering engine used: QWebView
> Running manuskript version 0.10.0.
> Preferred translation: ['de-DE', 'en-US'] (based on available ui languages)
> Loaded translation: manuskript_de_DE.qm
> Traceback (most recent call last):
>   File "manuskript", line 13, in <module>
>   File "manuskript\main.py", line 231, in run
>   File "manuskript\main.py", line 142, in prepare
>   File "manuskript\main.py", line 115, in respectSystemDarkThemeSetting
> AttributeError: type object 'GlobalColor' has no attribute 'white'
> Failed to execute script manuskript

The machine is running Windows 10 64-bit and a fresh install of Manuskript, having deleted the old instance and installed it anew from the 0.10.0 archive. Deleting it and installing it again didn't fix the issue. Manuskript 0.9.0 previously ran perfectly fine. Changing Windows' colour scheme to bright (which is the default) rather than dark fixed the issue.

worstje:
Weird. That did not come up during testing. (I feel terrible since I wrote that patch...)

I guess it is due to the version of PyQt bundled with the PyInstaller build. I'll look into it.

worstje:
@mxamber I am very sorry that you (and likely others) are affected by this bug. I should have caught that; I thought I was distrustful enough of the version differences as it was but apparently I need to be even more paranoid in the future.

I have submitted PR #660 along with the request that @gedakc creates a new Windows build.

(Unfortunately I can't provide a proper Windows build for you myself at this time; if I could I would have provided a download with the fix included.)

gedakc:
A win32 PyInstaller package with PR 660 applied is ready for testing.

https://sourceforge.net/projects/manuskript/files/manuskript/testing-only/

worstje:
I have tested that new package on my end and the problem has been fixed. 👍

gedakc:
Closing this issue because PR #660 (which prevents the crash) has been merged with the **develop** branch.

Further the test package file has been renamed:  
`manuskript-0.10.0-2-win32.zip`  
and uploaded to the official release areas.

PR #660, opened 2019-09-30
Fix for Windows 10 Dark Theme on older Qt versions

worstje:
QtCore.Qt.GlobalColor does not have any accessors for the predefined
colors on PyQt versions before 5.11 despite the object itself existing.

It would have been nice if the documentation* had mentioned that object
being broken on older versions, but I should have tested with even older
versions of PyQt before submitting the original patch.

Apparently the most supported way to access these colors is through the
Qt namespace itself, but those aren't documented in the slightest. Ugh.

My apologies to all those affected. Fixes issue #659.

*: https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtcore/qt.html#GlobalColor

@gedakc Can you create a new Windows build with this fix implemented and put it on the website? (I cannot currently test it on Qt 5.5 but since it reproduced on 5.8 I assume this bug is fixed.)

I will look into setting up a proper test environment for Qt5.5 on my end in the future so I won't have to depend on you to test it on older PyQt versions anymore. I should have caught this before I submitted the PR. 😢

### DIFF
--- a/manuskript/main.py
+++ b/manuskript/main.py
@@ -112,26 +112,26 @@ def prepare(tests=False):
             darkColor = QColor(45,45,45)
             disabledColor = QColor(127,127,127)
             darkPalette.setColor(QPalette.Window, darkColor)
-            darkPalette.setColor(QPalette.WindowText, Qt.GlobalColor.white)
+            darkPalette.setColor(QPalette.WindowText, Qt.white)
             darkPalette.setColor(QPalette.Base, QColor(18,18,18))
             darkPalette.setColor(QPalette.AlternateBase, darkColor)
-            darkPalette.setColor(QPalette.ToolTipBase, Qt.GlobalColor.white)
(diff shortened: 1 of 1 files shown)


gedakc:
A win32 PyInstaller package with PR 660 applied is ready for testing.

https://sourceforge.net/projects/manuskript/files/manuskript/testing-only/

worstje:
@gedakc It has been properly fixed with that test build. I tested both the original release package and the new one with the fix applied.

gedakc:
Thank you @worstje for confirming that the new package works as expected and prevents the crash.

Further I plan to rename the test package file from:  
 `manuskript-0.10.0-with-PR660-dark-theme-fix-win32.zip`  
to:
`manuskript-0.10.0-2-win32.zip`  
and to upload it to our official release areas.

Issue #666, opened 2019-10-15
Numbers trigger the spell checker.


vithiri:
It seems like numbers are triggering the spell checker.

**Steps to reproduce:**

- Toggle spell checking on
- Type 1234

**Expected outcome:**

The number is not flagged as a misspelled word.

**Actual outcome:**

The number is flagged as a misspelled word.

gedakc:
Thanks @vithiri for the report. I confirmed that the spell checker incorrectly highlights numbers as incorrectly spelled in the current 0.10.0 release and all the way back to 0.4.0, and possibly earlier.

PR #669, opened 2019-10-23
Fix numbers triggering the spellchecker

vithiri:
Fixes the issue found in https://github.com/olivierkes/manuskript/issues/666 and prevents numbers from triggering the spell checker.


vithiri:
This fix really just hides the issue. Closing for now.

worstje:
It was a partial fix for a non-ideal situation. I might have discussed it a bit with you on Discord since you pinged me there, and I may have pointed out some considerations, but it wasn't a total loss.

It would have been fine if you had polished it a bit; bugfixing rarely comes in the form of perfect solutions. A good enough fix (which includes the check for the popup menu) would have been fine.

PR #681, opened 2019-11-03
Adds a simple text find dialog.

kbeecher:
Adds a Find option to the Edit menu. Executing it when focused on the main editor brings up an input dialog allowing the user to enter a search string.

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -9,7 +9,7 @@ from PyQt5.QtCore import (pyqtSignal, QSignalMapper, QTimer, QSettings, Qt, QPoi
                           QRegExp, QUrl, QSize, QModelIndex)
 from PyQt5.QtGui import QStandardItemModel, QIcon, QColor
 from PyQt5.QtWidgets import QMainWindow, QHeaderView, qApp, QMenu, QActionGroup, QAction, QStyle, QListWidgetItem, \
-    QLabel, QDockWidget, QWidget, QMessageBox
+    QLabel, QDockWidget, QWidget, QMessageBox, QInputDialog
 
 from manuskript import settings
 from manuskript.enums import Character, PlotStep, Plot, World, Outline
@@ -26,6 +26,7 @@ from manuskript.ui.about import aboutDialog
 from manuskript.ui.collapsibleDockWidgets import collapsibleDockWidgets
 from manuskript.ui.importers.importer import importerDialog
 from manuskript.ui.exporters.exporter import exporterDialog
+from manuskript.ui.findDialog import FindDialog
 from manuskript.ui.helpLabel import helpLabel
 from manuskript.ui.mainWindow import Ui_MainWindow
 from manuskript.ui.tools.frequencyAnalyzer import frequencyAnalyzer
@@ -129,6 +130,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         self.actCopy.triggered.connect(self.documentsCopy)
         self.actCut.triggered.connect(self.documentsCut)
         self.actPaste.triggered.connect(self.documentsPaste)
+        self.actFind.triggered.connect(self.doFind)
         self.actRename.triggered.connect(self.documentsRename)
         self.actDuplicate.triggered.connect(self.documentsDuplicate)
         self.actDelete.triggered.connect(self.documentsDelete)
@@ -247,6 +249,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         for i in [self.actCut,
                   self.actCopy,
                   self.actPaste,
+                  self.actFind,
                   self.actDelete,
                   self.actRename]:
             i.setEnabled(tabIsEditor)
@@ -1620,3 +1623,8 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         self.dialog = exporterDialog(mw=self)
         self.dialog.show()
         self.centerChildWindow(self.dialog)
+        
+    def doFind(self):
+        if self._lastFocus and self._lastFocus == self.mainEditor:
+            f = FindDialog(self, ew=self.mainEditor)
+            f.exec_()
--- a/manuskript/ui/editors/editorWidget.py
+++ b/manuskript/ui/editors/editorWidget.py
@@ -424,6 +424,8 @@ class editorWidget(QWidget, Ui_editorWidget_ui):
         if self.getCurrentItemView(): self.getCurrentItemView().paste()
     def rename(self):
         if self.getCurrentItemView(): self.getCurrentItemView().rename()
+    def find(self, text, options):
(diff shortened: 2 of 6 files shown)


worstje:
First of all, thank you for contributing! It is great to see another PR! 😄 Especially for this issue which had me tempted to do it on more than one occasion, but the GUI stuff tends to scare me away. 🤣 

I had a brief look over your PR, and it appears functional. But I wonder if it is functional enough right now for inclusion...? Let's test. Since searching is a rather basic thing, I will use Windows Notepad as the basic standard I consider to be 'fit for purpose'.

I can search for something, and it is selected. So far, so good.

Let's search again. Habitually, I hit **F3**... whoops, not supported.

Okay, I'll hit **Ctrl+F** again. The textbox is empty. Awkward. I'd expect it to be either filled with my previous search, or with whatever I have selected in the editor right now.

No biggie; I'll just re-enter my search term... and it is stuck on the exact same item.

To get it to find the next item, I have to move my cursor to be after the first letter or whatever got selected. (The left boundary of a selection is typically considered to be the cursor.)

Additionally, this does not cover some other basic features I see in Windows Notepad:

* It always searches downwards. I cannot make it search in an upward direction. (Useful for finding the last few mention of a character or place, for example.)

* It ignores case sensitivity. Sometimes case sensitivity is useful, like when searching for names. It is a simple way to avoid partial matches in the middle of words. (Example: Matt and matte).

* No wrap around functionality where it will keep searching from the other end of the document. (Useful when you are at the bottom of a document and forget to start searching from the very top.)

Of course, such extra features would probably exceed the capability of `QInputDialog`, meaning more UI work. Either way, as a whole, I believe this PR still has a little ways to go before it is ready. (Admittedly, a very basic feature is still better than no feature...)

I hope you won't consider this little review discouraging; I am very glad to see someone tackling a search function! 👍 

(Note: I am just an ordinary contributor like you and @gedakc may very well think differently than I do!)

kbeecher:
Hi. Thanks for the reply. I meant to leave a message after creating the PR, but I got distracted.

I intended to say that, as you say too, it's not ready for merging. I created the feature for myself because I really missed a Find function and it was enough for me, then later thought I'd share it to see if there was any interest. I can write Python, but I'm not a Qt developer, so I had little idea of what I was doing.

If there is interest, I'll gladly clean it up. If anyone could give me a few implementation hints, I'd be grateful. For example: Which Qt widget is the best for the job (QLineInput, a custom QDialog)? If I should create a custom class, where should it go?

Thanks for any hints. And thanks for Manuskript, I really enjoy using it!

worstje:
I'd definitely love to see someone fully embrace the implementation of a Find feature. (And a replace one as well, but I feel bad enough about widening the scope as it is! 😇)

I am not the right person to point you are the best places since I have mostly been on bugfix sprees and quality-of-coding-life enhancements. That said, I think the manuskript/ui folder is the right spot; I see some other windows and controls in there already.

Just now I looked around a bit for pre-made solutions. The most obvious idea was [QtFindReplaceDialog](http://qtfindreplace.sourceforge.net/), but unfortunately that is not maintained nor applicable for PyQt. More promising is [this article](https://www.binpress.com/building-text-editor-pyqt-3/) which explores various aspects of making an editor, among which is find/replace functionality.

Mind you, I do not quite know the license that code comes in, and it would also need quite a bit of editing to fit in with our codebase, so I would use it as a reference at most. (No copy-pasting even though it is a tutorial.) Admittedly, some functionality can only be implemented in so many ways, so it is unavoidable your code will resemble theirs in some spots.

kbeecher:
OK, I redid the approach. It's still very basic, but I took inspiration from the GNOME text editor which is just a textbox and FindUp and FindDown buttons (see screenshot).

gedit

It still ignores case and doesn't wrap, but I think it's an improvement.

worstje:
I haven't gotten around to looking at your PR again, but I just spotted a newly opened issue:

[Any plans to support regular expression searches? (#682)](https://github.com/olivierkes/manuskript/issues/682)

Since someone asked about it, there is definitely interest in having the feature. (I know I myself would definitely enjoy having the feature available.)

Hopefully I'll get around to checking out your work before this week is over. 😄

worstje:
It is kind of at the very last moment of the week, but I finally found an opportunity to give your changes a shot. (My test case was to solve a very common typo I seem to keep making while writing, so I wanted to quickly check a few dozen chapters.)

I came across the following issues:

- There is no feedback to the user when it doesn't find anything.

 - Search defaults to 'backward' search as that is the button that is bound to the text box; I have to f3-type-tab-enter if I want to search in a forward direction.

- The window always pops up in the middle of the screen. Not necessarily a bad thing, but with its general workflow I found it got in the way of the text I was looking at.

- Clicking a text on the left side of the Editor would break the Find command; I had to manually click in the text area before it would start working. This is probably something unrelated to your changes and something we'll have to look into at a later point, but your current implementation does draw a lot of attention to it.

As it is, I do not think it is user-friendly enough to merge quite yet.

Searching is in my experience a feature that works the same everywhere and has to behave seamlessly according to expectations for it to be decent. Right now, I feel the core functionality is there, but it is hidden really awkwardly behind a lot of stumbling blocks that make it unintuitive to use.

My personal recommendation would be to look more closely at existing software. Things I believe are worth keeping in mind while doing so are:

1) Searching in a web browser of PDF is not the same as searching in a text editor. Similarly, a word processor might also breed specific expectations under its users. I would personally focus on text editors: Manuskript users tend to be at a higher level than grandma who just types a letter in all capitals. (One example of this is the issue I mentioned above which has a user requests for search functionality that incorporates regular expressions.)

2) Workflows are key. Do you keep a window open on the side, or is initiating a search a one-time operation that is repeated with another shortcut like F3? If you invoke the feature with your keyboard, how many keypresses does it take? How convenient and fool-proof are common operations?

3) On one side of the proverbial coin there is the 'Find' functionality, and on the flip side is the scary boogeyman known as the 'replace' functionality. These two share a lot of functionality: what direction to search in, case sensitive, etc. Ideally the user experience (and by proxy the user interface) for the two are very similar, or perhaps the same screen.

I think I had some other things I wanted to touch upon, but I am tired and lost my line of thought. 😁

Please don't take my comments as anything negative; you are making good progress digging into the functionality and creating something that will in the end become a gorgeous addition to Manuskript.

And while I might sound as if I want to see a top of the line limousine of searching appear, I would definitely be happy with a functional basic system. My concern is that the basics need to be solid, especially for the users. If the users get used to a certain set of keypresses or behaviours, any future changes to the feature can mess with habits that have since been established. I'd rather frustrate our users a bit less by thinking a bit more on the initial design. 😄

TheJackiMonster:
I merged the new global search from #717 already which solves a similar problem. The search is global though instead of local to the open file. Currently I would think we're fine with the global search working but I will look into this for upcoming releases.

PR #690, opened 2019-11-13
added 3 buttons to the textEditView that allow quickly adding new items

nagolinc:
Thought this would be handy so you can quickly create characters/plots/world items from the editor.

1) Select text of name of the object you want to create in the textEdit view
2) Right click
3) Click on the type of object you want to create

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -378,8 +378,12 @@ class MainWindow(QMainWindow, Ui_MainWindow):
         else:
             self.chkPersoPOV.setCheckState(Qt.Unchecked)
 
-        self.chkPersoPOV.stateChanged.connect(self.lstCharacters.changeCharacterPOVState, F.AUC)
-        self.chkPersoPOV.setEnabled(len(self.mdlOutline.findItemsByPOV(ID)) == 0)
+        try:
+            self.chkPersoPOV.stateChanged.connect(self.lstCharacters.changeCharacterPOVState, F.AUC)
+            self.chkPersoPOV.setEnabled(len(self.mdlOutline.findItemsByPOV(ID)) == 0)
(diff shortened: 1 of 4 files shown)


TheJackiMonster:
This merge request needs to be rebased to the current state or at least the latest release. The models could potentially be different.

Also when I try to add a character, the whole applications crashes currently on my system without specifying the exact issue.

> Fatal Python error: Aborted
>
>Current thread 0x00007f35a80cd740 (most recent call first):
>  File "/home/thejackimonster/github/manuskript/bin/../manuskript/main.py", line 221 in launch
>  File "/home/thejackimonster/github/manuskript/bin/../manuskript/main.py", line 234 in run
>  File "/home/thejackimonster/github/manuskript/bin/manuskript", line 13 in <module>

nagolinc:
Thanks for looking at this! I think the bug should be fixed now!

TheJackiMonster:
I assume the problem wasn't the missing value because you have added default values already and it's still crashing. ^^'

TheJackiMonster:
Ok, the issue seems to be fixed. Now I have one thing which would be great to work properly:
 - When I right-click on a word to automatically select the word and open the context menu, your actions don't appear.

I think if this would work, the commits only need to be squashed to one then I could merge it.

TheJackiMonster:
Ok, I have fixed some problems related to the automatic selection of words. If you squash the commits I can just merge it. The problem I had should then also be solved.

nagolinc:
I think I did the squash right?  My git-foo is not the best.

TheJackiMonster:
Yes, it worked now. I have adjusted the context menu a bit to make it work consistently with spell checking together. I think I will abstract the context menu in a different file if the functionality grows even more. ^^'

PR #701, opened 2019-11-24
More german translations

argail1980:
While at it, translated some more. Unsure, if 
`<location filename="../manuskript/mainWindow.py" line="1600"/>
        <source>Proceeding might crash and lose data</source>
`
is translated correctly. Meaning of my translation:
"If you proceed, the software could crash and you could lose data"


gedakc:
@argail1980 does the new wording in PR #715 address your concerns?

gedakc:
Merging these translations updates for inclusion in the next release of Manuskript.

PR #704, opened 2019-12-04
Fixed bugs caused by parallel access during multithreading

TheJackiMonster:
Now the 'textEditView' should synchronize model and text thread safe without problems. The QMutex solves the causes of parallel access. I guess most people haven't got any problems because it is so rare. If I find similar pieces of code I will probably add Mutexes too.

Keep up the good work on this project!

Also here is some information I looked up to get the implementation right:
https://doc.qt.io/qtforpython/PySide2/QtCore/QMutex.html

### DIFF
--- a/manuskript/ui/views/lineEditView.py
+++ b/manuskript/ui/views/lineEditView.py
@@ -50,39 +50,39 @@ class lineEditView(QLineEdit):
         self.updateText()
 
     def submit(self):
+        self._updating.lock()
+        text = self.text()
+        self._updating.unlock()
+
         if self._index:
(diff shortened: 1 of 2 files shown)


gedakc:
Thanks @TheJackiMonster for contributing a bug fix.

There appear to be several commits in this merge request.  Is this a work-in-progress or is this patch complete?

If it is complete then it would help us if you rebase this merge request on the current **develop** branch, and also squash all the commits into one.

Then when we have time we can review this MR.

TheJackiMonster:
Yes, it should be complete. I will do that as soon as possible. ^^

Issue #688, opened 2019-11-12
References Mysteriously Delete in Summary and Reference/Notes Fields


opensourcefiction:
I've really been enjoying how Manuskript has matured over the last few years. I recently imported a project from ZimWiki and have really appreciated the fiction-specific  features Manuskript offers over Zim. It's also much easier to tag scenes with metadata than in Bibisco. 
Right now I am using Manuskript 0.10.0 on Linux Mint 18.2.

An issue I've been encountering is when I attempt to tag all my scenes with relevant metadata using references. I started making a list of characters in each scene, important objects, and plot threads in the references tab. Sometimes it went great, other times I would have completed the entire list, attempt to make a new reference, or click in the field, and everything would be deleted (including text that was not a reference, such as a note). As a work around I tried doing the same in the Summary field, with similar results. I have yet to be able to find the consistency in the bug (as in, which actions will definitely lead to the mysterious deletion).

Does anyone have any insight into this issue? Thanks in advance!

worstje:
No insights other than that I noticed the same thing happening yesterday. I haven't used the notes / references field in a while, so I don't know if it is something new or if it goes back a little ways.

All I know is that sometimes things would stick around, and sometimes it would be in a hurry to poof again. Really annoying.

obw:
Could it be the same Error as in my Issue? #670

worstje:
@obw It could be, good point! I hadn't associated it with your report because I thought you were talking about the 'Summary' field as opposed to the 'Notes / References' field.

There's definitely some searching that needs to be done on this issue.

Cuillerechan:
It seems to happen precisely when the "insert reference" is activated, which is how I find myself with something like 
![image](https://user-images.githubusercontent.com/4191253/68965061-33696980-07db-11ea-904d-dd7c19623b8c.png) : All previous text has just disappeared.

**From what I can tell**, it does seem to happen with all types of references, but only when using the ctrl+SPACE shortcut (versus the use of the context menu) and only when there is already text written (that it will use to look up the reference).

I also noticed that even though the ctrl+z shortcut usually works in there, when the bug happens, the shortcut is unable to go back and undo the accidental delete.

Hope this might help :)

worstje:
I have had the deletion happen without relying on the popup or Ctrl+V. All I did was paste a few links which would mysteriously disappear. So I am not sure it is the Ctrl+Space or popup itself that is at fault.

Cuillerechan:
[...]

Issue #672, opened 2019-10-25
After use of ctrl-z or ctrl-v Manuskript goes in inactive State


obw:
Goes back active, when Text is changed.

![manuskript_after_use_of_ctrl_v](https://user-images.githubusercontent.com/3050806/67582771-a5094700-f74a-11e9-8582-0b94f40ced80.png)

Forgot to mention:
Latest Dev-Branch, I think cloned yesterday,
Linux under KDE

Verwendete Software-Versionen:
     Python 3.6.8
     PyQt 5.13.1
     Qt 5.13.0

worstje:
Can you clarify the nature of the problem more?

Perhaps it is obvious to a KDE user, but just by looking at the redacted screenshot I don't immediately see what might be wrong.

Does the text box lose focus? Or does the entire window lose focus? Or do you mean something else entirely when you talk about an 'inactive' state?

obw:
The Problem, that you must change the Text in the Scene, to enable the Selectboxes again in '**Einstellungen**', sorry I actual don't know the English label, it should be '**Configuration**'

A Picture say more the thousand words:)

![manuskript_after_use_of_ctrl_v_what_it_should](https://user-images.githubusercontent.com/3050806/67620567-9ed9a000-f808-11e9-8f7a-05afd02a4723.png)

obw:
#706 - this Patch solve this Problem!

If merged with develop and master Branch, this Bug can be closed!

gedakc:
Closing as fixed by PR #706.

Issue #670, opened 2019-10-25
Cut and Paste - Comment Field - First shows, then disappiers


obw:
I'm trying to get a Overview for a little bigger project of mine. A Part of it, is to add Locations, Outfits, Plot and Persons to the comment of every Chapter. To get a quicker grasp of it. I have this information mostly in the Scenes, where something change and so on.
So I tried to copy the comment of the scene, to the comment of the Chapter and Manuscript loose it:(.

After that I tried to copy something in a Scene comment, there it also got lost.

If there was some text, this Text will go also away, not found again...:(

Forgot to mention:
Latest Dev-Branch, I think cloned yesterday,
Linux under KDE

Verwendete Software-Versionen:
     Python 3.6.8
     PyQt 5.13.1
     Qt 5.13.0

obw:
ALso in the Content Overvier or so, I use German Locale, there the Name is '**Inhaltsangabe**'

obw:
#706 - this Patch solve this Problem!

If merged with develop and master Branch, this Bug can be closed!

gedakc:
Closing as fixed by PR #706.

Issue #81, opened 2017-01-13
(l)ubuntu dependencies for develop branch


trivers:
Hi,
The dependencies I needed for lubuntu 16.04 (32 and 64bit) where a little different than those listed in the read me, for using the develop branch.

I also needed to have 'python3-pyqt5.qtwebkit'

This is what gets it running:
`sudo apt install python3-pyqt5 libqt5svg5 python3-lxml python3-enchant zlib1g python3-pyqt5.qtwebkit`

It's possible this is also what is needed for other *buntu distros, but I have not tested that;)

Renkineko:
Not needed on my ubuntu mate 16.04. Maybe it could help on my xubuntu, i'll test it later.

yarko:
on my 16.04, this (nor other things) have worked.

Also, looking at `makefile`, I not there is no manuskript/setup.py

python-3.6;
python  PyQt5 doesn't have interface for PyQt5.QtWebKitWidgets:

```
~/workspace/writing/manuskript$ bin/manuskript 
Warning: failed to load translator for locale en_US...
Traceback (most recent call last):
  File "bin/manuskript", line 13, in <module>
    main.run()
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/main.py", line 63, in run
    launch()
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/main.py", line 67, in launch
    from .mainWindow import MainWindow
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/mainWindow.py", line 22, in <module>
    from manuskript.ui.exporters.exporter import exporterDialog
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/ui/exporters/exporter.py", line 10, in <module>
    from manuskript import exporter
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/exporter/__init__.py", line 4, in <module>
    from manuskript.exporter.manuskript import manuskriptExporter
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/exporter/manuskript/__init__.py", line 6, in <module>
    from manuskript.exporter.manuskript.HTML import HTML
  File "/home/yarko/workspace/writing/manuskript/bin/../manuskript/exporter/manuskript/HTML.py", line 4, in <module>
    from PyQt5.QtWebKitWidgets import QWebView
ModuleNotFoundError: No module named 'PyQt5.QtWebKitWidgets'
```

[...]

PR #706, opened 2019-12-13
Fixed bugs caused by parallel access during multithreading

TheJackiMonster:
Now the 'textEditView' should synchronize model and text thread safe without problems. The QMutex solves the causes of parallel access. I guess most people haven't got any problems because it is so rare. If I find similar pieces of code I will probably add Mutexes too.

Keep up the good work on this project!

Also here is some information I looked up to get the implementation right:
https://doc.qt.io/qtforpython/PySide2/QtCore/QMutex.html

Squashed version from here:
https://github.com/olivierkes/manuskript/pull/704

### DIFF
--- a/manuskript/ui/views/lineEditView.py
+++ b/manuskript/ui/views/lineEditView.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # --!-- coding: utf8 --!--
+from PyQt5.QtCore import QMutex
 from PyQt5.QtWidgets import QLineEdit
 
 from manuskript.enums import Outline
@@ -13,7 +14,7 @@ class lineEditView(QLineEdit):
         self._indexes = None
         self._index = None
         self._placeholderText = None
-        self._updating = False
+        self._updating = QMutex()
 
     def setModel(self, model):
         self._model = model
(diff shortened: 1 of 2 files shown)


obw:
Not able to start this Version:(

import-im6.q16: not authorized `os' @ error/constitute.c/WriteImage/1037.
import-im6.q16: not authorized `sys' @ error/constitute.c/WriteImage/1037.
./manuskript: 7: ./manuskript: Syntax error: "(" unexpected

I have tried multiple things, including copying only the changed Files to a working copy of V0.10.0dev

I'm on a Linux mashine(4.15.0-72-generic #81-Ubuntu), under KDE

 - Python 3.6.9
 - PyQt 5.13.1
 - Qt 5.13.0

TheJackiMonster:
Okay, here is one of my setups which allows me to start and use it:
I'm on Archlinux (linux-5.4.6.arch3-1), under GNOME (using wayland)
- python-3.8.1-1
- python-pyqt5-5.14.0-3
- qt5-base-5.14.0-1

Your error seems related to Python (the error-message does not refer to the changed files from this commit). Can you start the release-version of manuskript? "not authorized" could refer to permissions too..

obw:
> Your error seems related to Python (the error-message does not refer to the changed files from this commit). Can you start the release-version of manuskript? "not authorized" could refer to permissions too..

I work nearly daily with Manuskript;)

My last attempt to start this version,was make a copy of my V0.10.0dev, start this copy

**no Problem**

Then copy the changed files in this copy...

**get the Error from above**

This is strange for me, perhaps it because I'm running 3.6 of python?
Python is for me as Programmer something completely new, something similar in PHP would no big issue for me:)

TheJackiMonster:
I can't reproduce the error you get. The closest result was the message I got when I commented out the shebang [this post on stackoverflow](https://stackoverflow.com/questions/55558605/import-im6-q16-not-authorized-error-os-error-constitue-c-writeimage-1037-fo/55570825) is referring to.

So maybe try executing the script at "bin/manuskript" with python3 directly ( `python3 bin/manuskript` ). I also checked executing it with python2 but it fails differently.

obw:
@TheJackiMonster : thanks for your advise, this is working;)

Also:
- #670 is solved by this patch
- #672 is solved by this patch
- #688 also solved

Perhaps I have forgotten a Issue;)

gedakc:
Thanks @TheJackiMonster for updating this patch, and thanks @obw for testing the patch and also linking to other issues that are also fixed by the patch.  It is a huge help to have others like yourselves contributing and helping others in the forum.  :-)

I'm thinking of creating a bug-fix release and plan to include this patch.  Before that I have to fix the Travis Continuous Integration as it appears to broken again due to [Apple and HomeBrew dropping support for older versions of Mac OS X, such as 10.12](https://travis-ci.org/olivierkes/manuskript/jobs/624768797).

gedakc:
I've done what I can in an attempt to get the Travis CI automated builds working for macOS X.  Because I do not wish to delay this PR any longer I will merge it now.

PR #668, opened 2019-10-20
Spelling: crash and dataloss

comradekingu:

PR #715, opened 2020-01-09
Change wording of import warning for PyQt/Qt versions 5.11 and 5.12

gedakc:
Two separate pull requests indicate an issue translating the warning
for when an import is attempted with PyQt/Qt versions 5.11 and 5.12.
As such change the warning message.

See PRs #668 and #701.

Following is a screen shot of the warning message when run on Ubuntu 18.10.

Manuskript-Import-Warning-Dialog

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -1596,7 +1596,7 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             print("WARNING:", warning1, warning2)
 
             msg = QMessageBox(QMessageBox.Warning,
-                self.tr("Proceeding might crash and lose data"),
+                self.tr("Proceed with import at your own risk"),
                 "<p><b>" +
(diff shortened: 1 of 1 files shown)


comradekingu:
@gedakc
Is it both versions of both causing issue, or respectively? Or maybe the combination?
PyQt 5.11 used with Qt 5.12 is known to …

gedakc:
> Is it both versions of both causing issue, or respectively? Or maybe the combination?

From our testing any combination of Qt and PyQt with versions 5.11.x and 5.12.x caused Manuskript to crash when choosing **File -> Import**.

Normally the PyQt versions mirror the version number of Qt, but not always.

> Edit: "a loss of data" is redundant. Could do without the "a".

'Agreed.  However I think I'll leave it in as it may already be translated in other languages and I'd prefer not to cause extra work for such a minor change.

Issue #709, opened 2019-12-23
All types of export with pandoc fail with error 97


wrtrtstr:
No matter what format I choose, the message from stderr says:
pandoc: Could not find data file /usr/share/pandoc/data/templates/default.*

The files, however, are there:
$ ls -l /usr/share/pandoc/data/templates/
total 144
-rw-r--r-- 1 root root  429 Jan 26  2018 default.asciidoc
-rw-r--r-- 1 root root 7483 Jan 26  2018 default.beamer
-rw-r--r-- 1 root root  218 Jan 26  2018 default.commonmark
-rw-r--r-- 1 root root 3708 Jan 26  2018 default.context
-rw-r--r-- 1 root root  719 Jan 26  2018 default.docbook
-rw-r--r-- 1 root root  569 Jan 26  2018 default.docbook5
-rw-r--r-- 1 root root  130 Jan 26  2018 default.dokuwiki
-rw-r--r-- 1 root root 4307 Jan 26  2018 default.dzslides
-rw-r--r-- 1 root root 1332 Jan 26  2018 default.epub
-rw-r--r-- 1 root root 1329 Jan 26  2018 default.epub3
-rw-r--r-- 1 root root    7 Jan 26  2018 default.haddock
-rw-r--r-- 1 root root 1581 Jan 26  2018 default.html
-rw-r--r-- 1 root root 1502 Jan 26  2018 default.html5
-rw-r--r-- 1 root root 1851 Jan 26  2018 default.icml
-rw-r--r-- 1 root root 7548 Jan 26  2018 default.latex
-rw-r--r-- 1 root root  512 Jan 26  2018 default.man
-rw-r--r-- 1 root root  218 Jan 26  2018 default.markdown
-rw-r--r-- 1 root root  130 Jan 26  2018 default.mediawiki
-rw-r--r-- 1 root root 2097 Jan 26  2018 default.opendocument
-rw-r--r-- 1 root root  244 Jan 26  2018 default.opml
-rw-r--r-- 1 root root  295 Jan 26  2018 default.org
-rw-r--r-- 1 root root  218 Jan 26  2018 default.plain
-rw-r--r-- 1 root root 7365 Jan 26  2018 default.revealjs
-rw-r--r-- 1 root root  501 Jan 26  2018 default.rst
-rw-r--r-- 1 root root  586 Jan 26  2018 default.rtf
-rw-r--r-- 1 root root 2541 Jan 26  2018 default.s5
[...]

PR #737, opened 2020-02-11
Add snap layout for pandoc templates directory

tomwardill:
pandoc is installed via apt into the snap, but we need the files that are installed to be available in the context that the snap is running in.
This uses the snap layouts feature to map the installed directory into the context filesystem.

Fixes https://github.com/olivierkes/manuskript/issues/709


gedakc:
Merging this PR as it is required by @tomwardill for the snap package.

PR #744, opened 2020-03-13
Travis CI Macintosh build using Xcode 11.3

sijpkes:
This build works on my machine OS X Catalina using:

```
nohup bin/manuskript
```


sijpkes:
I'm unable to devote the time to this but I hope the existing edits will suffice.

Issue #725, opened 2020-01-27
[Feature Request] Alternate character counter


Nasgarath:
I write in both French and English. While the English market works with word counts, the French market is entirely focused on characters (spaces included).

It's not vital, because the word count would still give me a good sense of progress and once the piece is done I can open it in Word and get the character count from there... But it'd definitely be a nice and very convenient feature ;)

obw:
This proposal exist with #607 Chinese/Korean and Japanese also need Charcount...

So this is a dupe...

Issue #607, opened 2019-07-17
Word count statistics incorrect when using Chinese characters


finitecow:
wold statistics incorrect when using Chinese character

worstje:
I do not speak/read Chinese, so I am unable to verify the exact nature of how it is wrong or what needs to be done to fix it.

Can you supply examples of Chinese text, what the statistics report and what the statistics ought to read? If you can offer up a sample Manuskript project that exhibits the problem, that would be ideal.

obw:
Chinese is a Language, where normally not Word- but Charcount ist used, like Japanese. A Char is often a complete Word!

Counting Chinese or Japanese Words is a art-form.

One Symbol can be a Word, but also it can be part of a Word. Also there are normally no white-spaces between Words. Simples and best Solution is a Switch to count all not white-space Chars. This is, be my knowledge what is used in these Countries!

gedakc:
Thanks @obw for the explanation that some Chinese characters are a complete word, and that may make counting Chinese words challenging.  This is something that likely requires a person with Chinese knowledge to address correctly.

obw:
Sorry English is not my first Language.

All Chinese Chars are one Word, but sometimes there are compound words... so two or more Words are combined to one.

Japanese is the real evil here:
4 Alphabets:
Hiragan, Katakana: Sybil chars, oh and the Sybils of both Alphabets are the same, but the Layout is different, one is used for loanwords and technical terms, the other for lyrical and if you don't know the Kanji. (Fun Fact the Term Lightnovel in Japan means, a book without seldom used Kanyis, so you don't need a dictionary too read it)
Kanji: see Chinese, they are literal Chinese Chars
Latin Alphabet, including Arabic Numbers... Why easy, when you can go too (insert the place you don't like here ;) )

I hope this is more correct an better for understanding...

I also repeat, I had look it up. All payments for Authors, Translators and so on are calculated over the char count (without white spaces).

TheJackiMonster:
I added a PR (https://github.com/olivierkes/manuskript/pull/746) for a char-counter which can be enabled/disabled in settings. Should I add a setting to ignore white spaces or shouldn't it count them anyway? I could still add something like this before it gets merged, I guess. ^^

taophp:
@TheJackiMonster I think a option for ignoring/including spaces is required as in France, for example, some publishers count in a way and some in the other.
Thanks for the job !

TheJackiMonster:
[...]

PR #746, opened 2020-03-27
Added char-count with settings to enable/disable it.

### LINKED ISSUE #334: Show character count progress indicator

TheJackiMonster:
I have seen an old PR (https://github.com/olivierkes/manuskript/pull/339) which should fix that issue and I adapted the changes from it. So this request will add a char-count next to the word-count or only via tooltip (default) depending on the general settings. There are also added settings to show the amount of chars in the tree-view.

This PR fixes #334 but it does count chars with whitespace. It is mentioned in other issues that whitespace should not be counted, so these will probably stay open until a setting for it is added.


TheJackiMonster:
Sorry btw for the commit-arrangement in this. First I tried to make it only one commit but then I thought it would be better to include/merge the old commits from the original PR somehow. Still works as intended by me.

TheJackiMonster:
This should fully solve #725 since the whitespace gets counted.

TheJackiMonster:
I've added a setting for disabling counting the space characters. So it should solve the problem mentioned in #607. But because it would require to recount characters for all files and folders after changing the setting which is quite complicated to implement properly. It will require a restart of the application to work properly and update the amounts. But this shouldn't be a bigger deal for now.

Issue #142, opened 2017-08-30
add languagetool


obw:
I have use LibreOffice with the languagetool Plugin for writing the last Years!

It's a great Help to make the writing better

TheJackiMonster:
As mentioned in #745, there are already some good looking python bindings and there is an open http-api. So it should be able to integrate it even without difficult setup. ^^

TheJackiMonster:
![Peek 2020-03-29 04-51](https://user-images.githubusercontent.com/13867397/77838959-fe9df380-7178-11ea-8bc8-e1fee5d01c00.gif)
Okay, I got this for basic grammar-correction and some stylistic information which can not be corrected via menu but manually. ^^'

I managed to implement it without to much messing around, I think. It checks if the module for Python ("language_check") is installed and uses it as new implementation of dictionaries. The other dictionaries had to get some new methods to make sure everything works without errors and such since LanguageTool is way more advanced than Enchant and others.

At the moment I have added three different error-visuals (underlining-colors) for "misspelling", "grammar / typographical" and "other". There is still much work to do, so that errors are all marked correct and this tool is still pretty slow. One thing I found out during testing is that the method to highlight the text is called more than once for the same text. So there is still much which can be optimited... ^^'

PR #747, opened 2020-03-29
Added basic support for LanguageTool via 'language_check' as advanced spellchecker

TheJackiMonster:
This should fix Issue #142. At the moment it is only very basic but if "language_check" is installed, it can be selected as dictionary which enables additional to the spellchecking some grammar- and style-checking.

Some additional mistakes in grammar and style can be solved via context-menu like the misspelling but some will only show some information. Others are just underlined/marked currently.

I would like to try to improve the performance before it gets merged since it has way more noticeable loading times at start of the program and during chapter/scene changing. ^^'

### DIFF
--- a/manuskript/functions/spellchecker.py
+++ b/manuskript/functions/spellchecker.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # --!-- coding: utf8 --!--
 
-import os, gzip, json, glob
+import os, gzip, json, glob, re
 from PyQt5.QtCore import QLocale
 from collections import OrderedDict
 from manuskript.functions import writablePath
@@ -27,6 +27,11 @@ try:
 except ImportError:
     symspellpy = None
 
+
+try:
+    import language_check as languagetool
+except:
+    languagetool = None
 
 class Spellchecker:
     dictionaries = {}
@@ -117,6 +122,17 @@ class Spellchecker:
             pass
         return None
 
+class BasicMatch:
+    def __init__(self, startIndex, endIndex):
+        self.start = startIndex
+        self.end = endIndex
+        self.locqualityissuetype = 'misspelling'
+        self.replacements = []
+        self.msg = ''
+
+    def getWord(self, text):
+        return text[self.start:self.end]
+
 class BasicDictionary:
     def __init__(self, name):
         self._lang = name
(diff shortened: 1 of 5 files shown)


obw:
When I understood this correctly, this use the public avail webservice.

It should be configurable to a local install of langtool for advance users.

Why:

1. It's faster
1. No Internet is needed while checking
1. Privacy

Needed fields in the configuration basic version and external start of the langtool:

- Local / public
- when local: Port of the Service

Better version, in which Manuskript started langtool :
- Local / public
- when local: Port of the Service
- Path to the langtool jar
- Path to java

Also thanks for your effort, this is a real enhancement for me;)

TheJackiMonster:
I thought so too but on Archlinux at least it installs a version of LanguageTool with language_check locally and opens a java-process running it.

I will try to test it on Windows too, so that it won't be too complicated to setup. ^^'

TheJackiMonster:
Without changing anything in the source I got a massive performance boost just by updating to the newest release of language_check (1.1 currently). It's still slower than the other dictionaries but I think it's usable if you don't swap scenes every second.
I should add that the java process which runs LanguageTool needs some memory (as expected from Java). It is at around 2.2 GB on my machine during testing and I don't know if it increases over longer usage. I wouldn't recommend using this on a small netbook without the memory capacity. ^^'

obw:
@TheJackiMonster this was something to be expected...

On both of my machines not Problem (8 or 16GB), but performance is a problem, when it's make Manuskript overall slower, that is definitely a Problem...

One of my bigger Projects (Around 500k Words), the startup of Manuskript alone is evil, two Minutes upward, also when I go to some of the older Parts, I could make Tea in the time of switching:(...

This is something we must work on soon...

TheJackiMonster:
@obw Wow, my biggest project is only about 33k words.. So I guess we will only know how slow it is, when you try the final merge. ^^'
I think performance could be optimized for the whole project in a separate PR. So this will only focus on the actual feature for now.

obw:
@TheJackiMonster I think, you are right, with a own Performance Ticket...

But I mention it, so that we all aware of it...

TheJackiMonster:
Okay, I implemented some sort of cache for the checking with LanguageTool. So once the text was opened it can reopened faster because as long as nothing gets changed, the tool does not have to check the text again. The results are saved in memory until you close the whole application. But I think it's fine since the process of the tool alone needs much memory and it does only stores a list of suggestions for errors and others. ^^

TheJackiMonster:
Okay, I think everything is ready for review and merge now. I have added tooltips which present the messages of the suggestions for correction. Every match has a different color based on its type and the selection which was always selecting whole words picks the correct part of the text if nothing was selected on purpose. The colors aren't perfect at the moment but most matches have the type 'uncategorized' anyway. Settings for customization would probably need another, so I will leave this open for now.

TheJackiMonster:
Okay, I added a commit to prevent Manuskript from crashing when selecting LanguageCheck as dictionary-library and Java is not installed. I basically disabled it as option as long as Java isn't installed properly. Java is definitive required to run LanguageTool locally.
The other thing I had to fix was: Some tooltips I added for errors and such were shown on hovering over words even when spellcheck was disabled in the menu. So I prevented that...

Issue #335, opened 2018-04-04
Enable/Disable POV-Option for Characters


ScullyBlue:
In my actual project, I see, that the list of characters can get ... long. But not all characters get a own POV in story. It would be very nice, if I could disable or enable for each character that he stands in the List of POV-Characters or not.

TheJackiMonster:
I think this shouldn't be too complicated, so I will try to add this next. I have a higher amount of characters too but the most of them get an own POV. But maybe in the future I can make use of it. ^^

TheJackiMonster:
Okay, I looked into the code and this request is probably more difficult than I thought to implement. It would require a separate model of all enabled characters next to the model of all characters and they have to be synchronized too. I think it's better to improve searching in such sort of selections for specific characters rather than filtering them. ^^'
So for the moment it looks like we have to deal with all characters for now. Maybe some alphabetic ordering (#321) would already help in the selection too.

obw:
@TheJackiMonster A searchable POV would be a Dream, My biggest project where I use Manuskript has about 100 Characters, still growing;).

When I ever done with this monster of an Idea there will be 10.000+ Characters.

TheJackiMonster:
I didn't implement a full search now but I got to fix the original issue as mentioned above. So a POV can be disabled for a character individually. It was easier as I thought yesterday because Qt had some classes and functions already in mind for such a filtering task. ^^

TheJackiMonster:
@obw It's probably good to open the searchable POV as new issue. So I would close this one for now.

PR #748, opened 2020-03-31
Enabling/Disabling POV for a specific character

TheJackiMonster:
I added a property to the characters to disable them for getting listed in selecting a POV. It can only be disabled if there aren't any elements in the plot which use the characters' POV. It uses a custom QSortFilterProxyModel to filter the actual model of characters.
This should fix the issue #335 completely.


obw:
Good Idea, but for me it's more work, because everyone can become POV...

I will see I test this Version, a searchable Selectbox, would be my dream solution for this...

But also, thanks for your efforts...

TheJackiMonster:
@obw There are some other PRs at the moment with a global search and search & replace. So maybe some of their work can be adapted to implement something like this without creating too much duplicated code.
But yes, you are right this current state doesn't help you much with many POVs. I have a story with around 40 different POVs too. ^^'

PR #788, opened 2020-07-22
Don't arbitrarily delete extra files in project folder

Blake-Eryx:
Currently, the program deletes any files that it doesn't recognize in your project folder. Perhaps useful for deleting "orphan" files, but if you store things like custom pandoc YAML files and various peripheral files in that folder, then it seems rather destructive to always be arbitrarily deleting files like these.

### DIFF
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -391,14 +391,6 @@ def saveProject(zip=None):
 
         # Removing phantoms
         for path in [p for p in cache if p not in [p for p, c in files]]:
-            filename = os.path.join(dir, folder, path)
-            log("* Removing", path)
-
-            if os.path.isdir(filename):
-                shutil.rmtree(filename)
-
-            else:  # elif os.path.exists(filename)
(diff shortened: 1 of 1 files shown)


manuskript/load_save/version_1.py
```diff
@@ -663,6 +655,9 @@ def loadProject(project, zip=None):
                 if f[-4:] in [".xml", "opml"]:
                     with open(os.path.join(dirpath, f), "rb") as fo:
                         files[os.path.join(p, f)] = fo.read()
+                elif f[-4:] in [".png", "jpeg",".jpg","bmp","tiff",".gif",".pdf"]:
```
obw, line 658: This should by configurable, I use often SVG images.

Also HTML Files often with research information or other filetypes. (bst -> bibtex and so on)


In the moment I store my Images in folder on the same level as the project-file (I do not us the zip/one-file mode!)... So I don't have this Problem.

TheJackiMonster:
I think that's something to look into when making changes improving the storage system overall.

PR #789, opened 2020-07-28
Update manuskript_fr.ts

DarkRedman:
Corrected the french word for folder `Dosser` to `Dossier`


DarkRedman:
The build pipeline is broken [Update manuskript_fr.ts on Semaphore](https://semaphoreci.com/olivierkes/manuskript/branches/pull-request-789/builds/1)
Also the .ts extension is for typescript we should use .mo for non-compiled translations files.

Issue #590, opened 2019-05-30
export error (pandoc)


krdau:
When exporting to PDF I get an Error in Manuskript

![manuskript-pandoc-error2019-05-30 19-37-55](https://user-images.githubusercontent.com/14748568/58653637-6e9fdb00-8316-11e9-84e0-47916d057200.png).

 (as well as in the terminal)

![manuskript-pandoc-terminal-2019-05-27 23-57-30](https://user-images.githubusercontent.com/14748568/58654480-4b762b00-8318-11e9-988a-90442c624028.png)

The Preview/PDF output seems to be ok. I could not identify the "missing character".

worstje:
Unfortunately, this is a pandoc error and not a Manuskript error, making it difficult for us to diagnose / fix based on so little information.

Have you tried narrowing down which part of your project causes the problem by exporting only parts of it? Or if you have already done that, by making a copy and then systematically removing parts until the error starts to disappear?

I'll submit an issue on the Pandoc github to ask them to improve that error message in a way that makes sense.

worstje:
In regards to identifying the missing character, [an issue has been submitted to the pandoc project](https://github.com/jgm/pandoc/issues/5538) and even completed already. Once this is in a release, it will probably help you figure out what character is causing the warning and whether it is something you can ignore, or whether you need to truly figure out a fix. (Although we should probably still try to identify why the character isn't properly showing up on our end to begin with...)

krdau:
Through export to LaTex I found that the problem in pandoc is the **zero width space character** (U+200B):

LaTex error message:
! Package inputenc Error: Unicode char ​ (U+200B)
(inputenc)                not set up for use with LaTeX.

The problem seems to be connected to manuskript: When I edit (delete the problematic character)
the tex source-code and copy it back into manuskript, then there is no export error message.

But when editing the text in manuskript, under certain conditions this zero width space is (automatically?) included in the text.

I could reproduce the problem only now and then – therefor I do not know what exactly causes the issue.

PR #790, opened 2020-07-28
Fixed pandoc command arguments

DarkRedman:
`--base-header-level` throws a new export error because it is deprecated, now we should use `--shift-heading-level-by`


FlawlessMiner:
@DarkRedman you have set the `default` and `min` in line 113 to 0, since the value for this parameter must be one less than for the deprecated `--base-header-level` param. Pandoc documentation says this.

TheJackiMonster:
Does this PR also fix #590?

TheJackiMonster:
> Does this PR also fix #590?

I have tested it and it seems to fix #590 as well. ^^

Issue #813, opened 2020-11-05
Manuskript is counting hyphens as if they were words


magicjebb:
The title explains itself.
As a writer, I need to count words, calculate averages, etc., but these reports are compromised if Manuskript counts hyphens and other symbols in the text as if they were words.
Is there any way to fix this?
Thank you for your attention and congratulations on your great work.

emgineering:
There are a couple existing solutions to counting words in markdown files ( #816 handles hyphens, but I think we should tackle the problem more generally), but I would like some maintainer feedback before implementing one of them.

Do either of these options look like something we would want in the project? Should the user be able to toggle what features are included in their word count?

Finally, are there any special cases where something other than a space separates words? For example, I often use a double hyphen--like this--as shorthand for an em dash. Are cases like these worth handling?

Thanks!

***
#### Extra python filters
I just tried integrating code based on [this repo](https://github.com/gandreadis/markdown-word-count/blob/master/mwc/counter.py) into my fork and it works very well - it no longer includes \, \< html tags \>, etc in the word count. 
Pros:
- Granularity in choosing which markdown features we want to strip out or keep.
- Despite the added filters, no noticeable slowdown from the current method - both took up to about 1 millisecond per file.
- We can also easily specify the list of special characters (such as hyphens, floating *'s) that should be ignored.

Cons:
- It handles every feature separately. If markdown syntax ever changes (hopefully not!) it would need to be updated manually.

#### Pandoc lua filters
https://pandoc.org/lua-filters.html#counting-words-in-a-document  I haven't tried this one myself, but the idea is to walk through the markdown file as though formatting/converting it, and keep a count of words along the way.
Pros:
- It relies on pandoc - so we can guarantee consistency with final exported files, and will probably not need to worry about maintaining.
- The sample code here includes a block for filtering out words that are "entirely punctuation", like hyphens.

Cons:
- no idea what speed is like - though probably not worth worrying about.
- it relies on pandoc, and on lua (a separate language). I'm not sure what the repo policy on external dependencies is.

emgineering:
Another thought: should files/folders that have been marked as "not compiled" be included in the word count of the parent folder?

TheJackiMonster:
> Another thought: should files/folders that have been marked as "not compiled" be included in the word count of the parent folder?

I think that's something which was issued some time ago if I remember correctly but I can't find it. I would think it's best to make something like this configurable.

[...]

PR #816, opened 2020-11-16
Hyphens no longer counted in word count

James-Joe:
def wordCount(text):
    t = text.strip(" ,").replace(" ", "\n").split("\n")
    t = [l for l in t if l != "-"]
    return len(t)

New to coding and GitHub, but I believe this solves the issue raised.


TheJackiMonster:
Can you rebase your changes to the current state? Because it's very difficult to review this with all these merged commits in it... ^^'

emgineering:
Looks like the code is only written in the comment, and is not actually a part of the commits - none of the commits in this list are from James-Joe.

@James-Joe - your code looks correct! But to make a PR, you want to fork the repository and make changes to the actual codebase, and then submit a pull request. That way the changes can be merged in directly ([this may help](https://opensource.com/article/19/7/create-pull-request-github)). If you are just suggesting a code edit, a comment on an issue or another PR may be more helpful.

I'm trying to tackle the word count problem more generally over at #813 and would appreciate feedback!

Issue #833, opened 2021-02-22
Loading another project leads to crash


TheJackiMonster:
I tried to debug another issue (I tried to reproduce a hang) and I encountered a crash which is quite consistently if you have multiple projects.

- Open manuskript
- Open a recent project A
- Close project A
- Open a recent project B
- It crashes
```
bin/manuskript
Debug: Web rendering engine used: QWebView
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
Running manuskript version 0.11.0.
Preferred translation: manuskript_de.qm (based on user setting)
Loaded translation: manuskript_de.qm
"/proc/20476/root"
Loading: /home/thejackimonster/Cloud/Manuskript/Test2.msk
Detected file format version: 1. Zip: False.
Projekt /home/thejackimonster/Cloud/Manuskript/Test2.msk geladen.
Loading: /home/thejackimonster/Cloud/Manuskript/Test.msk
Detected file format version: 1. Zip: True.
Projekt /home/thejackimonster/Cloud/Manuskript/Test.msk geladen.
Traceback (most recent call last):
  File "/home/thejackimonster/github/manuskript/bin/../manuskript/ui/welcome.py", line 146, in loadRecentFile
    self.mw.loadProject(act.data())
  File "/home/thejackimonster/github/manuskript/bin/../manuskript/mainWindow.py", line 593, in loadProject
    self.makeConnections()
  File "/home/thejackimonster/github/manuskript/bin/../manuskript/mainWindow.py", line 955, in makeConnections
    self.btnAddPerso.clicked.connect(self.lstCharacters.addCharacter, F.AUC)
TypeError: connection is not unique
Fatal Python error: Aborted

Current thread 0x00007f0c6a00e740 (most recent call first):
  File "/home/thejackimonster/github/manuskript/bin/../manuskript/main.py", line 221 in launch
  File "/home/thejackimonster/github/manuskript/bin/../manuskript/main.py", line 234 in run
  File "/home/thejackimonster/github/manuskript/bin/manuskript", line 13 in <module>
make: *** [makefile:10: run] Aborted (core dumped)
```

So if I understand this correctly manuskript seems to keep data from one project which leads to a crash loading a different one. Restarting manuskript to load a different project works because the application doesn't contain any data of a project during a fresh start. Fixing this could also fix some other bugs and crashes around saving and loading projects.

[...]

PR #834, opened 2021-02-22
Properly disconnect add person connection.

BentleyJOakes:
When disconnecting, properly unhook the connection changed in previous commit.

Should fix https://github.com/olivierkes/manuskript/issues/833


TheJackiMonster:
I think line 955 should also be added in this hacky try & catch block... ^^'
`self.btnAddPerso.clicked.connect(self.lstCharacters.addCharacter, F.AUC)`

Actually seeing this code makes me want to address it properly but it should be fine then.

Issue #786, opened 2020-07-06
Missing "Index Card" background image causes crash on load


Oneymus:
When loading a project with a custom image set as the "Index Card" that doesn't exist, Manuskript fails to load the project.

Error output:
```
Traceback (most recent call last):
  File "/usr/share/manuskript/manuskript/ui/editors/tabSplitter.py", line 118, in restoreOpenIndexes
    self.mainEditor.setCurrentModelIndex(idx, newTab=True)
  File "/usr/share/manuskript/manuskript/ui/editors/mainEditor.py", line 221, in setCurrentModelIndex
    editor = editorWidget(self)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 48, in __init__
    self.setupUi(self)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget_ui.py", line 52, in setupUi
    self.corkView = corkView(self.cork)
  File "/usr/share/manuskript/manuskript/ui/views/corkView.py", line 25, in __init__
    self.updateBackground()
  File "/usr/share/manuskript/manuskript/ui/views/corkView.py", line 39, in updateBackground
    url=img.replace("\\", "/")
AttributeError: 'NoneType' object has no attribute 'replace'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/manuskript/manuskript/ui/welcome.py", line 163, in openFile
    self.mw.loadProject(filename)
  File "/usr/share/manuskript/manuskript/mainWindow.py", line 578, in loadProject
    self.mainEditor.tabSplitter.restoreOpenIndexes(settings.openIndexes)
  File "/usr/share/manuskript/manuskript/ui/editors/tabSplitter.py", line 126, in restoreOpenIndexes
    self.mainEditor.setCurrentModelIndex(QModelIndex(), newTab=True)
  File "/usr/share/manuskript/manuskript/ui/editors/mainEditor.py", line 221, in setCurrentModelIndex
    editor = editorWidget(self)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 48, in __init__
    self.setupUi(self)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget_ui.py", line 52, in setupUi
    self.corkView = corkView(self.cork)
  File "/usr/share/manuskript/manuskript/ui/views/corkView.py", line 25, in __init__
    self.updateBackground()
[...]

PR #850, opened 2021-04-04
Fixed project not opening with missing background

rbb8403:
This should fix Issue #786 where Manuskript project files wouldn't open if the custom corkboard image was missing. The problem is that findFirstFile returns None if it doesn't find a file, so I added a check to see if the image was None to prevent it from trying to open a resource which doesn't exist. This is the quick way to patch this. I'd recommend changing the findFirstFile function in functions/__init__.py for a more permanent solution, but I don't know the knock-on effects of that, so this should suffice for now. Another suggestion could be to bundle custom image resources in the project file. Or, throw in some error handling.


TheJackiMonster:
Looks fine to me for now. ^^

Issue #687, opened 2019-11-10
Distraction Free Mode Opens in Different Display


BenjaminCecchetto:
**Manuskript Version**: 0.10.0
**Python**: 3.7.3
**PyQt**: 5.12.1
**Qt**: 5.12.2
**OS**: Ubuntu 19.04
**Monitors**: Laptop as secondary display (monitor 1), external HDMI monitor as primary display (monitor 2).
**Problem**: The Manuskript window is on monitor 2. When I enter distraction free mode, either from pressing F11 or the fullscreen button on the bottom right, manuscript opens up distraction free mode on the laptop display (monitor 1)

worstje:
Where is your Manuskript window? I don't quite recall how the code works, but I _think_ it might be trying to put the distraction free environment on the same screen as the main window.

BenjaminCecchetto:
The Manuscript window is on the external display (monitor 2, primary
display) and it opens the fullscreen mode on the laptop window (monitor 1).

TheJackiMonster:
I would close this issue as duplicate of #456.

PR #851, opened 2021-04-08
Fix #456 - Force the distraction free window on the display of the main window

### LINKED ISSUE #456: select which monitor in full screen mode
I use two monitors and currently by default the second monitor is selected for full screen distraction free editing.
But I'd like to change that, use the first monitor for full screen. I haven't found any setting to configure this. 
Maybe I've overlooked it, if not, it would be good to see such an option.

thanks,
Bart

belug23:
For me and some others the distraction-free window would open on our secondary display #456 and  #687.
I suspect that it's because of my secondary display is on the left and the coordinate 0,0 is on that one.

This fix forces the window to be displayed on the same display as the main window by moving it to the coordinates of the screen of the main window and then calling `showFullScreen()`.


TheJackiMonster:
I understand, your solution makes sense. Thank you very much, I will merge it. ^^

PR #854, opened 2021-04-08
Fix #846 close Fullscreen when exiting main editor

### LINKED ISSUE #846: Fullscreen and Exit -> Fullscreen survives
When you open Full-screen and go to the Main window and end Manuskript, the fullscreen window does not close.

Escape works, then, but I expect that everything is closed, when I quit the Program.

Greetings

belug23:
If you have 2 screen you could have the main editor in between them and if you hit full-screen you could still have access to the exit buttons. When the the exit button was used with the full-screen window open, it would stay open.

This patch adds a cleaning method on the `MainEditor` to close the full-screen widget. It also adds a signal on the exiting full-screen so the `MainEditor` can clean the broken wrapper, so it'll not cause an error when trying `leaveFullScreen` with an empty C++ object.

It also removes the `__del__` function, I suspect it was there to try and solve this problem, but it was only causing an error when exiting the app because the object was already deleted.

Furthermore, it also moved duplicate code to a single function.

### DIFF
--- a/manuskript/mainWindow.py
+++ b/manuskript/mainWindow.py
@@ -794,6 +794,10 @@ class MainWindow(QMainWindow, Ui_MainWindow):
             # Remembering the current items (stores outlineItem's ID)
             settings.openIndexes = self.mainEditor.tabSplitter.openIndexes()
 
+            # Call close on the main window to clean children widgets
+            if self.mainEditor:
+                self.mainEditor.close()
+
             # Save data from models
             if settings.saveOnQuit:
                 self.saveDatas()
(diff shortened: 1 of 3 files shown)


TheJackiMonster:
If you replace the `print()` statement with the newer `LOGGER.debug()` statement and maybe squash your commits, I will merge it.

belug23:
@TheJackiMonster The print was removed in the second commit, It was a debug print, and it was not necessary.

belug23:
Ok Sorry I didn't realise I had change the `LOGGER.debug` on exit by a print... Really sorry.

For the squash you're supposed to be able to select a squash and merge option when accepting the pull request, I think that's the cleanest option, but if you insist I'll do it on my side.

TheJackiMonster:
Oh, you're right. I'll try that... no problem. ^^'

PR #856, opened 2021-04-09
Fix #855 - Avoid a crash when there's no model

### LINKED ISSUE #855: When using the + on editor or outline the application crashes
While trying to find more info on another problem I tried using the `+` button in the editor section on the develop branch and the app crashed with this error message:
```
Warning: floating outline item (has no parent or associated Outline model).
An unhandled exception has occurred!
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/outlineBasics.py", line 272, in addText
    self.addItem("text")
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/outlineBasics.py", line 283, in addItem
    item = outlineItem(title=qApp.translate("outlineBasics", "New"), _type=_type)
  File "/home/belug/dev/manuskript/bin/../manuskript/models/outlineItem.py", line 37, in __init__
    abstractItem.__init__(self, model, title, _type, xml, parent, ID)
  File "/home/belug/dev/manuskript/bin/../manuskript/models/abstractItem.py", line 57, in __init__
    self._data[self.enum.ID] = self._model.requestNewID()
AttributeError: 'NoneType' object has no attribute 'requestNewID'
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/outlineBasics.py", line 272, in addText
    self.addItem("text")
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/outlineBasics.py", line 283, in addItem
    item = outlineItem(title=qApp.translate("outlineBasics", "New"), _type=_type)
  File "/home/belug/dev/manuskript/bin/../manuskript/models/outlineItem.py", line 37, in __init__
    abstractItem.__init__(self, model, title, _type, xml, parent, ID)
  File "/home/belug/dev/manuskript/bin/../manuskript/models/abstractItem.py", line 57, in __init__
    self._data[self.enum.ID] = self._model.requestNewID()
AttributeError: 'NoneType' object has no attribute 'requestNewID'
```

I guess there's something that changed in the initialization of the views since the last release. I'll investigate into it.

belug23:
The commit bee24d45b0ae9bcd29760301bb47c7556997f5f8 Introduced an error when you don't have models and try to add an element in the tree views.

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -47,14 +47,9 @@ class abstractItem():
         if parent:
             # add this as a child to the parent, and link to the outlineModel of the parent
             parent.appendChild(self)
-        elif not model:
-            print("Warning: floating outline item (has no parent or associated Outline model).")
 
         if ID:
             self._data[self.enum.ID] = ID
-            self._model.updateAvailableIDs(ID)  # Informs the ID distributor that this ID number has been used
-        elif model:
-            self._data[self.enum.ID] = self._model.requestNewID()
 
 
 
@@ -64,6 +59,11 @@ class abstractItem():
 
     def setModel(self, model):
         self._model = model
+        if not self.ID():
+            self.getUniqueID()
+        elif model:
+            # if we are setting a model update it's ID
+            self._model.updateAvailableIDs(self.ID())
         for c in self.children():
             c.setModel(model)
 
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -53,7 +53,7 @@ class abstractItem():
         if ID:
             self._data[self.enum.ID] = ID
             self._model.updateAvailableIDs(ID)  # Informs the ID distributor that this ID number has been used
-        else:
+        elif model:
             self._data[self.enum.ID] = self._model.requestNewID()
 
 


emgineering:
Thanks for catching that! I would note that this doesn't solve the root problem that there is no model; and causes further crashes down the road (create empty project, add files, close project, try to reload -> crash. This may be because no IDs have been assigned? I'll look into it further.)

Our best bet may be to either make model a required argument (subideal) or figure out how to link the item to the main model when no model is passed.

belug23:
Sorry I missed that one because I thought I had that warning before the crash, looks like I've stared at that bisect too long and saw a false warning. I'll try to find the root cause.

emgineering:
It seems like `if parent` is always false, actually (for one, there's a line with `self._parent = None`  - but even after fixing that it seems no parent is getting passed in. This is a problem because new outlineItems naturally inherit the model from their parent...)

belug23:
in `manuskript/ui/views/outlineBasics.py` line 283 the `outlineItem` is created without parent and it's causing the problem for me.

belug23:
Would a mix of my fix and re-adding that:
https://github.com/olivierkes/manuskript/commit/bee24d45b0ae9bcd29760301bb47c7556997f5f8#diff-f4cf93cc159767af20e6853ad072b574ccdd13d257e3b9b2adcedb91b393a7ffL138

Could solve the problem? Because this part of the code looks like it handles the empty ID.

Cya tomorrow.

emgineering:
Oh nice find!

You're right about that handling the empty ID. For semantic purposes I would put that in `setModel` rather than 'insertChild` since it's part of the "adding to model" process.

Also, if the item already has an ID we would also need to inform the model as it's added. So I would do `setModel` like this:
```
def setModel(self, model):
    self._model = model
    if not self.ID():
        self.getUniqueID()
    else:
        self._model.updateAvailableIDs(self.ID())
    for c in self.children():
        c.setModel(model)
```

I would rearrange `__init__` a bit too - Your original fix took care of the reference to model in the second branch, but the first branch also references model which may eventually cause problems. And now that we are adding a unique ID on setting the model there's some redundancy there. This should take care of that:
```
# If an ID is given, set to that
if ID:
     self._data[self.enum.ID] = ID

if parent:
     # add this as a child to the parent, and link to the outlineModel of the parent
     parent.appendChild(self)
elif not model:
     print("Warning: floating outline item (has no parent or associated Outline model).")
```
We could probably also take out the warning. I didn't realize there were cases where items would be created then appended to a parent afterward, which makes it more an annoyance than anything.

Anyway, I'm now able to load/edit/close/reopen both empty projects and previous projects; I think we may be good here.

worstje:
@belug23 I haven't got much to offer in terms of the fix itself, but with the logging PR finally included in develop, I want to point out that we should not use `print` for debugging statements anymore. You probably want to use `LOGGER.warning()` instead, and leave out the manual 'Warning:' label on the string. (If LOGGER is not yet defined in the file, look in pretty much any of the other files for the two lines at top that initialize it.)

Additionally, I'd probably include the known ID of the item in question as it will help identify which item has issues when trying to diagnose the cause.

belug23:
@worstje The `print` comes from the previous commit that broke the build. After your comment I would have replaced it, but we'll remove it since it's not needed and could cause confusion.

@emgineering Ok I'll apply your suggestions and submit a new pull request.

belug23:
@emgineering Ok done and it works well. I had to validate the model in the `setModel()` since when you delete an item `setModel(None)` is called on it.

Tested with a new project everything reopens well and with an old project it's good too.

TheJackiMonster:
Can you squash both commits to one? I would merge it then.

Issue #857, opened 2021-04-09
Code and PR guidelines


emgineering:
This came up in #856 from @worstje :

> With the logging PR finally included in develop, I want to point out that we should not use `print` for debugging statements anymore. You probably want to use `LOGGER.warning()` instead, and leave out the manual 'Warning:' label on the string. (If LOGGER is not yet defined in the file, look in pretty much any of the other files for the two lines at top that initialize it.)

Is there a page anywhere listing guidelines like this for the repository? If so, let's make it more easily locatable; if not, let's make one. Are there any other major guidelines we would like to uphold where possible for new PR's?

worstje:
I think for the most part we follow PEP8, but beyond that my experience is that the codebase is a wild west of sorts. (I mean no offense to anyone who contributed by saying that, btw!)

My impression of the way original development happened was very much on a 'lets add this and get it done' approach, which is great for starting projects. However, I am also noticing a wide variety of other styles. Some parts are heavy on inheritance, other things are massively squashed away in catchall files like functions/__init__.py, and then there's also Qt who has its own inheritance-styled methodology... which we seem to partly rely on its tooling for, but there are also parts where we have basically hacked away at issues until they started to work.. until they break again. (One night long debugging session trying to figure out what the hell was broken about rearranging things in the outline view comes to mind.) How about settings which are in the preferences screen, but some are part of a project, and others are program-wide? There are dozens of very inconsistent-with-itself things in Manuskript that drove me nuts until I grew numb about them.

(Again: me saying this is not a complaint at those who got work done. It is merely how I have experienced this codebase. Many 'it works' solutions can work well when you have developers hacking away at code who are super familiar with it, but once these people disappear, it is really hard to get things done.)

Regarding the logger.. there isn't much to it for regular developers beyond what you quoted. All the hard work of setting up our own logging infrastructure based on Python's `logging` module is already taken care of. I think simply browsing around codebase for a few minutes will show perfectly how to log things as developers.

We do need people to tackle documentation, especially for end users. With the addition of the commandline interface (finally implemented using the Python `argparse` module), it will be very useful to have information on the wiki that goes into more detail. While a good usage of the `argparse` module is self-documenting, it will always be terse, which might not be good enough for some users.

belug23:
I don't think there's a page where there's a documentation of the coding guidelines. It could be a new page referenced on that page https://www.theologeek.ch/manuskript/contribute/

[...]

PR #858, opened 2021-04-09
Fixing the tests for travis-CI

belug23:
As mentioned in #857 there are tests in the project, but they've been failing for a while. This is the first step into fixing them. This is fixing the Linux part.

That PR fixes:
- 2 warnings with the deprecation of the `imp` module
- An error while splitting text caused by a self._model that is None
- An error with the test trying  to set a Model to an invalid value

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 .idea
 .project
 .pydevproject
+.python-version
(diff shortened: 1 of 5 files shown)


belug23:
Yeah Travis-CI is passing! https://travis-ci.org/github/olivierkes/manuskript/builds/766537246

Issue #852, opened 2021-04-08
Improving user-friendliness of log files


worstje:
With #667 finally merged, the final comments left there made me think that while it is a definite improvement in regards to overall code health, it isn't everything it could be yet for our users. Because that PR was originally started to preserve developer sanity and improve codebase health, ease-of-use for our end users had fallen through the cracks a little as a concern...)

The default log file is wiped by default.
---

Because the log file kept swelling immensely while I was testing (there are a lot of debug statements everywhere in the code) I implemented it to wipe the file every time we started logging a new session. (I could already anticipate people submitting log files of 1GB in size from their many writing sessions! 😂 )

This has the following problems:

❌ ❓  Running two Manuskript instances simultaneously without overriding the log file location might throw an error, strangely mix the logs or just make one disappear. _(I remember vaguely that Windows and Linux behave differently in regards to the way they handle open files, and although writing should work similarly, paranoia is a good thing...)_

❌ ❌ ❌  Users recovering from a crash by restarting Manuskript are likely to wipe the log file that could shine light on the issue they ran into.

I am not sure what the desired default behaviour should be, either.

💭  Do we want to implement some sort of builtin logrotate where we maintain our own default logs folder?

💭  Is there a proper way to detect a logfile that is currently in use for writing to perhaps skip over that one?

💭 ❕ ❕  Should we just do TIMESTAMP-PID.log files and delete ones older than 30 days?

💭 ❔  Maybe ignore the issue and leave it to some other software tool on the users system to handle... but is that even an option given our cross-platform nature? (A silently ballooning log file folder isn't nice for users.)

Accessibility
---

We probably want to offer a help menu option that shows either the log file in a file manager, opens it in an editor, and/or outright displays it in action... or a combination thereof. Log files are useless if they are hard to use for reporting issues.

I'm fine with simple solutions. As long as they are useful.

Reporting issues
---

We probably want to update some wiki templates or whatever other knows and documentation we have to clearly document the best ways to report bugs and get their hands on Manuskript.

I originally intended to just code up something for the most pressing issue of overwriting default log files, but since I don't know the preferred method, I figured it was best suited for a community discussion. I think a `(data-dir)/logs/` folder with `TIMESTAMP-PID.log` files would be most suitable when combined with a simple check that purges old logs, but I am not sure if anyone else has strong feelings on the way to approach this.. or any of these subjects, for that matter.

[...]

PR #859, opened 2021-04-10
Friendly logging for end users

worstje:
I ended up doing a bit about improving logging friendliness than was discussed in issue #852.

**This PR implements the following features:**

✔️ It now logs to a file in $datadir/logs/ of which the format includes a timestamp and process id.
✔️ The -L argument switch also benefits from this; it is now processed by a slightly improved `strftime()`, with the added placeholder *%#* which inserts the process id.
✔️ The Help menu now has a menu item that leads to a new (and currently very-dummy) page that should come to serve as a landing page for people seeking help.
✔️ The Help menu also contains a menu item to help locate the current log file. The easier we make it to submit logs, the better.

**Things I have _not_ tested:**

❔ All lines should be translateable, but I honestly have no clue how the translation system of Qt works exactly.
❔ I am not sure how well the 'Locate log file' function works on Mac and Linux as I have no way to test those for myself.


TheJackiMonster:
I don't think the logs have to be translatable... or at least they should not be translated. Otherwise it's way more difficult for developers to read a users logfile, finding an error or fixing a bug. ^^'

TheJackiMonster:
By the way on Linux with Gnome and Nautilus the fallback for opening the file explorer works fine.

worstje:
> I don't think the logs have to be translatable... or at least they should not be translated. Otherwise it's way more difficult for developers to read a users logfile, finding an error or fixing a bug. ^^'

The translatability is not in regards to the logs themselves, but in regards to the menu items and associated dialogs which offer the extra information to be concerned about.

TheJackiMonster:
I see... then it shouldn't be a problem. Qt usually scans the .ui- or its generated .py-files for strings to generate the translation dictionaries which can be filled for different languages. So as developer there is not much effort to do.

worstje:
As long as it checks the not-autogenerated `manuskript/mainWindow.py` in this commit (which is not to be confused with its twin in `manuskript/ui`!) , because that's where the simple `QMessageBox`es with some new translatable strings is hiding. :-)

TheJackiMonster:
Translation should not be an issue. I have tested and created a translation from English to German with Qt Linguist. So this should work fine.

TheJackiMonster:
I think I still gonna merge this pre-release. It adds some parts into GUI which need extra translation but I'm unsure if people have started going through that process already. In worst case I will need to merge translation files which should be fine... I assume. ^^'

Issue #860, opened 2021-04-11
Small bug with LanguageTool


belug23:
I've found a bug with the use of LanguageTool that should not cause problems to most people, mostly developers.
When you start manuskript in a fresh python setup, you have no spell checkers. After installing LanguageTool to that setup, starting the app would crash with this error message:
```
CRITICAL> An unhandled exception has occurred!
Traceback (most recent call last):
  File "bin/manuskript", line 13, in <module>
    main.run()
  File "/home/belug/dev/manuskript/bin/../manuskript/main.py", line 291, in run
    app, MW = prepare(arguments)
  File "/home/belug/dev/manuskript/bin/../manuskript/main.py", line 172, in prepare
    MW = MainWindow()
  File "/home/belug/dev/manuskript/bin/../manuskript/mainWindow.py", line 61, in __init__
    self.setupUi(self)
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/mainWindow.py", line 170, in setupUi
    self.txtSummarySentence = MDEditCompleter(self.tabSummaryPage1)
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/MDEditCompleter.py", line 18, in __init__
    dict=dict, autoResize=autoResize)
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/MDEditView.py", line 32, in __init__
    autoResize=autoResize)
  File "/home/belug/dev/manuskript/bin/../manuskript/ui/views/textEditView.py", line 83, in __init__
    self._dict = Spellchecker.getDictionary(self.currentDict)
  File "/home/belug/dev/manuskript/bin/../manuskript/functions/spellchecker.py", line 108, in getDictionary
    dictionary = Spellchecker.getDefaultDictionary()
  File "/home/belug/dev/manuskript/bin/../manuskript/functions/spellchecker.py", line 91, in getDefaultDictionary
    default = impl.getDefaultDictionary()
  File "/home/belug/dev/manuskript/bin/../manuskript/functions/spellchecker.py", line 580, in getDefaultDictionary
    default_locale = languagetool.get_locale_language()
AttributeError: module 'language_tool_python' has no attribute 'get_locale_language'
```

It looks like there's another difference between `language_check` and `language_tool_python` the `get_locale_language` function isn't directly in the LanguageTool module but in the utils submodule.

I already have a fix for it, I'm sending the pull request in a few minutes.

PR #861, opened 2021-04-11
Fix 860 languagetool get locale language

belug23:
This is a fix for #860
Another difference between `language_check` and `language_tool_python` can cause a crash when starting Manuskript after installing `language_tool_python` in a fresh python setup.
The tools have a different emplacement for the function `get_locale_language`. I've applied the same kind of fix that was done for other differences.

### DIFF
--- a/manuskript/functions/spellchecker.py
+++ b/manuskript/functions/spellchecker.py
@@ -530,7 +530,7 @@ def get_languagetool_languages():
     else:
         return languagetool.LanguageTool()._get_languages()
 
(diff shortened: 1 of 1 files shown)


belug23:
Oops I didn't realise that I have test commit in that branch, I'll fix that give me a few minutes.

belug23:
Ok now my mistake is fixed. Sorry for that.

TheJackiMonster:
Thanks for the fix.

PR #894, opened 2021-06-21
Deal with race condition.

zeth:
We cannot call `self._model.requestNewID()` if `self._model` doesn't exist yet.

It seems to work like this. I don't know if the guard could just be around self.setData?

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -203,6 +203,9 @@ class abstractItem():
     ###############################################################################
 
     def getUniqueID(self, recursive=False):
+         if not self._model:
+            return
+
         self.setData(self.enum.ID, self._model.requestNewID())
 
         if recursive:
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -203,7 +203,7 @@ class abstractItem():
     ###############################################################################
 
     def getUniqueID(self, recursive=False):
-         if not self._model:
+        if not self._model:
             return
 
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
I think it should be fine like this because I don't think the case of calling this method without a model being assigned may happen at all. Though I'm not sure if it does. That's one of my core problems with the current code base. We have way too many methods and Qt implicitly uses multi-threading at some instances which allows non-debuggable breaking code. ^^'

For example `getUniqueID()` gets called from `setModel()` and `checkIDs()`. I'm not certain but I think `setModel()` will only be called with a model excluding `None`. I'm not sure about `checkIDs()` though but the break fixed here may never happen. However the code allows it to happen.

So I would much more appreciate having less functions but them being safer. That's mostly the reason I refactor much more code than may be necessary in my fork, splitting up data uses and making access more clear. ^^

zeth:
I did get a crash from this situation. I think it's because we are not totally duplicate id safe yet when cutting and pasting scenes. I think how I have done it here might be best as if you already have duplicate id, making a load of children with messed up ids makes it worse.

TheJackiMonster:
> I think how I have done it here might be best as if you already have duplicate id, making a load of children with messed up ids makes it worse.

Yes that makes sense. Currently in [my refactored code](https://github.com/TheJackiMonster/manuskript/blob/gtk/manuskript/data/unique_id.py) the IDs should not be able to becomes duplicates to fix those issues. But this MR could improve the current situation a little.

PR #895, opened 2021-06-21
Kill old insecure pickle setting files (untested)

zeth:
Everyone with really old projects will lose their settings once but will get a new settings json file. I don't have a really old data directory to test this with.

### DIFF
--- a/manuskript/load_save/version_0.py
+++ b/manuskript/load_save/version_0.py
@@ -168,10 +168,13 @@ def loadProject(project):
     else:
         errors.append("outline.xml")
 
-    if "settings.pickle" in files:
-        settings.load(files["settings.pickle"], fromString=True)
+    if "settings.txt" in files:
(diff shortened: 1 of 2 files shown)


TheJackiMonster:
I don't have old projects as well to be honest. The example project is using the newer format as well. So I would wait until we have a project to debug this MR. I will need an old project to refactor the whole load/save functionality for the older format anyway. So maybe I will generate an older project using the current manuskript code.

TheJackiMonster:
I think I'm gonna apply this MR anyway. The old file format (which gets loaded via `version_0.py`) is from 2016. So in the upcoming version we will just apply this patch fixing the vulnerability and after that I will probably drop support for the older files. It should still be possible to convert them by using any version from 0.3.0 and upwards but it's really some legacy code which does not need maintenance since Manuskript automatically saves all projects in the newer format.

Issue #897, opened 2021-06-28
news on scripts for stage and screen 


Reaper10:
Is there any news on scripts for stage and screen syntax highlighting?

Issue #421, opened 2018-11-27
Adding more template options


RiderExMachina:
I would like to add a few more template options to the Welcome screen. These would be "TV", "Short Film", "Film", and "Play". My question is, would it be better to put these under a new "heading" (e.g. TV/Play [as seen in the screenshot]) or keep them under the Fiction "heading"?

I have attached a screenshot for reference.
![python_2018-11-26_17-16-33](https://user-images.githubusercontent.com/8017578/49052051-25bc6d80-f19f-11e8-84de-83cff725e328.png)

gedakc:
I think that if we add films/plays then creating a new section makes sense because these are different from fiction novels and non-fiction research papers.  (Note that my knowledge of the requirements for films/plays is extremely limited.)

Before you proceed, if you are interested in writing films and plays then you might want to check out another Free Software application that was written specifically for screen writing -- [Trelby](https://www.trelby.org/).  Trelby might already offer the features you seek.

Often software designed for a specific purpose can be better suited to that audience.  Currently Manuskript is designed for creating other works of fiction and non-fiction.

What are your thoughts on Trelby for screen writing?

RiderExMachina:
I haven't used Trelby, so I'll have to try that out and see. That said, I think that getting started on a "Screenwriting" section would help Manuskript in the long run.

I'm also thinking about taking TV and turning into two templates: TV - 30 Min and TV - 60 Min

gedakc:
One thing to keep in mind when presenting options to users is to try to keep it as simple as makes sense.  Ideally option lists should be kept to 7 or less items.

RE: Trelby

The [main trelby project](https://github.com/oskusalerma/trelby) has not been updated since 2012, but I believe that there is at least one [active user fork](https://github.com/auztec/OpenScreenWriter) of the code.

Reaper10:
Script types/ general writtig icons #256

Reaper10:
need a video game option

Issue #268, opened 2017-12-10
Format An Audio Drama Script 


Reaper10:
Format An Audio Drama Script  https://www.youtube.com/watch?v=LonxJjxhoPU&t=32s

PR #899, opened 2021-07-02
Add a mode to write scripts?

CavaleriOmar:
Would it be possible to exploit the considerable potential of manuskript to write screenplays?
That would be the best, so I could do worldbuilding and my manga script in one program.

### DIFF
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -24,17 +24,14 @@ MW = None
 
 
 def wordCount(text):
-    t = text.strip().replace(" ", "\n").split("\n")
-    t = [l for l in t if l]
-    return len(t)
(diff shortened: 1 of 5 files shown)


TheJackiMonster:
I think this should be an issue to request a feature rather than a merge-request. ^^'

Please look through the currently open issues if somebody is already mention your request or open up an issue if you feel a functionality is not covered by them. I think one of them could describe a similar request: #897 #268 or #421

I will close this MR because any MR should not merge to the master branch anyway. If you want to add anything through coding, you can still open up a MR to solve an open issue. I hope this helps. ^^' (There are definitely people with similar needs. So discussing how you would select such a mode or which features it would require can already help any developer to implement it.)

Issue #998, opened 2022-01-16
Crash when attempting to view folder as text


amconners:
While in the editor with a folder selected in the project tree, clicking the Text button at the bottom of the screen causes the program to crash:

```
CRITICAL> An unhandled exception has occurred!
Traceback (most recent call last):
  File "/usr/share/manuskript/manuskript/ui/editors/mainEditor.py", line 83, in <lambda>
    lambda v: self.setFolderView("text"), AUC)
  File "/usr/share/manuskript/manuskript/ui/editors/mainEditor.py", line 369, in setFolderView
    self.currentEditor().setFolderView(view)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 108, in setFolderView
    self.setCurrentModelIndex(self.currentIndex)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 312, in setCurrentModelIndex
    self.setView()
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 256, in setView
    addTitle(item)
  File "/usr/share/manuskript/manuskript/ui/editors/editorWidget.py", line 181, in addTitle
    edt = MDEditView(self, html="<h{l}>{t}</h{l}>".format(l=min(itm.level() + 1, 5), t=itm.title()),
  File "/usr/share/manuskript/manuskript/ui/views/MDEditView.py", line 30, in __init__
    textEditView.__init__(self, parent, index, html, spellcheck,
  File "/usr/share/manuskript/manuskript/ui/views/textEditView.py", line 53, in __init__
    self.setAutoResize(autoResize)
  File "/usr/share/manuskript/manuskript/ui/views/textEditView.py", line 380, in setAutoResize
    self.sizeChange()
  File "/usr/share/manuskript/manuskript/ui/views/textEditView.py", line 372, in sizeChange
    self.setMinimumHeight(docHeight)
TypeError: setMinimumHeight(self, int): argument 1 has unexpected type 'float'
```
As far as I can tell it's another instance of the same issue that causes #992.

Tested on the package repository version of Manuskript 0.13.1 on Arch Linux.

PR #999, opened 2022-01-16
Fix for #992 and #998

### LINKED ISSUE #992: Crash when changing index card style
Clicking the "Old style" option in Settings -> Views -> Index Cards causes Manuskript to instantly crash with the following output:

```
  File "/usr/share/manuskript/manuskript/ui/views/corkDelegate.py", line 214, in paint
    self.paint_v1(p, option, index)
  File "/usr/share/manuskript/manuskript/ui/views/corkDelegate.py", line 491, in paint_v1
    r = self.mainLineRect.adjusted(-m, -m, m, m / 2)
TypeError: adjusted(self, int, int, int, int): argument 4 has unexpected type 'float'
WARNING> QPaintDevice: Cannot destroy paint device that is being painted
```
This was in Manuskript 0.13.1 on Arch Linux, installed via the package repository.

amconners:
This PR fixes #992 and #998 by explicitly casting to int the args that were causing the crashes.

### DIFF
--- a/manuskript/ui/views/corkDelegate.py
+++ b/manuskript/ui/views/corkDelegate.py
@@ -488,7 +488,7 @@ class corkDelegate(QStyledItemDelegate):
         fullSummary = item.data(Outline.summaryFull)
         if lineSummary or not fullSummary:
             m = self.margin
-            r = self.mainLineRect.adjusted(-m, -m, m, m / 2)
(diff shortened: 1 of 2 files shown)


TheJackiMonster:
Thanks for implementing the fix. Hopefully this have been all remaining instances of this issue. ^^'

PR #1000, opened 2022-01-17
hard_line_breaks support

Mte90:
So I found the commit implementing in pandoc to set the pandoc minimum version https://github.com/jgm/pandoc/commit/8a101cffe36f4d6301ff45fd1b491bad1fcf1244

This should fix a part of a request there.

~~I was not able to test it.~~ I tested it and works :-)


TheJackiMonster:
Thank you for implementing. I will merge it as soon as the CI won't fail anymore. There's currently a problem with the Ubuntu CI test which doesn't seem to be related to that change but I don't want to merge code before the CI works properly again.

Mte90:
Perfect, I will wait it isn't a problem :-)

TheJackiMonster:
Okay, I have restarted the CI without changes and the Ubuntu CI works fine now. So I think this can be merged. Anyway I adjusted the Mac CI as well because half the time of each run it only applied updates via homebrew for packages we don't need. ^^'

Issue #719, opened 2020-01-24
outliner folder gets renamed and contents replaced when dragged over other folder, only in upward direction


bartart3d:
Hi all,

I've got a weird, but serious issue with the latest manuskript edition, 0.11, but I can't reproduce it with a newly created document.

In the outliner, when I drag a folder up, so, more to the beginning of the document tree, the first folder it is being dragged over ends up being renamed to the same name as the folder that I am dragging. The contents is being replaced too. Wen dragging a folder down the document tree, further away from the beginning, this does not happen. Perhaps this info can pin down the cause of it.

First I had automatic saving on when exiting. Luckily I had a very recent backup. From now on I uncheck 'save on project close' for every new file I make. Over the years I have encountered way too many bugs causing data loss. I strongly suggest this to become the default setting for the time being.

bartart3d:
I just reïnstalled 0.10.0-1, and there it happens too.

gedakc:
Did you **expand** the destination folder and look inside?

For example when I drag the **Chapter 5** folder on top of the **Chapter 1** folder then the **Chapter 5** folder becomes a child of the **Chapter 1** folder.

![Manuskript-Drag-Folder-Up-Example](https://user-images.githubusercontent.com/10405019/73107233-89811a00-3eba-11ea-99b8-f1ce10047505.png)

When I tried the steps you mentioned, the screen shot above is what occurred.  Hence I am unable to reproduce the issue you described.

bartart3d:
Hi,
@gedakc : thanks for looking into it. I couldn't reproduce it with a new file either, but no, it's not what you describe. 
The end result of the bug are two different folders having te same name, with the same contents, not nested.

gedakc:
I also tried dragging a lower folder and dropping on top of a higher folder with an existing project and was unable to reproduce the error you experienced.

If you are able to determine the steps with a new project that cause the issue then that would go a long way towards identifying the problem.

bartart3d:
I'm starting to think it is because of a slow, or bad disk. I already mentioned in earlier reports the slow rendering of a folder full of text files as one long text. It could take minutes to appear on the screen.

I've been looking inside the manuskript file at what happens when dragging folders and moving them around, and if I understand it correctly, files and folders are actually being renamed or copied, or overwritten, to change the order since there is no database or xml tree involved in the actual document. I only saw an xml file for the plots.

[...]

Issue #1001, opened 2022-01-19
Re-ordering items in the outline causes data loss/text overwritting


bnichs:
This one has been really hard to reproduce so I apologize for however vague this is, but the symptoms are that text is unrecoverably deleted so it seems important. 

What I'm seeing is that if I have a structure like:

* Folder 
    * Chapter 1.md 
    * Chapter 2.md 

_sometimes_ if i drag Chapter 2 to be in front of Chapter 1, Chapter 1 will be deleted and _replaced_ with chapter 1. This does not happen every time or even in every project, but it happens. I suspect it has to do with the filenames colliding, but repro has been fairly hard to do with a blank project so something else is happening in the background which i dont understand. 

The log I provide below is for a blank project but in order to repro, i did have to copy in actual text and actual titles, "Chapter 1" and "Chapter 2" wont do it, but for some reason my chapter titles do. In the log I had a chapter "Pressure drop" and then a few just called "New." When I dragged pressure drop to be at the top, it replaced "New." I was then able to continue until all 3 Texts were the same "Pressure Drop" chapter and all the "new" were gone.   

Details that might matter: 
* AFAIK, this only happens with "revision mode" on, i haven't had enough luck reproducing to test otherwise. 
* I use folder save mode. Not sure if it matters, but the overlapping IDs log makes me wonder if the save format matters. 
* Every time it happens, the project is saved immediately. I'm sure this is just an event hook, but it makes irrevocable bugs like this a lot more painful because you can't even ctrl-z your way to safety. 
* I think this might be a symptom of the file names having the order in their name. I get that this is part of the file format chosen, but as a user, when I re-arrange texts, i don't ever expect the contents to change, just the metadata (names, order, etc).

```openjdk version "11.0.13" 2021-10-19
OpenJDK Runtime Environment (build 11.0.13+8-Ubuntu-0ubuntu1.21.04)
OpenJDK 64-Bit Server VM (build 11.0.13+8-Ubuntu-0ubuntu1.21.04, mixed mode, sharing)
openjdk version "11.0.13" 2021-10-19
OpenJDK Runtime Environment (build 11.0.13+8-Ubuntu-0ubuntu1.21.04)
OpenJDK 64-Bit Server VM (build 11.0.13+8-Ubuntu-0ubuntu1.21.04, mixed mode, sharing)
Saving to: zip
Project ordering-test.msk saved.
Saving to: zip
Prose has been renamed (outline/0-New → outline/0-Prose)
[...]

PR #1002, opened 2022-01-21
Potential fix for #719, #1001: data loss when reordering items in upward direction

olivierkes:
I got some issue with this (#719, #1001) as well.

With that fix, I haven't been able to reproduce the bug.

# Description of the fix

In my testing, it happens only when items have duplicate IDs (though I haven't been able to understand why):

> WARNING> There are some items with overlapping IDs: ['5', '5']

And this happens *at least* if items are copied and pasted (or dragged and dropped with `CTRL`).

The issue comes from here:

https://github.com/olivierkes/manuskript/blob/0182a43b8ec5bf16217da659078843d201388b2b/manuskript/models/abstractModel.py#L429-L441

Since `getUniqueID` does not do anything if item has no model:

https://github.com/olivierkes/manuskript/blob/0182a43b8ec5bf16217da659078843d201388b2b/manuskript/models/abstractItem.py#L207-L215

and at that moment in the drop, items have no models.

Well, the easier fix here is to simply remove items IDs, so they get new ones when theyre created, which this commit does.

# Further things to consider

*IF* this effectiverly fixes this bug, in my mind the `WARNING> There are some items with overlapping IDs:` should be treated with more seriousness. It shouldn't be allowed that IDs are not unique.

### DIFF
--- a/manuskript/models/abstractItem.py
+++ b/manuskript/models/abstractItem.py
@@ -224,14 +224,22 @@ class abstractItem():
         if max([self.IDs.count(i) for i in self.IDs if i]) != 1:
             LOGGER.warning("There are some items with overlapping IDs: %s", [i for i in self.IDs if i and self.IDs.count(i) != 1])
 
+        _IDs = [self.ID()]
         def checkChildren(item):
+            "Check recursively every children and give them unique, non-empty, non-zero IDs."
             for c in item.children():
                 _id = c.ID()
-                if not _id or _id == "0":
+                if not _id or _id == "0" or _id in _IDs:
                     c.getUniqueID()
+                    LOGGER.warning("* Item {} '{}' is given new unique ID: '{}'".format(_id, c.title(), c.ID()))
+                _IDs.append(_id)
                 checkChildren(c)
 
         checkChildren(self)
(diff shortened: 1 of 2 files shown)


olivierkes:
After some sleep, I decided it will be much safer to fix duplicate IDs on project opening, if this is actually the cause of data loss.
The tradeof is that some references might not work anymore, since IDs are used to reference a scene. But between data loss and the possibility of a few broken links for the subset of people using that feature, I think the choice is clear.

Now since I don't understand why duplicate IDs were the cause of the data loss, I guess more testing is needed to see if this really fixes #719 or not.

TheJackiMonster:
> After some sleep, I decided it will be much safer to fix duplicate IDs on project opening

I think we actually do this already in `manuskript/models/abstractItem.py`:
https://github.com/olivierkes/manuskript/blob/d56c2225a36d46a27644914255eec8b1e2783b73/manuskript/models/abstractItem.py#L223-L234

So the question is do we still have another cause for invalid/non-unique IDs besides local file changes? Otherwise I think this PR could fix the issues with the reordering via drag&drop quite well.

PR #1003, opened 2022-01-25
Fix crash when regex is not valid (Fixes olivierkes#989)

jdanielp:
Wrap regex preparation in try-catch
Log when there is a problem preparing the regex

### DIFF
--- a/manuskript/ui/search.py
+++ b/manuskript/ui/search.py
@@ -13,6 +13,8 @@ from manuskript.enums import Model
 from manuskript.models.flatDataModelWrapper import flatDataModelWrapper
 from manuskript.ui.searchMenu import searchMenu
 from manuskript.ui.highlighters.searchResultHighlighters.searchResultHighlighter import searchResultHighlighter
+import logging
+LOGGER = logging.getLogger(__name__)
 
 
 class search(QWidget, Ui_search):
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
Could you disable regex in the search UI if it's already the `import re` failing the use of it already? Otherwise we create UI which is misleading, I assume. Otherwise this looks as a good fix for now. Thank you.

jdanielp:
I assume that  "import re" does not fail since it's standard library functionality. But, it would be easy enough to disable the checkbox. Problem is that the search functionality uses regex for all search, so if re doesn't import all search is dead in the water. Have I misunderstood you?

I was thinking we could change the text color in the lineedit when the regex is not valid. A better approach might be to put a fake search result, e.g. "bad regular expression" in the results with more feedback or add feedback to the status bar when the regex is not valid

TheJackiMonster:
Mhmm... I also assumed `import re` would never fail at first. But it's inside the try & catch clause. So I wasn't sure. ^^'

Well... if it shouldn't fail, I think it should be imported outside the function like the other imports.

jdanielp:
Yeah, I wrapped the whole function without discriminating. I'll fix that. Should we open a separate enhancement for indicating the regex is wrong or do you still want to handle it as part of this PR?

jdanielp:
I went ahead and added the text color change when the regex fails to compile since it was straightforward.  Also will update when the search menu changes (checking, unchecking items).  I tested on my system (Windows 10) with a dark system theme.

jdanielp:
Is there anything else needed on this PR?  I've fixed all the issues.

TheJackiMonster:
I'll test it soon and merge it then.

PR #1004, opened 2022-01-25
Fix crash when files are locked for writing (Fixes olivierkes#950)

jdanielp:
Wrap file opens in try-catch
Applied fix for project file as well as other files that are opened during save

### DIFF
--- a/manuskript/loadSave.py
+++ b/manuskript/loadSave.py
@@ -63,6 +63,6 @@ def loadProject(project):
     LOGGER.info("Detected file format version: {}. Zip: {}.".format(version, isZip))
 
     if version == 0:
-        v0.loadProject(project)
+        return v0.loadProject(project)
     else:
-        v1.loadProject(project, zip=isZip)
+        return v1.loadProject(project, zip=isZip)
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -327,7 +327,6 @@ def saveProject(zip=None):
     # Save to plain text
 
     else:
-
         global cache
 
         # Project path
@@ -378,11 +377,17 @@ def saveProject(zip=None):
                 LOGGER.debug("* Writing file {} ({})".format(path, "not in cache" if path not in cache else "different"))
                 # mode = "w" + ("b" if type(content) == bytes else "")
                 if type(content) == bytes:
-                    with open(filename, "wb") as f:
-                        f.write(content)
+                    try:
+                        with open(filename, "wb") as f:
+                            f.write(content)
+                    except PermissionError as e:
+                        LOGGER.error("Cannot open file " + filename + " for writing: " + e.strerror)
                 else:
-                    with open(filename, "w", encoding='utf8') as f:
-                        f.write(content)
+                    try:
+                        with open(filename, "w", encoding='utf8') as f:
+                            f.write(content)
+                    except PermissionError as e:
+                        LOGGER.error("Cannot open file " + filename + " for writing: " + e.strerror)
 
                 cache[path] = content
 
@@ -412,8 +417,12 @@ def saveProject(zip=None):
                     pass
 
         # Write the project file's content
-        with open(project, "w", encoding='utf8') as f:
-            f.write("1")  # Format number
+        try:
+            with open(project, "w", encoding='utf8') as f:
+                f.write("1")  # Format number
+        except PermissionError as e:
+            LOGGER.error("Cannot open file " + project + " for writing: " + e.strerror)
+            return False
 
         return True
 
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -429,13 +429,14 @@ def saveProject(zip=None):
             LOGGER.error("Cannot open file " + project + " for writing: " + e.strerror)
             filesWithPermissionErrors.append(project)
 
-        dlg = ListDialog()
-        dlg.setModal(True)
-        dlg.setWindowTitle(dlg.tr("Files not saved"))
-        dlg.label.setText(dlg.tr("The following files were not saved and appear to be open in another program"))
-        for f in filesWithPermissionErrors:
-            QListWidgetItem(f, dlg.listWidget)
-        dlg.exec()
+        if len(filesWithPermissionErrors) > 0:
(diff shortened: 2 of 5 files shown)


jdanielp:
This fix does not add message boxes to warn user as that would be a design change.  Please let me know how you want that handled and point me to an example in the code so I know how to implement in the same style as what exists.  Thanks!

TheJackiMonster:
I assume a prompt or message box needs to be implemented depending on the call of the save() function. For example a manual save failing should result in a message box warning that the save-action failed. But automatic saves should not create potentially thousands of message boxes stopping the user writing. ^^'

If you could implement a message box only in manual saving, this would be awesome. Thank you.

jdanielp:
I'll add a message box for manual save.

I'll need to create a list of files that don't save since it's more than just the project file that can cause the problem. Since that can be a long set of files, I might use a custom UI to add a scrollable list of files.

This will take me a little bit while I learn how the project deals with .ui files and code. I'm a c++ & Qt dev for my day job, but I have PyQt experience - shouldn't take me too long to get acquainted.

TheJackiMonster:
Alright, looking forward to it. I think we already use a message box to show the list of files being potentially deleted with an action. This was merged with the PR #932. So maybe that's a way to do it even though an actual listbox is probably prettier. ^^

jdanielp:
Great! I'll have a look

jdanielp:
That PR doesn't use a custom UI.  It uses a QMessageBox.  If there were many items to delete that message box would be very tall and potentially be taller than the screen.  I'll go the custom UI route with a scrollable list.  Shouldn't be too bad.  Just need to dust off my Designer chops.  I always have to re-learn how to make layouts fill the entire dialog. :)  Tomorrow or the next day.

jdanielp:
When testing this, you can use the following powershell script to lock all txt files in the directory


`gci -Filter *.txt -Recurse | ForEach-Object -Process {[System.io.File]::Open($_, 'Open', 'Read', 'None')}`

jdanielp:
Looks like the problem also happens on open when files are locked.  I'll have a look to see what can be done there as well.

jdanielp:
I've fixed the crash on save and open, and display a Dialog containing all the files that cause problems on open.  This PR is ready for review.

jdanielp:
I've fixed the crash on save and open, and display a Dialog containing all the files that cause problems on open.  This PR is ready for review.

TheJackiMonster:
I've tested the code as good as I could (it is certainly not easy to lock files for me on Linux... I assume most editors don't lock files you are editing these days ^^').

I noticed an issue, I think. When I enforce the new dialog on loading a project during startup, I get a segmentation fault directly after closing the dialog. I assume it is not the fully correct way to open and close the dialog properly with the synchronized call of `exec()` but I'm not sure. However I have managed to fix that by using the method `open()` instead and I also found that the main window wasn't set as parent already which failed the dialog to show up at all during saving.

So to adjust both issues, you can use this patch or apply something similar manually. The other part of this patch is that the `loadProject()` function in `loadSave.py` didn't return the error list yet which might be an issue in general.
```
From 88d73f7415be4f67273c8321d8d1b9b8cae6858b Mon Sep 17 00:00:00 2001
From: TheJackiMonster <thejackimonster@gmail.com>
Date: Sun, 30 Jan 2022 04:02:49 +0100
Subject: [PATCH] Add main window as parent of dialogs and use open() instead
 of exec() to avoid segfaults

Signed-off-by: TheJackiMonster <thejackimonster@gmail.com>
---
 manuskript/loadSave.py            | 4 ++--
 manuskript/load_save/version_1.py | 8 ++++----
 2 files changed, 6 insertions(+), 6 deletions(-)

diff --git a/manuskript/loadSave.py b/manuskript/loadSave.py
index 7657376..1adfe4c 100644
--- a/manuskript/loadSave.py
+++ b/manuskript/loadSave.py
@@ -63,6 +63,6 @@ def loadProject(project):
     LOGGER.info("Detected file format version: {}. Zip: {}.".format(version, isZip))
 
     if version == 0:
-        v0.loadProject(project)
+        return v0.loadProject(project)
     else:
-        v1.loadProject(project, zip=isZip)
+        return v1.loadProject(project, zip=isZip)
diff --git a/manuskript/load_save/version_1.py b/manuskript/load_save/version_1.py
index da0bc3e..b2ad66c 100644
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -430,13 +430,13 @@ def saveProject(zip=None):
             filesWithPermissionErrors.append(project)
 
         if len(filesWithPermissionErrors) > 0:
-            dlg = ListDialog()
+            dlg = ListDialog(mw)
             dlg.setModal(True)
             dlg.setWindowTitle(dlg.tr("Files not saved"))
             dlg.label.setText(dlg.tr("The following files were not saved and appear to be open in another program"))
             for f in filesWithPermissionErrors:
                 QListWidgetItem(f, dlg.listWidget)
-            dlg.exec()
+            dlg.open()
 
         if project in filesWithPermissionErrors:
             return False
@@ -920,13 +920,13 @@ def loadProject(project, zip=None):
     mdl.rootItem.checkIDs()
 
     if len(filesWithPermissionErrors) > 0:
-        dlg = ListDialog()
+        dlg = ListDialog(mw)
         dlg.setModal(True)
         dlg.setWindowTitle(dlg.tr("Files not loaded"))
         dlg.label.setText(dlg.tr("The following files were not loaded and appear to be open in another program"))
         for f in filesWithPermissionErrors:
             QListWidgetItem(f, dlg.listWidget)
-        dlg.exec()
+        dlg.open()
 
     return errors
 
-- 
2.35.1
```

Otherwise the PR seems fine to me and I will merge it as soon as this or a similar patch is applied. Thank you for your contributions so far.

jdanielp:
Good catch.  Sorry about that.  I'll have a look on Sunday or Monday to apply your patch and see if there is anything else I might have missed.

jdanielp:
I applied the patch per your suggestion.  I still need to do some testing on my end.

TheJackiMonster:
@jdanielp Can this be merged or do you still need to do more testing?

jdanielp:
I think it can be merged

PR #1007, opened 2022-01-30
Change welcome page spacer size policy to allow welcome to expand to fill most of screen (fixes olivierkes#996)

jdanielp:


jdanielp:
We might consider changing the minimum size of the spacers, but otherwise this PR is ready for review.

katafrakt:
I checked on Linux and MacOS and it works well :+1:  The spacing around the widget is maybe a bit too big for my liking, but I guess it's a matter of taste. Definitely an improvement over what we have now.

PR #1013, opened 2022-02-18
char-count goal implement & fix translation using weblate (Japanese)

KeitoTobi1:


TheJackiMonster:
Can you explain what exactly your PR implements? Then I would have a look at it. Because checking the source code is one thing. Checking if it works as intended is another. ^^'

Otherwise it's best to commit translations on Weblate only. They get merged before each release anyway and it's better if translators resolve their conflicts on Weblate potentially understanding each other than me deciding which translation looks more exact by guessing (because I can only understand English and German really well... so I shouldn't resolve conflicts with translations in most cases).

PR #1015, opened 2022-02-19
catch AttributeError in plotTreeView double click event (fixes #1014)

### LINKED ISSUE #1014: [Bug] double click causes crash in outline or plots view
If you double click in the empty space (i.e. not on any plot) in the list on the left of the outline or plots views, Manuskript will crash with the following output:

```
  File "/usr/share/manuskript/manuskript/ui/views/plotTreeView.py", line 204, in mouseDoubleClickEvent
    if item.parent():
AttributeError: 'NoneType' object has no attribute 'parent'
```

Tested on Arch Linux using both the 0.13.1 binary from the software repository and the latest version from the develop branch.

amconners:


TheJackiMonster:
I have adjusted the fix to avoid a try & catch block if possible. It should fix the mentioned issue in a similar way though. Maybe you can verify that it works for you and I will merge it.

amconners:
Works just fine for me.

PR #1056, opened 2022-07-28
Create a macOS installer (fixes #567)

### LINKED ISSUE #567: Package as App Bundle for macOS

marosoft:
An initial version of the app bundle and the installer to allow people to test it and see if there are any issues with it.

Fixes #567

The zip file and the installer are available @ https://drive.google.com/drive/folders/1b-4zxMYRGX5EICVTwdDyE6kKXGOGuX0D


kirmaha:
Hello,

I installed Manuskript on my Mac (OS 12.0.1) by drag-dropping manuskript.app in the Application folder. When opening the app, at first, Mac OS X runs a "verification" test which looks like succeeding. Then, the app bumps for some time in the dock, and it ends up with a pop up window saying that manuskript.app can't be launched as Apple can't verify that it is free of any malware.
Capture d’écran 2022-08-06 à 14 27 21
In OS X security settings, a button appears, allowing you to start Manuskript "anyway" (as in "even if we told you it could be a dangerous malware").

Then, the software opens and seems to run well. I didn't encounter anything wrong, except a unexpected "&" in tabs names :
Capture d’écran 2022-08-06 à 14 56 46

I did some writing, editing, compiling, and it didn't crash. No special characters issues either.

Note that I have no previous experience with Manuskript, not sure I'm the most reliable tester ;-)
Not sure this kind of newbie feedback helps but I hope it does.

TheJackiMonster:
Sounds good. I mean the lack of official verification is something we need to deal with on any OS. So this is something to be expected. The "&" characters in the UI are probably a separate issue in translation files or the UI itself.

Thanks for testing.

TheJackiMonster:
Today I could verify that the provided .dmg file works properly once a user has worked around the security warning. However I have problems reproducing such a file with the adjusted build scripts. On macOS Big Sur I get a warning when I try to run the `prepare_osx.sh` scripts with the line `brew postinstall python` stating it wouldn't complete the step successfully.

It's kind of weird to see that because it runs fine on the xcode image from Travis-CI, I think. But on the machine it fails to install pip3 because of this and I can't build the rest. I'll try to get it working but I also researched that the new essential tools required to build the .dmg file are `codesign` and `dmgbuild` (both should be running via Python). So I was wondering if I could be able to generate the whole file on my Archlinux system if I get those tools running.

@marosoft Do you have any insight on this whether it's possible? This would ease a lot for upcoming releases like my script to generate the package for Windows via wine.

TheJackiMonster:
@marosoft The line `package/osx/rebuild_mac_icon.sh` in `build_osx.sh` does not work. I think you missed to add this script to your commit.

marosoft:
> Today I could verify that the provided .dmg file works properly once a user has worked around the security warning. However I have problems reproducing such a file with the adjusted build scripts. On macOS Big Sur I get a warning when I try to run the `prepare_osx.sh` scripts with the line `brew postinstall python` stating it wouldn't complete the step successfully.
> 
> It's kind of weird to see that because it runs fine on the xcode image from Travis-CI, I think. But on the machine it fails to install pip3 because of this and I can't build the rest. I'll try to get it working but I also researched that the new essential tools required to build the .dmg file are `codesign` and `dmgbuild` (both should be running via Python). So I was wondering if I could be able to generate the whole file on my Archlinux system if I get those tools running.
> 
> @marosoft Do you have any insight on this whether it's possible? This would ease a lot for upcoming releases like my script to generate the package for Windows via wine.

I am sorry I did not reply earlier.

The code itself is compiled by `pyinstaller` and then have to be re-sign using `codesign` as there is a known issue with `pyinstaller`. When that is completed you have the app already available as a zip archive.
Then `dmgbuild` is used to create a proper installer.

I am not familiar with code signing and/or dmg building on the other OS than macOS. I am not if that is possible.

marosoft:
> @marosoft The line `package/osx/rebuild_mac_icon.sh` in `build_osx.sh` does not work. I think you missed to add this script to your commit.

@TheJackiMonster thanks for spotting this one. I actually noticed that script in the #570 PR and I borrowed it from there to create the icon file. 
I can remove that line from the script or if @dschanoeh is ok with that I could include it in this PR.

dschanoeh:
@marosoft sure, please go ahead.

TheJackiMonster:
Ok, I will wait for you adding the script. Then I think this should be ready to get merged.

marosoft:
I added the missing script and rebased my branch.

I created a new build manually using the script on my Mac and pushed the files into: https://drive.google.com/drive/folders/15vCEZD4TTUVkMP6fxXiFQmy2ePtHkTsu

I called it 0.14.1 just to differentiate it from the previous build.

holdenmad:
> I added the missing script and rebased my branch.
> 
> I created a new build manually using the script on my Mac and pushed the files into: https://drive.google.com/drive/folders/15vCEZD4TTUVkMP6fxXiFQmy2ePtHkTsu
> 
> I called it 0.14.1 just to differentiate it from the previous build.

This works for me, I'm using Monterey and an M1 Macbook. It was a little slow, and I needed to manually right-click to open the application like kirmaha mentioned, but it looks like it's working fine to me.

PR #1088, opened 2022-11-19
Small UI Typo String Fixes

MC42:
image

On current builds (Windows, 0.14.0) this string shows with an ampersand in the middle, and seems to be related to this QT Creator .ui file.  This, in theory, should remove that odd ampersand.  If someone could confirm that who has a full dev environment set up, that'd be ideal.

Thank you in advance :)


MC42:
So I got a Windows setup to run this, and it _seems_ to be working at least, and the menus open and render the text correctly.

Where those ampersands were introduced is news to me, but this at least means the search bar renders correctly.  Fixed both the en_US_GB translation and the sourcecode UI file in theory, as well as normalized them a teensy bit.

MC42:
And now the dockable title for the Cheatsheet is more correct and doesn't include this typo.  Rebuilt the mainWindow.py for good measure too.

image

MC42:
This latest change also causes the icon for the Cheat Sheet section to render correctly; it seems my instance didn't have it prior so it was an empty button... but this does work at least?  idk.

TheJackiMonster:
Thank you for contributing.

Please note that changes to the translation files should be done via [Weblate](https://hosted.weblate.org/projects/manuskript/translations/) though. The reason for this is that if they get adjusted on Github and Weblate on the same time, conflicts may appear and somebody needs to solve them. However this requires at least the ability to read/understand the given language to decide which change is the better solution.

For the changes here it should be fine but please use Weblate in the future. The UI and code changes can be made via Github of course. That's totally fine. I will look into these before the next release to get them merged.

MC42:
I didn't realize translations were expected to be done through Weblate at time of PR creation, apologies.  The wiki entry [here ](https://github.com/olivierkes/manuskript/wiki/Translate-Manuskript)had said you _can_ use Weblate, implying a priority of GitHub > Weblate.

I'll see about using Weblate shortly.  Thanks for the help :)

TheJackiMonster:
What about the other changes you've made so far? Do you really want to close this PR? ^^'

MC42:
Re-opened the PR, good point.  Most of it was still string translations though.  I'll clean it up and squash it shortly to hopefully avoid conflicts between source and Weblate.  The non-translation source-strings will stay, as will the small UI tweaks (and then this PR can be renamed for the third time haha)

As to Weblate, I don't seem to have permissions to suggest a new translation there as it's "already translated".  I was largely going to clean up the English errata addressed here, but more correctly, haha!  If you can do anything to help about that, I'd be most appreciative.

I did update the wiki page to reflect the current behavior though and added a date disclaimer though.

TheJackiMonster:
I think you should be able to make suggestions because I'm not aware of other users having any additional permissions. But be aware that there are two languages listed as English. The fully translated option is the list of source strings which can't be edited because it would break all other languages. But you can make suggestions or changes to [English (United Kingdom)](https://hosted.weblate.org/projects/manuskript/translations/en_GB/), I assume. Hope this helps.

TheJackiMonster:
I think you definitely got rid of the ampersands. But in all languages it uses "Cheatsheet" now instead of the translation. "Search" still seems to get translated fine from my testing.

I've updated the upstream repository now with all translations done in Weblate. So now there are some conflicts because you made changes to similar files. If you can solve the conflicts, I can merge this. It's also possible to drop the changes in those files and apply them via Weblate. Maybe that would solve the issue with the "Cheatsheet" label but in worst case that needs to be translated again.

MC42:
Strange, and I seem to have properly broke the PR... Dangit... I'll see why it's not localizing properly momentarily.

MC42:
I totally mucked up this branch.  Just going to re-fork and make a new PR.  The history will be complicated enough from this.  Ty for the help, sorry for the hassle haha.

TheJackiMonster:
No worries... ^^'

PR #1091, opened 2022-12-10
Resolve "Search" and "Cheat sheet" strings not being translated correctly in source string

MC42:
Recreating the changes in a smaller manner, I can also replicate the presence of "Cheat Sheet" not localizing for some reason.  it does appear to be configured the same otherwise, which is part of why this is so strange.  

Changing the source string (like "Sea&rch" was to "Search") seems inconsistent at best.  Very strange.


MC42:
it seems that to update a source string, you need to run (like the wiki says) the `lrelease i18n/manuskript_XX.ts` step for your language, or the makefile step for it, which expands it to all items in the repo.  This isn't an automatic step though, as on Windows you cannot trivially use the make expansion.

Should have it fixed in a bit for all languages.

PR #1092, opened 2022-12-15
Add GitHub Actions Windows CI build to commits

MC42:
This will permit automatic Windows builds in CI/CD when the develop branch has new commits made and/or when people make Pull Requests, to test them.  It does *not* run pytest.  The commit hash SHA short will be in the "About" screen of the build artifacts.

Example screenshot demonstrating the hash short version functionality.

image


TheJackiMonster:
Looks good but I would recommend to remove some of the libraries after pyinstaller has finished. Because otherwise the build will not work on Windows 7 and Windows 8, I assume. More details [here](https://github.com/olivierkes/manuskript/blob/develop/package/build_for_windows.sh) in lines 83 to 91.

Maybe we can even remove more Qt5 libraries from the final build which we don't need, reducing the size further. But at least the files I removed in the script using Wine are necessary as far as I've tested (the exact reasons depend on Windows though... ^^').

Besides do you think it's possible to have a macOS CI via Github actions as well to build a binary? Just in case the travis CI is not able to make a release because I'm not sure whether the FTP server to store the macOS binary still runs. ^^'

TheJackiMonster:
I think in case of Qt5 libraries, they have been moved by newer pyinstaller versions into a different subdirectory. So they need to be removed in `PyQt5/Qt5/bin/Qt5Bluetooth.dll`. I will just correct it.

PR #1096, opened 2022-12-24
Add settable color palettes

TEParsons:
Allows users to specify what colour palette they want the app to show in. For now, I've added a dark and light palette but I tried to make it as easy as possible for other contributors to add colour schemes.

I couldn't get my head around the Settings UI and wasn't sure how best to implement it even if I could, so I've set it to default to light and left the settings option for more experienced contributors to figure out. So long as there's a setting called `applicationPalette`, its value should be used in much the same way as `style` to set the palette, and I've tested it by manually switching between `themes.light` and `themes.dark` and the palette works!

Light palette:
image

Dark palette:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 .settings/org.eclipse.core.resources.prefs
 .vscode
 .vimrc
+desktop.ini
 ExportTest
 Notes.t2t
 dist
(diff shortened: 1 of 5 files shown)


TheJackiMonster:
I've added a switch to the settings dialog now. It's still not fully functional. For example the theme won't get recogniced when opening the dialog to set the current index of the combobox properly.

Also I think it would make sense if different palettes have a name defined in their python files instead of doing that in the settings dialog. So see this as work-in-progress if anything.

It's also not syncing changes to the settings yet.

Either way because the feature adds new translatable strings to the UI (palette names and the label in the settings panel), I would think it makes sense to keep this feature open for a future release rather than the very next. Feel free to implement the missing parts I mentioned if you don't want to wait for me to do it. ^^

Issue #1095, opened 2022-12-23
Search no longer works


jrstrick:
As of the develop branch pulled on December 22,2022, the search function crashes Manuskript every time I try to use it. Regex seems to choke on the search string (which was *the* in this example, minus the asterisks.) It throws the following error:

  File "/home/jim/manuskript/bin/../manuskript/functions/__init__.py", line 472, in search
    return [(m.start(), m.end(), getSearchResultContext(text, m.start(), m.end())) for m in searchRegex.finditer(text)]
TypeError: expected string or bytes-like object

Full dump follows. Some file names have been edited out to protect the innocent. I'm unclear what's causing the 'strange things in file' errors. Are my directories out of sync with the folder's metadata file?

```
Logging to STDERR.
Logging to file: /home/jim/.local/share/manuskript/manuskript/logs/2022-12-23_16-45-31_manuskript#52369.log
Operating System: Linux-6.0.12-76060006-generic-x86_64-with-glibc2.35
Hardware: x86_64 / x86_64
Running from unpackaged source code.
* sys.executable = '/usr/bin/python3'
* sys.argv = ['/home/jim/bin/manuskript']
* sys.path = ['/home/jim/manuskript/bin',
 '/home/jim/manuskript/bin/..',
 '/usr/lib/python310.zip',
 '/usr/lib/python3.10',
 '/usr/lib/python3.10/lib-dynload',
 '/home/jim/.local/lib/python3.10/site-packages',
 '/usr/local/lib/python3.10/dist-packages',
 '/usr/lib/python3/dist-packages']
* sys.prefix = '/usr'
Manuskript 0.14.0#a36bc55 (Python 3.10.6)
* PyQt 5.15.6 (compiled against Qt 5.15.3)
  * Qt 5.15.3 (runtime)
* lxml.etree 4.8.0.0
  * libxml   2.9.13 (compiled: 2.9.13)
  * libxslt  1.1.34 (compiled: 1.1.34)
* pyEnchant N/A (libenchant: N/A)
* pySpellChecker N/A
* Symspellpy N/A
* Markdown 3.3.6
Web rendering engine: QTextEdit
Preferred translation: ['en'] (based on available ui languages)
No translation found or loaded. (manuskript_en.qm)
Using the builtin translation. (U.S. English)
openjdk version "11.0.17" 2022-10-18
OpenJDK Runtime Environment (build 11.0.17+8-post-Ubuntu-1ubuntu222.04)
[...]

PR #1098, opened 2022-12-27
Fix to error  #1095

TheShadowOfHassen:
Error #1095 is an error where searching crashes when typing  anything it gives an error:

  File "/home/jim/manuskript/bin/../manuskript/functions/__init__.py", line 472, in search
    return [(m.start(), m.end(), getSearchResultContext(text, m.start(), m.end())) for m in searchRegex.finditer(text)]
TypeError: expected string or bytes-like object

It happened to me also. I was able to fix it by just putting a str() around the text in the last bit so its:
    return [(m.start(), m.end(), getSearchResultContext(text, m.start(), m.end())) for m in searchRegex.finditer(str(text))]

This is my first pull request ever so sorry if I did anything wrong. Tell me and I'll fix any problems.

### DIFF
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -469,7 +469,7 @@ def search(searchRegex, text):
     :return:                list of tuples (startPos, endPos)
     """
     if text is not None:
-        return [(m.start(), m.end(), getSearchResultContext(text, m.start(), m.end())) for m in searchRegex.finditer(text)]
+        return [(m.start(), m.end(), getSearchResultContext(text, m.start(), m.end())) for m in searchRegex.finditer(str(text))]
     else:
         return []
 


jrstrick:
I made this patch manually and it fixed the error for me, too.

TheJackiMonster:
Thanks for providing the quick solution. I assume the issue appeared when Python updated its implicit type conversion policy.

TheShadowOfHassen:
Alright do I need to do anything else or is it all on the Collaborators/Maintainers? I've looked it up I couldn't get a clear answer

TheJackiMonster:
> Alright do I need to do anything else or is it all on the Collaborators/Maintainers? I've looked it up I couldn't get a clear answer

Don't worry. I just waited for our new automatic tests to complete. It's merged now and will be part of the upcoming release.

Issue #1107, opened 2023-01-07
Gtk Port: Character More info bug


TheShadowOfHassen:
So for my first addition to the Gtk port I decided to implement a feature that I've wanted for years. A template for Your More Details in character. It's the listbox thing. I always found that it really saps your writing momentum when you originally start with the characters and then take ten minutes to set someones eye color/ hair color etc.

So to my point. I noticed that in the gtk port manuskript uses a dictionary to hold its more info `{DataName:DataValue}` the original had two inside lists `[[DataName,DataValue]]` This is less a bug and more a potential bug when we get saving done, unless you count my template and then its a bad bug. Dictionaries allow only one name for each thing so no `{DataName:Eggs, DataName:Bacon}` That makes sense because you wouldn't have Hair color:Red Hair color: Blondm, but its not intuitive. That is way to much of a learning curve for anyone and frankly it'd be easier to fix than add a warning checker if you add two or more with the same name.

@TheJackiMonster If this is a problem and you don't have some brilliant design scheme I'll fix it within the next few days and put it in my pull request with the Template and the other bug fixes I've made. I just wanted to be sure though before I started.

TheJackiMonster:
Using a dictionary for those values is an intentional design choice. The reason for this is that you can't refer to any attribute otherwise because they wouldn't have a unique identifier. So I was thinking about referencing character attributes in text or descriptions and it didn't make much sense to not implement it with a dictionary then.

Obviously we need to add warnings here avoiding that users create duplicates accidentally and data gets dropped during saving process. But the UI allows temporarily that multiple values under the same name are stored. So showing a warning in UI should be fine as well as a dialog to rename attributes or merging their values when saving.

TheShadowOfHassen:
OK I think I know how to fix this. I'll add a check through on the code so that if you have a duplicate atrabute it'll just automatically put a 2 or 3 or etc at the end to clue people in. It'd fix any saving errors also. I'll add it to my pull request I'm working on.

TheJackiMonster:
Hmmm, unsure if that's ideal. It's certainly a way many applications deal with that issue. However I don't think that's a solution for plotting anything with exact custom properties. I would mark all matching attributes in red or orange, show a notification with a warning and add a dialog when saving explaining it will cause an issue if continued.

[...]

PR #1111, opened 2023-01-08
Gtk pull request with way to many things

TheShadowOfHassen:
Sorry for the size. I made some of the changes with a downloaded zip file of the branch, and it just dawned on me that most pull requests are one thing at a time. Big omnibus pull requests won't happen again.

So I'm including a change-log, so you know everything I changed. Also, as a note you can probably tell anyway because I comment a lot. So You'll be able to tell if I looked at something because I comment on it.

So first it's probably pointless, but I changed the about screen to show the most recent version of Manuskript (0.14.0) changed the website to show the link. Personally I like to know what I'm clicking on, so I assumed others would, and finally I added a note that it's still in development. Just in case someone else stumbles upon it.

Second I remade the book of acts example because the other ran errors when I tried to open it.

Third I added a test.io in the main directory. It can be removed if you don't want it there, but I couldn't get it to run in its original place. Also, I added an Icon to the app.

Fourth because I wanted to get the hang of the manuskript port/libhandy I added what I've wanted for years. A template for the more info. It always sapped my writing speed having to keep over and over adding things for each character's hair and eye color and all the other stuff.

Next I fixed issue #1107 at least temporary I know @TheJackiMonster and I were discussing different ways to do this but its easy enough to change back if needed.

Finally there is a tip of the day glade file and a blank box pane now in the overview glade file They can be deleted, but it's the start of another feature I want to add, but I wanted to discuss the feature before I implemented it. Yes I know that if I just implement 100 things I want in manuskript it'll still not work but I can't figure out custom drag and drop.


TheJackiMonster:
You removed some `.gitignore` files with your commits (for example with [2470b89](https://github.com/olivierkes/manuskript/pull/1111/commits/2470b89ff8cdb3e65e6f17dc9b91d092eef251d4)). Please put them back in place to not commit redundant backup files from Glade or Gedit. You can also add all files starting with `._` to the gitignore mask, I assume.

Also please do not commit the sample project a second time or other projects upstream if not necessary. This will only inflate the commit size and makes finding and checking code changes more difficult.

TheShadowOfHassen:
Sorry. I didn't know I'll fix it but I have to destroy this pull request ( think) I'll get another pull request back in a moment.

PR #1136, opened 2023-03-04
Fix grammatical error in Brazilian Portuguese

samuel-s-marques:
Changed "vizualizar" to "visualizar" to correct a grammatical error in the text. "Vizualizar" is not a correct spelling in Brazilian Portuguese, so it was replaced with "visualizar". This change improves the accuracy and readability of the text in Portuguese.

### DIFF
--- a/i18n/manuskript_pt_BR.ts
+++ b/i18n/manuskript_pt_BR.ts
@@ -574,7 +574,7 @@
     <message>
         <location filename="../manuskript/ui/mainWindow.py" line="1538"/>
         <source>&amp;Frequency Analyzer</source>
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
Thanks for your contribution, however please use [Weblate](https://hosted.weblate.org/projects/manuskript/translations/) for any contributions in the translation files. Otherwise I can not ensure there won't be conflicts between translations on Weblate and here in the repository which is a problem since I can't resolve any conflicts of translations from languages I don't speak.

samuel-s-marques:
Thank you. I didn't know I needed to use Weblate. I'm going to fix and translate the language there.

Issue #1041, opened 2022-06-20
[Enhancement] Keyboard shortcut for "Insert Reference"


AKANexus:
Unless I missed something, there could be a keyboard shortcut to "Insert Reference". I found myself instinctively pressing Visual Studio's `Ctrl+.` shortcut to do it, but any available keyboard shortcut would be interesting.

Also, not sure if it warrants a separate issue, but a "back" button would also be appreciated after clicking a reference hyperlink. Most often than not I find myself needing to check/add information to a linked entry and then going back to where I was. (Think about Visual Studio's `Ctrl+-`)

TheJackiMonster:
In the notes or summary from the metadata area you can press `Ctrl+Space` to add a reference which should be common in most IDEs for auto-completion.

AKANexus:
Ah, great. Wasn't aware of that. What about the "Back" shortcut?

Also, I think I _may_ have found a bug:
I left the computer unattended for almost 1.5h with Manuskript open on a page I was editing. When I came back (the computer hadn't entered sleep mode), I double clicked a word, and pressed `Ctrl+Space` and it just crashed.

> 2022-06-20 18:43:46,440 - manuskript.logging - CRITICAL - An unhandled exception has occurred!
Traceback (most recent call last):
  File "manuskript\ui\editors\completer.py", line 59, in submit
AttributeError: 'NoneType' object has no attribute 'data'

I reopened Manuskript, chose the same word, pressed the same combo, and it worked flawlessly.

TheJackiMonster:
I don't think there's a shortcut to go back yet.

PR #1165, opened 2023-05-13
Simple history back/forward navigation

alfar:
Like the title says, I implemented a simple history that keeps track of what places you've visited. This was in response to #1041.

It works pretty much like the history of a web-browser, so if you go back some steps and then navigate to a new place, it cuts off your forward history and adds the new place to the end.

I'm pretty new to python programming, so I wasn't sure if there's a library I was supposed to use for the Signals, but I tried to make it look/feel like the signals from Qt so it wouldn't stick out too much.

I'm not 100% sure of the hot keys. ctrl+< and ctrl+> works for me because I have a Danish keyboard layout. I reckon it should work for an English keyboard as well. Main reason I decided against backspace/shift-backspace is that if you end up in a text field, you'll remove letters instead of navigating. Likewise, alt+left/alt+right seems to be used by the tree components to navigate around there, so half the time, my history actions would get eaten.

Any feedback would be appreciated.


JPhB:
I'd like to use it in version 7.0. How do I do that, not knowing anything about Python?

alfar:
> I'd like to use it in version 7.0. How do I do that, not knowing anything about Python?

Well, I would think that it's not version 7.0 anymore if you add code to it...

https://github.com/olivierkes/manuskript/wiki has a section explaining how to run Manuskript from code on various operating systems. You could grab my version from https://github.com/alfar/manuskript and try running that, but it's based on the latest version of the source code, probably somewhere after version 0.15.0 (which seems to be the latest release)

You'd have to fork my code and cherry-pick my commit into your fork of the older version and then probably fiddle a ton to make it play together. I'm not saying you can't, just wondering if it's that important to use a specific version?

Of course, I'm new to this project, so I don't know if there's something magical and wonderful (or stable) about version 7.0 :)

JPhB:
Thanks for responding. Is your version available on Windows? When I follow the link you included and try to download Manuskript (https://www.theologeek.ch/manuskript/download), I get this error:
Fatal error: __autoload() is no longer supported, use spl_autoload_register() instead in /home/clients/90412ec2d8e55d60c960a48d035ee4d5/web/manuskript/wp-includes/compat.php on line 502

alfar:
I'm coding on a Windows machine, seems to be working ok. You'll need to download the source from https://github.com/alfar/manuskript and then run it in the way described on the other link.
Good luck :)

TheShadowOfHassen:
You should probably update to the latest version if you still run. 0.7...

Also the download error has happened before: #1175 It apparently is an issue.

TheJackiMonster:
I've tested it and it works fine. I don't have any issues with the key bindings but maybe we can have some configuration for the key bindings anyway to make sure it works for everyone.

Anyway good job and thanks for the contribution.

TheJackiMonster:
@alfar One small note: I didn't notice it because the code was running fine on my system. But we use Python 3.9 on the automatic testing system and potentially even an older version when shipping towards Windows to have maximum compatibility with older versions of Windows. So match and case syntax does not work yet on all systems. I've pushed a fix though.

alfar:
> @alfar One small note: I didn't notice it because the code was running fine on my system. But we use Python 3.9 on the automatic testing system and potentially even an older version when shipping towards Windows to have maximum compatibility with older versions of Windows. So match and case syntax does not work yet on all systems. I've pushed a fix though.

Oh, I'm sorry, didn't realize. Thanks for fixing it - hope it wasn't too much of a hassle. :)

TheJackiMonster:
> > @alfar One small note: I didn't notice it because the code was running fine on my system. But we use Python 3.9 on the automatic testing system and potentially even an older version when shipping towards Windows to have maximum compatibility with older versions of Windows. So match and case syntax does not work yet on all systems. I've pushed a fix though.
> 
> Oh, I'm sorry, didn't realize. Thanks for fixing it - hope it wasn't too much of a hassle. :)

Nah, I replaced it with some `if` and `elif` branching. Not as fast theoretically but it should be fine for navigation.

PR #1113, opened 2023-01-10
Gtk Pull Request with way to many things (Again)

TheShadowOfHassen:
First here is everything that the old closed pull request says:

Sorry for the size. I made some of the changes with a downloaded zip file of the branch, and it just dawned on me that most pull requests are one thing at a time. Big omnibus pull requests won't happen again.

So I'm including a change-log, so you know everything I changed. Also, as a note you can probably tell anyway because I comment a lot. So You'll be able to tell if I looked at something because I comment on it.

So first it's probably pointless, but I changed the about screen to show the most recent version of Manuskript (0.14.0) changed the website to show the link. Personally I like to know what I'm clicking on, so I assumed others would, and finally I added a note that it's still in development. Just in case someone else stumbles upon it.

Second I remade the book of acts example because the other ran errors when I tried to open it.

Third I added a test.io in the main directory. It can be removed if you don't want it there, but I couldn't get it to run in its original place. Also, I added an Icon to the app.

Fourth because I wanted to get the hang of the manuskript port/libhandy I added what I've wanted for years. A template for the more info. It always sapped my writing speed having to keep over and over adding things for each character's hair and eye color and all the other stuff.

Next I fixed issue https://github.com/olivierkes/manuskript/issues/1107 at least temporary I know @TheJackiMonster and I were discussing different ways to do this but its easy enough to change back if needed.

Finally there is a tip of the day glade file and a blank box pane now in the overview glade file They can be deleted, but it's the start of another feature I want to add, but I wanted to discuss the feature before I implemented it. Yes I know that if I just implement 100 things I want in manuskript it'll still not work but I can't figure out custom drag and drop.

So there still is two more msk files there and if you don't want them I'll delete them next pull request(The only way I can think of doing it now is deleting this fork and starting all over again, but They were necessary for me because the original book of acts ran errors. I'll open an issue on why.

PR #1181, opened 2023-06-21
Pull request with too many things (again, again)

TheShadowOfHassen:
I decided to do this on a new branch and new PR because it was easier.

This is the same PR as #1113


TheJackiMonster:
This is definitely much better to review and merge now. For example the changed files dropped from 31 to 12 and lines of code deleted/changed from existing code from 428 to only 28.

I hope it makes sense why I wanted this to be clean? ^^'

The PR is now actually only adding features instead of changing everything.

TheShadowOfHassen:
So do you want to fix the conflicts or should I?

TheJackiMonster:
> So do you want to fix the conflicts or should I?

I'll fix the conflicts. Don't worry about it.

TheJackiMonster:
image

The information overlays spawn on the left side of the application window. I think the anchor point needs to be set properly. Would be good to correct this.

Also I don't really like the alignment of the elements. Buttons should all be aligned in a column on the right, I think. 

image

Then the names can already cause user error. So either the names should be limited or better the labels could use ellipsis and show the actual name on hover as proper overlay (in case template names are changed externally without respecting a hardcoded limit).

image

The icon of replacing the template should be adjusted as well. I recommend using the icons from Adwaita (best to use symbolic icons) which is typically a GTK default and other themes will adjust from it. You can use the GTK icon browser application to lookup them easily to pick one.

TheShadowOfHassen:
I'll make the changes as soon as I can.

TheShadowOfHassen:
I changed the alighments of the buttons and the icon. Was there anything else.

TheShadowOfHassen:
Yay! Thanks!

TheShadowOfHassen:
@TheJackiMonster I realized that because this PR had do many things in it it also had a fix for the character more info that would force them to use unique names for the first part because you use dictionaries to store the more info. Does that need merged to the gtk normal? Because without it, if you create things with duplicate names the data in the UI is not what is being stored.

TheJackiMonster:
For now I'll go with the same behavior as in the Qt version. So I think it would need to be addressed differently. In the Qt version you can actually store multiple values under the same name in the character detailed info.

TheShadowOfHassen:
OK, just this PR changed that so if you have something else in mind it'll have to be changed back (eventually)

PR #1194, opened 2023-08-06
welcome screen word count fix

### LINKED ISSUE #1163: Fedora 38 application crashes when cycling through the home options  at first launch
Not sure how to fix the issue 
```python
2023-05-05 18:01:33,134 - manuskript.main - INFO - Using the builtin translation. (U.S. English)
2023-05-05 18:01:35,777 - manuskript.logging - CRITICAL - An unhandled exception has occurred!
  File "/usr/share/manuskript/manuskript/ui/welcome.py", line 260, in changeTemplate
    self.updateTemplate()
  File "/usr/share/manuskript/manuskript/ui/welcome.py", line 329, in updateTemplate
    self.updateWordCount()
  File "/usr/share/manuskript/manuskript/ui/welcome.py", line 370, in updateWordCount
    self.template[1][templateIndex][1])
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
IndexError: list index out of range
```

Dreaded-Gnu:
Added set of templateIndex to None when it shall be deleted ( after call of `deleteLater` within `clearLayout` ). Also adjusted updateWordCount to check whether templateIndex is not None, before updating the total amount.

Source of the issue is that the method `deleteLater` schedules deletion without removing it directly. That leads to the issue during switch between templates on welcome screen two templates are considered for update of word count.

Fixes #1163

### DIFF
--- a/manuskript/ui/welcome.py
+++ b/manuskript/ui/welcome.py
@@ -277,6 +277,7 @@ class welcome(QWidget, Ui_welcome):
                 i = l.takeAt(0)
                 if i.widget():
                     i.widget().deleteLater()
+                    i.widget().setProperty("templateIndex", None)
                 if i.layout():
                     clearLayout(i.layout())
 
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
I fixed the crash via https://github.com/olivierkes/manuskript/commit/14392909c3592a7546e9c76f914349c98749e02d and adjusted the logic to calculate `total` as well because it was actually wrong in some cases. Unsure if this is fixed here as well.

Issue #1169, opened 2023-05-31
Crash on Mac not on windows with same project


Niwreg:
Hi,

I'm have installed Manuskript both on my Mac and on an windows 10 machine. I've got an NAS share where the project folder is stored (not into 1 file just the folder structure) When opening on windows there is no issue but when I open the project on Mac it crash. the logs on the Mac report the following:

> cat 2023-05-31_11-52-12_manuskript\#35931.log         
> 2023-05-31 11:52:12,452 - manuskript.logging - INFO - Logging to file: /Users/gerwinvanderkamp/Library/Application Support/manuskript/manuskript/logs/2023-05-31_11-52-12_manuskript#35931.log
> 2023-05-31 11:52:12,474 - manuskript - INFO - Operating System: macOS-11.7.6-x86_64-i386-64bit
> 2023-05-31 11:52:12,474 - manuskript - INFO - Hardware: x86_64 / i386
> 2023-05-31 11:52:12,474 - manuskript - INFO - Running in a frozen (packaged) state.
> 2023-05-31 11:52:12,474 - manuskript - DEBUG - * sys.frozen = True
> 2023-05-31 11:52:12,474 - manuskript - DEBUG - * sys._MEIPASS = /Applications/manuskript.app/Contents/MacOS
> 2023-05-31 11:52:12,474 - manuskript - DEBUG - * sys.executable = '/Applications/manuskript.app/Contents/MacOS/manuskript'
> 2023-05-31 11:52:12,474 - manuskript - DEBUG - * sys.argv = ['/Applications/manuskript.app/Contents/MacOS/manuskript']
> 2023-05-31 11:52:12,475 - manuskript - DEBUG - * sys.path = ['/Applications/manuskript.app/Contents/MacOS/base_library.zip',
>  '/Applications/manuskript.app/Contents/MacOS/..',
>  '/Applications/manuskript.app/Contents/MacOS/lib-dynload',
>  '/Applications/manuskript.app/Contents/MacOS']
> 2023-05-31 11:52:12,475 - manuskript - DEBUG - * sys.prefix = '/Applications/manuskript.app/Contents/MacOS'
> 2023-05-31 11:52:12,475 - manuskript - INFO - Manuskript 0.15.0-041fda5-041fda5 (Python 3.10.8)
> 2023-05-31 11:52:12,475 - manuskript - INFO - * PyQt 5.15.7 (compiled against Qt 5.15.2)
[...]

PR #1198, opened 2023-08-12
added extra check for the filebase dir not to start with an .

Niwreg:
added an extra check in load_save/version_1.py to it seems fix bug #1169


Niwreg:
Hi,

If someone might want to look at this, never wrote something in python before, and I'm not sure if there are more places this might neet fixing

TheJackiMonster:
I think your code could actually replace the check above. I'll look into it. But it should already be good to merge this fixing #1169.

TheJackiMonster:
Seems fine like this.

PR #1205, opened 2023-08-20
Skip comments in word count

sagev9000:
Hi! I'm not sure if others have run into this, but there have been a few times were I've left notes laying around and thought I was closer to my word count goal than I actually was. Wanted to see if others might appreciate dropping comments from the word count.

Notable caveat: I'm not sure a good way to handle the possibility of comment tokens inside of a code block without re-implementing a lot more parsing logic. I also understand that the current solution is much much more elegant than this proposal.

So, I'd happily take suggestions and fully support shooting this down, but I figured it was worth a shot. Maybe someone sharper than me would have a better idea for it :sweat_smile:

Thanks for taking a look!

### DIFF
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -34,10 +34,9 @@ def wordCount(text):
     count = 0
     in_comment = False
     for word in words:
-        if in_comment:
-            if word == '-->':
-                in_comment = False
-        else:
+        if in_comment and word == '-->':
+            in_comment = False
+        if not in_comment:
             if word == '<!--':
                 in_comment = True
             else:
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -30,18 +30,7 @@ def safeTranslate(qApp, group, text):
         return text
 
 def wordCount(text):
-    words = re.findall(r"\S+", text)
-    count = 0
-    in_comment = False
-    for word in words:
-        if in_comment and word == '-->':
-            in_comment = False
-        if not in_comment:
-            if word == '<!--':
-                in_comment = True
-            else:
-                count += 1
-    return count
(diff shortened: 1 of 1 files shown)


TheShadowOfHassen:
I'm not familiar with comments in Markdown, but this is in an idea that I think is worth it. I sometimes write notes to myself and while they won't throw off my word count much It's a nice thing to have.

What is a comment in markdown exactly?

TheJackiMonster:
> I'm not familiar with comments in Markdown, but this is in an idea that I think is worth it. I sometimes write notes to myself and while they won't throw off my word count much It's a nice thing to have.
> 
> What is a comment in markdown exactly?

Markdown does not support comments. So in Manuskript you do HTML comments which wil not be included into the exported document since Pandoc translates Markdown into HTML as intermediate, I assume. That's also the reason you can use HTML tags inside the Markdown files and it will likely apply properly to the final document. I've used this for alignment in the past for example.

Anyway an issue I have with this MR is that it will cause a significant hit to performance/latency of the editor. Because we calculate the word count quite a lot (every time the text changes) and it has been a bottle-neck in the past already. Also there are still exceptions when the code suggested would not work properly (HTML comments do not require any spacing for example to work, so opening and closing comments might be in the middle of a regex matching word).

Instead I suggest a different approach: We could use a regular expression to remove all comments from our text and search words using the existing regular expression on that result. Using regex will be much faster because it's implemented inside the Python runtime in C instead of using Python loops. We can avoid the edge cases, I mentioned, as well.

TheJackiMonster:
This should do it:

```
def wordCount(text):
    len(re.findall(r"\S+", re.sub(r"(<!--).+?(-->)", "", text)))
```

sagev9000:
@TheJackiMonster Brilliant

PR #1212, opened 2023-09-04
Fix regex for ignoring multi-line comments.

sagev9000:
`.*` doesn't pick up newlines, so multi-line comments were not being correctly dropped from the word count.
`[\s\S]` is a simple workaround, though any of `[\w\W]`, `[\d\D]`, or `(.|\s)` should also work.

### DIFF
--- a/manuskript/functions/__init__.py
+++ b/manuskript/functions/__init__.py
@@ -30,7 +30,7 @@ def safeTranslate(qApp, group, text):
         return text
 
 def wordCount(text):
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
You are right. But I think we should use the `re.DOTALL` or `re.S` flag instead since it's more legible. See [here](https://stackoverflow.com/questions/33889952/python-re-sub-multiline-on-string) and in the [docs](https://docs.python.org/3/howto/regex.html#compilation-flags).

PR #1214, opened 2023-09-13
Update spellchecker.py And Fixed crashing

TheShadowblast123:
Corrected issue where words within certain quotations would show up as incorrectly spelled (ex: 'I'm fine') would indicate both words are misspelled
image
In the image, "I'm fine" is confirmed to be in the dictionary as in a previous line it is not showing up as misspelled, however in single quotes it doesn't work. I made sure that it would work with any additional punctuation and not just apostrophes or single quotes just in case.


TheShadowblast123:
I found the editor was crashing so I decided to see if I could fix it myself.
image
I found that textColor only gets set whenever there's text, but if paint gets called without there being text it would crash because it's trying to call setPen and passing an undefined variable as a value. 
image
My solution was to use try and except to set define textColor. I believe that a better method would be to set textColor to a default value based on the user's theme, but I'm not even sure what was calling the function to begin with.

Issue #1224, opened 2023-09-19
Crash on spellcheck context menu


MatHuf:
When right clicking on a misspelled word to open the context menu Manuskript crashes.
This appears to be when a word has no suggestions (very misspelled or not in dictionary) as it works just fine for misspelled words that have suggestions.
This is reproducible for any word that has no suggestions.

OS: Manjaro
Manuskript Version: 0.15.0
Software Versions in Use:
     Python 3.11.5
     PyQt 5.15.9
     Qt 5.15.8
Dictionary: pyspellchecker

2023-09-19 17:38:23,941 - manuskript.logging - CRITICAL - An unhandled exception has occurred!
Traceback (most recent call last):
  File "/usr/share/manuskript/manuskript/ui/views/textEditView.py", line 496, in contextMenuEvent
    popup_menu = self.createStandardContextMenu()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/manuskript/manuskript/ui/views/textEditView.py", line 590, in createStandardContextMenu
    suggestions = self._dict.findSuggestions(text, cursor.selectionStart(), cursor.selectionEnd())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/manuskript/manuskript/functions/spellchecker.py", line 220, in findSuggestions
    match.replacements = self.getSuggestions(word)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/manuskript/manuskript/functions/spellchecker.py", line 366, in getSuggestions
    if word in candidates:
       ^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable

PR #1226, opened 2023-09-20
Crash on spellcheck context menu #1224

drmousse:
Pull request for issue 1224:
Right clicking on unknown words (for spellchecker pyspellchecker and symspellpy) lead to crash. 
In these cases the object was None and therefore not iterable. The fix in code should solve this issue.


MatHuf:
Commenting to add a link to the issue so this PR shows up as mentioned there: https://github.com/olivierkes/manuskript/issues/1224

PR #1227, opened 2023-09-20
Prevent crash when spellcheck candidates is None

MatHuf:
This addresses a crash when right clicking a misspelled word that has no suggestions. Documented in this issue: https://github.com/olivierkes/manuskript/issues/1224


MatHuf:
Closing this as I didn't see there was already a PR with a fix: https://github.com/olivierkes/manuskript/pull/1226

Issue #1299, opened 2024-06-06
Crash By Clicking Outline


djsmith4570:
[2024-06-06_16-44-46_manuskript#7612.log](https://github.com/user-attachments/files/15686817/2024-06-06_16-44-46_manuskript.7612.log)

I set my qt theme to fusion after reading a previous issue from back in 2019. It still crashes every time I click on the outline section regardless of themes. I'm assuming there's something else going on here. I've tried nothin and I'm all outta ideas.

djsmith4570:
2024-06-06 16:44:46,097 - manuskript.logging - INFO - Logging to file: /home/smith-d/.local/share/manuskript/manuskript/logs/2024-06-06_16-44-46_manuskript#7612.log
2024-06-06 16:44:46,107 - manuskript - INFO - Operating System: Linux-5.15.0-112-generic-x86_64-with-glibc2.35
2024-06-06 16:44:46,108 - manuskript - INFO - Hardware: x86_64 / x86_64
2024-06-06 16:44:46,108 - manuskript - INFO - Running from unpackaged source code.
2024-06-06 16:44:46,108 - manuskript - DEBUG - * sys.executable = '/usr/bin/python3'
2024-06-06 16:44:46,109 - manuskript - DEBUG - * sys.argv = ['/usr/bin/manuskript']
2024-06-06 16:44:46,109 - manuskript - DEBUG - * sys.path = ['/usr/bin',
 '/usr/share/manuskript/',
 '/usr/lib/python310.zip',
 '/usr/lib/python3.10',
 '/usr/lib/python3.10/lib-dynload',
 '/usr/local/lib/python3.10/dist-packages',
 '/usr/lib/python3/dist-packages',
 '/usr/lib/python3.10/dist-packages']
2024-06-06 16:44:46,110 - manuskript - DEBUG - * sys.prefix = '/usr'
2024-06-06 16:44:46,113 - manuskript - INFO - Manuskript 0.12.0 (Python 3.10.12)
2024-06-06 16:44:46,113 - manuskript - INFO - * PyQt 5.15.6 (compiled against Qt 5.15.3)
2024-06-06 16:44:46,113 - manuskript - INFO -   * Qt 5.15.3 (runtime)
2024-06-06 16:44:46,137 - manuskript - INFO - * lxml.etree 4.8.0.0
2024-06-06 16:44:46,137 - manuskript - INFO -   * libxml   2.9.13 (compiled: 2.9.13)
[...]

PR #1315, opened 2024-08-13
Fix crash when clicking outline

ByteOfBrie:
Fixes the crash in #1299, which ended up being very simple.

I wrote a decent bit of explanation in both of the commits, but for the short summary: At least in the more recent versions of PyQt5, `subElementRect` for `qproxystyle` (which is used in `paint`) expects three arguments, instead of having a default value for widget of `None`.

This shouldn't be a problem even if older versions do have a default value, since it would already be `None`.

### DIFF
--- a/manuskript/main.py
+++ b/manuskript/main.py
@@ -220,7 +220,7 @@ def launch(arguments, app, MW = None):
             kernel.initialize(['python', '--matplotlib=qt'])
 
             # Create the console in a new process and connect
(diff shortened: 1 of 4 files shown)


ByteOfBrie:
I managed to get another crash still based on this, so it's apparently not ready to merge quite yet.

ByteOfBrie:
This is ready to be merged whenever you want, it's totally done on my end. There aren't any more crashes around this behavior that I'm aware of.

Issue #1323, opened 2024-09-09
Crash when opening one specific .msk


kgollihu:
I'm consistently having Manuskript crash when I try to open one specific .msk file on Ubuntu 22.04.4. I am able to open Manuskript to the file select screen, but once I choose this one .msk, it immediately closes. It successfully opens other files.

The log file after opening Manuskript but before opening the file:
![image](https://github.com/user-attachments/assets/e05b0ad7-fc30-49fa-8b84-db5288b4c4b5)
![image](https://github.com/user-attachments/assets/6054b011-665c-4f75-9f7b-531bcdabda6c)
![image](https://github.com/user-attachments/assets/f5fe8d2b-4b7d-44a3-88f8-9376aa1c2302)
![image](https://github.com/user-attachments/assets/41e509ca-6838-43af-b475-e13d3961a6f9)

The log file after attempting to open the problematic .msk:
![image](https://github.com/user-attachments/assets/3c6f2406-360a-4825-a0bc-cdba07b8b0ba)

These are the files in the .msk folder, in case that's relevant:
![image](https://github.com/user-attachments/assets/664f9daa-84de-45ce-8265-f7f3407099aa)

phenomenoeme:
It was the image files.

I removed each of the image files to a new folder and removed that from the .msk source folder. That allowed me to open the .msk file.

Issue #1117, opened 2023-01-20
Gtk Port: Saves Discussion


TheShadowOfHassen:
Okay, before I say anything. It is assumed that for all feature versions of Manuksript there would be backwards compatibility. If not instant loading by some import settings.

So with the Gtk port we're actually pretty close to needing a save. The character tab has 100% functionality (I think and so does the general and summary and plots) The world just needs drag and drop. The actual editor needs a lot of work but some of that work could actually hinge on what way we want.

I think that the save method needs to have these following criteria:
1. Auto save( Anything can do this)
2. Revisions Issue #871 talks about using Git and I think that's a smart Idea. I've used Git for my one project and so far its working quite well with a obsidian project for world building and a manuskript in folder mode for the story. It also kept track of my maps in Krita.
3. No crashing and data loss. I know people who have had this.
4. Everything in easy to read/ edit without Manuksript format. Story as .MD files, plots as .xml etc. the project settings can be something like json, but it'd be still nice to be able to edit them in last resort. Some of this could be worked around with exporting but still one of the cool things I liked with Manuksript is even if it disappeared of the face of the earth I could still use my story.

So currently Manuskript has two options, a file and a folder mode. I like the idea of a compressed zip file with the .msk name, however I don't know if the git revisions will work on a compressed format and also it might be smart to set up some sort of streaming saving/loading thing for Manuksript because if you have 30 text files with 5000 + words they don't need loaded always. I think we don't have this problem with Qt manuksript but you get my point. Also if we're going to do an update with images, python's native Zipfile module has no easy way to copy images into it.

Personally I don't like the folder way currently because it has a folder and a .msk file albiet an empty one. I think that can get confusing for people. "Oh, why is this folder here I'm going to delete it." That would be bad.

So I'll propose two different ways to do saving. If people have other ideas they can add them. Both ways are just uncompressed files. This way the Git would work as well as simple ways to do images and if we wanted any other kind of file. (Maybe this should be a different issue but adding things like a Krita map to Manuksript and be able to press a button to open it in Krita So all your files are in the same place. I don't know if it fits what Manuksript is supposed to be, but it's an idea.)

First is Bibsco's way. I don't know if anyone checks on that project, but they have a folder where they store files for each project. It makes it look modern and clean I thought, butsome people might not like that.

[...]

PR #1332, opened 2024-10-19
fix: when loading ignores files that are not text files.

AlexBCa:
This contribution fixes issue [https://github.com/olivierkes/manuskript/issues/1323](https://github.com/olivierkes/manuskript/issues/1323).

In this small change, an exception is handled if the file cannot be opened because it is not a text file; it simply ignores it and loads the project.

This is my first contribution to the project, and I wanted to do something simple, although I have ideas for refactoring the project loading to make it more robust. If you are open to some refactoring, I would appreciate any feedback so I can get started, knowing what is currently being sought in the project.

### DIFF
--- a/manuskript/load_save/version_1.py
+++ b/manuskript/load_save/version_1.py
@@ -694,6 +694,10 @@ def loadProject(project, zip=None):
                         filename = os.path.join(dirpath, f)
                         with open(filename, 'rt', encoding="utf8") as fo:
                             files[os.path.join(p, f)] = fo.read()
+
+                    except (UnicodeDecodeError, FileNotFoundError, IsADirectoryError) as e:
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
Sounds good to me. Theoretically I've already started on a separate branch to make bigger changes to the IO code. You can find it here:

- https://github.com/olivierkes/manuskript/tree/gtk/manuskript/io
- https://github.com/olivierkes/manuskript/tree/gtk/manuskript/data

Also I think most details about the changes I made so far can be read in issue https://github.com/olivierkes/manuskript/issues/1117. So feel free to look into it and change or improve it. I'd also be happy to see someone try porting equivalent changes into the Qt code base since my changes only work on the GTK port/branch. But either way making the IO code more reliable is great.

Issue #1334, opened 2024-10-28
Wrong Accelerator Key


fattybamboo:
Wrong accelerator key in manuskript_zh_CN.ts (and may also in manuskript_zh_CN.qm), looks like a typo.
![wak](https://github.com/user-attachments/assets/9746b634-6af3-4da6-b27e-7ad095804ad9)

PR #1339, opened 2024-10-30
Update manuskript_zh_CN.ts

fattybamboo:
Fix a typo.


TheJackiMonster:
I assume this fixes #1334, right?

TheJackiMonster:
In the future, please use Weblate (https://hosted.weblate.org/projects/manuskript/translations/) for making changes to the translation files. This is the recommended option because it allows people who know the languages to make proper merges to find the fitting translation of each label.

Otherwise it could happen that the maintainer on Github needs to guess merges from different translations of the same language without even knowing the language. ^^'

fattybamboo:
> I assume this fixes #1334, right?

Yes.

PR #1378, opened 2025-07-29
feat(settings): add configurable tooltip styling with system‑default toggle

MustardBell:
**UPD**

Initially, this PR just fixed a Linux-specific tooltip contrast issue by hardcoding black text on a yellow background. After some discussion, it was clear that forcing a single style wasn't ideal, since themes vary widely.

Now the fix is more flexible:

* Added new **"Style"** tab in settings with configurable tooltip colors (text, background, border).
* Default behavior remains unchanged ("Use system defaults" checked by default).
* Users can now easily customize tooltip colors only if needed.

This approach solves the original readability issue without overriding anyone's preferences by default.

~Problem: On Linux systems, tooltips display white text on pale yellow background making them difficult to read.~
~Solution: Added QToolTip stylesheet with black text (#000000) on light yellow background (#ffffdc) for better readability on Linux.~

~Note: This is a quick fix - could be made configurable in settings for future enhancement.~


MustardBell:
low-contrast-default-snip

This is how it looked before the fix.

high-contrast-snip

This is how it looks after the fix.

TheJackiMonster:
Honestly, I don't think this is a general Linux related issue but it depends on the individual themes you are using. At least I've never encountered this issue on a bunch of different Linux distributions.

So I'm not sure whether it's the right thing to add in fixes for individual systems.

MustardBell:
You're not wrong. This is why I've changed it from a hardcoded fix into something more configurable. Instead of applying a static solution for Linux systems, users can now customize the tooltip styling through the settings.

This gives users control over their tooltip appearance regardless of their system or theme, making it a more flexible solution that caters to personal preferences.

Maybe I should add a toggle like "Use system defaults"? That way nothing changes **normally** but if something does break, users can change it

MustardBell:
Made those changes. Now, unless users have a problem with how their tooltips are rendered, nothing will change for them. They can, however, clear the checkbox and use custom colors - both if they experience technical problems and if they just want to play with the settings. If they check the checkbox again, system defaults will be used for the tooltips

TheJackiMonster:
Thanks. These changes are pretty good.

PR #1405, opened 2026-01-09
Gtk: updates

Kristof5933:
Hi, this is my take to some changes to the Gtk branch.

I tried to implement some of the elements that were missing in this branch from the official version, with some fixes : 
1. The exit button in the about dialog
2. Some refactoring for a single point of configuration of Gtk requirements
3. fixing the change of world item name, which only worked for root elements
4. Adding/removing characters from plot
5. Insert template to world view
6. drag and drop in world view
7. Handling some startup arguments to load a specific file
8. Allow the editing of the label in the outline view
9. Updating the POV combo in the outline view and reflect changes to the underlying objects
10. Add a scrollbar to the plot list in the ouline view
11. Add missing and modifiable POV column to the outline view (which also updates the combo)
12. Update the title of the outline item
13. Reflect changes of the status combo in the outline view.

In comparison with the Qt version, I must admit that the changes are easy to perform (but maybe I am biased, I have no experience with Qt, unlike GTK for which I have some experience for a project I did 10 years ago).

I hope that at some point we will be able to compare both versions in a nearly identical state and decide which way to go.


TheJackiMonster:
Okay, multiple things I noticed. I think sometimes naming is not entirely consistent like using `_onLabelSelected`. I've used the prefix with one underscore for events anyway, so the "on" is redundant. Also the naming like `_onAddingCharacter` does not give you the same information about which event exactly or what widget it caused... (should probably be `_characterPickerCharacterSelected` or something like that - so the underscore states it's an event, the widget name is obvious and the event type is clear).

Also you are not writing argument types in some methods but you do in others. If possible do it all the time since that makes everything more readable which is a huge point of the port to GTK. The old codebase is very much unreadable in many cases.

Then you have made changes to the data classes which confuse me. First of all I don't think `DropPosition` belongs in that namespace at all. It's essentially a utility class/function for UI only. That has nothing to do with the data model and I definitely don't want to have that unclear in the future with this branch.

Second you skip over processing the metadata for an outline text when its loading with optimization but in that case it is only loading metadata which needs to be processed. This seems like a breaking change to me for no real reason.

Third you add methods like `removeChild` or `addChild` that make changes to the world tree but they do not guarantee the same changes are reflected in the ID management of it. In other words they are not safe to use from outside of the data model. I did not use things like `parent.children.append(...)` before because I couldn't abstract it into a method but because I didn't want to abstract it. Adding or removing items should only be done via `addItem()` and `removeItem()`. If moving is too expensive in terms of processing it might make sense to add another `moveItem()` method but that should only be done inside the `World` class.

Besides that you made changes to the `validInt()` function because the POV in an outline item was a string. That doesn't really make sense, does it? The POV should probably become a character or None value similar as with label and status. So I assume I have simply been lazy in the past doing this properly already. However this change is like making two steps. The UI expects a unique character ID for dropdowns and such, the data is theoretically an optional int but represents a character via ID... it makes no sense to keep it as string but converting it back and forth using helper functions.

So if anything the last issue is on be not doing this properly. But that needs to be changed before merging this.

Otherwise you are doing quite some UI code for the different picker widgets. I generally like what they can do. But I don't like that so much UI design of them is hard coded when it could be done inside a .glade-file. So I would prefer for new widgets that they use some UI file(s) for the general layout but the logic is done in scripts. Because that way people unfamiliar with coding might still be able to help with design changes and the other way around.

Kristof5933:
Hi, thanks for the review. I'm working on it as soon as I can.

Kristof5933:
About the loading of metadata, the reason why I changed the code is that when the user is in the outline window and changes anything like the POV, status, etc. with the outline not being fully loaded, the background task that loads the markdown file erases the user edit. On my computer, with a large novel (each chapter contains 5.000 words), I have the time to see the word count update slowly and test this out.

So the way I understand the loading process, the only difference between optimized and fully loaded is the markdown file, the rest of the metadata being loaded from the start, so I didn't see the need to overwrite data that was already loaded and potentially changed by the user.

TheJackiMonster:
> About the loading of metadata, the reason why I changed the code is that when the user is in the outline window and changes anything like the POV, status, etc. with the outline not being fully loaded, the background task that loads the markdown file erases the user edit. On my computer, with a large novel (each chapter contains 5.000 words), I have the time to see the word count update slowly and test this out.
> 
> So the way I understand the loading process, the only difference between optimized and fully loaded is the markdown file, the rest of the metadata being loaded from the start, so I didn't see the need to overwrite data that was already loaded and potentially changed by the user.

I see. So it loads the metadata on startup but then overwrites it later on when it loads the markdown... yeah, that needs to be solved, definitely. I'll look into writing a fix for that so the data outline items keep such temporal changes.

Your workaround essentially disabled loading the metadata initially but it was likely still loaded when it called the function a second time for the markdown content.

Kristof5933:
> > About the loading of metadata, the reason why I changed the code is that when the user is in the outline window and changes anything like the POV, status, etc. with the outline not being fully loaded, the background task that loads the markdown file erases the user edit. On my computer, with a large novel (each chapter contains 5.000 words), I have the time to see the word count update slowly and test this out.
> > So the way I understand the loading process, the only difference between optimized and fully loaded is the markdown file, the rest of the metadata being loaded from the start, so I didn't see the need to overwrite data that was already loaded and potentially changed by the user.
> 
> I see. So it loads the metadata on startup but then overwrites it later on when it loads the markdown... yeah, that needs to be solved, definitely. I'll look into writing a fix for that so the data outline items keep such temporal changes.
> 
> Your workaround essentially disabled loading the metadata initially but it was likely still loaded when it called the function a second time for the markdown content.

I didn't notice any metadata missing while testing, so that's why I assumed my fix was okay. Can you point out what didn't work on your side, so that I can better understand the issue?

TheJackiMonster:
> I didn't notice any metadata missing while testing, so that's why I assumed my fix was okay. Can you point out what didn't work on your side, so that I can better understand the issue?

Ah, nevermind I read it wrong. It still loaded the metadata first but then didn't overwrite. However the logic is still wrong. In theory the outline item could be loaded not using the optimization to only load metadata first and the rest later which would result in an item without metadata. So less an practical issue but theoretical. As mentioned in the first review comments, I'd like to have all data classes to function independent of the UI logic.

I've pushed a proper fix (2ee7fab) on the upstream branch now that checks the current state of the outline item. So it will only skip loading metadata on a second call that's intended to complete it. Otherwise it will always load the metadata.

Kristof5933:
I took your remarks into account and updated my code accordingly, by:
1. adding variable types
2. prefixing events with an underscore
3. Refactoring the drag and drop logic, which initially had nothing to do in the data model and restoring most of the original of the data model
4. I also extracted most of the popover UI to a glade file

If you notice anything still not in line with the expected code guidelines, please let me know, I'd gladly update my code.

TheJackiMonster:
I've made minor changes. I will probably update the other data types using Optional too in the future when they initialize attribute values with `None` and allow it as value.

So most important thing I changed was that a characters POV attribute wasn't `True` as default when loading no value from file which was not the expected behavior anymore. Also I added an icon to the search entry in the picker to resolve a warning from Gtk regarding scaling issues.

But otherwise from my perspective the PR looks good to get merged.

Kristof5933:
Thanks.

PR #1412, opened 2026-02-05
Fixed file filter for image background dialog

sylverdrag:
File filter separator was incorrect (should be " ", not ";"), and as a result, the dialog did not show jpg files, it was only accepting *.png files because that was the last entry in the filter list.


TheJackiMonster:
This is a translatable string though which is the reason for the usage of the `self.tr()` function call. So this change breaks the translation while new translations could potentially be in wrong syntax again. So maybe it would make sense if we find a solution that works in any case. Essentially so that translation can't break the syntax of the file filter.

sylverdrag:
I checked the translation files, same issue everywhere, so there's no escaping the need to fix the translations too. Fortunatelly, it's a safe find and replace operation:
```
Find: *.jpg; *.jpeg; *.png
Replace *.jpg *.jpeg *.png
```
Then the translation files need to be recompiled and it should be fine.
If there are plans to support more image formats, the filters themselves can be moved to a separate variable, outside of the translations
Either way, it doesn't work right now, so fixing it would be an improvement over having  a properly translated filter that doesn't work.

TheJackiMonster:
I think it would probably make sense to move everything inside the brackets (or even including them) outside of the translation string. Because if we make any changes here, it'll break the translations anyway.

sylverdrag:
I edited he language files, but found that my version of Qt linguist (6.4.2) is producing potentially breaking changes when I save/release the .ts language files:

- replaces all special characters with xml entities, 
- changes the language codes of some languages (ca => ca_ES; ca@valencia =>_US, da =>da_DK...)
 
What do you guys use to compile the language files? Is it ok if I just commit the modified .ts files without releasing them through Linguist? (this would force "image files (*.jpg...)" for all languages until the language files are compiled by someone else.

TheJackiMonster:
> What do you guys use to compile the language files? Is it ok if I just commit the modified .ts files without releasing them through Linguist? (this would force "image files (*.jpg...)" for all languages until the language files are compiled by someone else.

The makefile has the target `make i18n` which compiles all `*.ts` files into `*.qm` files. Generally translations and changes to these files are done via [Weblate](https://hosted.weblate.org/projects/manuskript/translations/) and I compile them before a release from the `*.ts` files. But since this commit changes an entry, I think committing that should be fine. Because like this we don't have to wait for all languages that translators create a translation manually.

sylverdrag:
Cool, thanks.

PR #1416, opened 2026-02-20
Gtk: connecting settings widget to settings class

Kristof5933:
This PR connects the settings class with the settings widget for the general, revisions and view tabs in the settings dialog. So the app now load the settings, sets the value of the widgets accordingly and any change will be saved when the project is saved.


TheJackiMonster:
I think changing anything in the settings should cause it saving them. At least during testing I assumed it would save my changes immediately. Additionally we would need some form of reloading when changes are applied. For example when the list of labels is edited, selection widgets in the editor need to refresh their list models.

I have started to implement a mechanism for that back then with the `Links` objects and `LinkAction` enum. But this was more intended for changes in one object rather than singular field changes in the settings. So maybe it makes sense to have a different mechanism for that.

Later on we would need to reload the complete UI in some way anyway allowing for changes in language selection for the UI translation.

Kristof5933:
I mimicked the original settings behavior, which expects the user to save their project in order to save their settings, but this can be changed, it makes more sense. 

Maybe I should also change the way the settings are organised. Some settings apply to the application (like language), others are specific to the project. I also find that the "Views" tab lists a lot of parameters in a single page.

TheJackiMonster:
If possible try to mimic the layout of the original settings if you change anything. I tried to do that mostly but I think the whole settings widgets from libhandy, I used don't interact well with a other containers like GtkNotebook and such. So that's probably why it ended up like this.

You can search any setting now though across different pages.

Kristof5933:
So I've added some changes to this PR :

1. Auto save on settings change
2. Updating the views/models on label/status changes
3. Fixed a tiny bug on the background loading task in editorView which was hogging the cpu on large projects
4. I couldn't change the design of the view page, due to the constraints imposed by Handy.

manuskript/ui/views/editorView.py
```diff
@@ -142,7 +143,7 @@ def __completeOutlineItem(self):
         completedItem = outlineItem
         while completedItem is not None:
             if completedItem in self.editorItems:
-                self.loadOutlineData(self.outlineItem)
+                self.loadOutlineData(completedItem)
```
TheJackiMonster: I don't think this change is right. Because it changes the behavior.

In case you would have opened an OutlineFolder, it would automatically load one of its children and open that. Sure, that might reduce the amount of automatic data loading, refreshing and updating of widgets. But it breaks expected behavior.

The call of `loadOutlineData(self.outlineItem)` was intentional to reload the current state of the editor view.

TheJackiMonster: So in case there's further optimization needed, that should happen inside of `loadOutlineData()` or `loadEditorData()` in the case of reloading the current item. Which would still need to update all stats and other potentially updated changes from its children and such.

Otherwise hogging the CPU in bigger projects is kind of expected since it's actually loading the project data to visualize proper stats in the GUI. I'd think that's better than wrong or incomplete data.

Kristof5933: Currently on startup, `self.outlineItem` is None (which makes sense, we aren't in the editor view), which calls `self.unloadOutlineData()` , after digging deeper in the code, all root elements of the editor view are updated via `self.__updateEditorOutlineItem()`, where the initial goal is to simply update the selected element.

For reference, my test project is around 950 000 words long, with 5 sections containing 5 chapters of 5 scenes. Without any fix and not opening the editor view, it takes more than half an hour to count all words and the CPU goes crazy. 

So I reverted my first attempt and added a condition. So far, I see no regression, all words are counted almost instantly and navigating through the outline and the editor shows the correct count for all levels of the project, with the proper percentage.

TheJackiMonster: image

Here, this is the regression. I open a project, the individual word counts of all children get updated properly but the sum in the right lower corner does not update. The text view does not update either.

I really don't think the loop or calling `self.loadOutlineData(self.outlineItem)` is the issue here. We could look into calculating the word counts of folders (which does not cache sums yet) or potentially skip other function calls to reduce CPU cost.

But it shouldn't be the issue in general to reload content of an outline folder once its children have loaded.

TheJackiMonster: Other idea is that we could queue a reload... so that it wouldn't call the `loadOutlineData()` method many times in a row while not visible. Then we could check when the navigation is switching towards the editor tab and reload once if queued.

Because I assume most CPU cycles go into clearing of widgets and filling them again with data. So that cost could be reduced while the widgets are in the background.

Kristof5933: Since calling `self.loadOutlineData(self.outlineItem)` with a None value triggered the whole update of both textview and word count, I added a call to that method to perform the necessary update for the last item in the queue.

TheJackiMonster: The problem I still see is that it should update when the editor view is visible during this async loading process or we say that any view of the text and such does not show the text with updates during the loading process. Potentially we would need to lock user input during that process anyway to avoid data loss because reloading the text would potentially interfere with the user changing the text.

I think it's better to still view but locking the text while loading. However we might still want to have some sort of icon or indicator showing the locked status and potentially even something to visualize which chapter/scene has loaded fully already and can be edited.

Ideally all of the loading from outline items does cause as minimal friction as possible. We could potentially replace the simple queue for loading scenes with some priority queue which processes the files, you like to edit first.

Kristof5933:
I implemented an overlay to the outline and editor while the project is loading, like you suggested.

I also implemented a small optimisation for the editor textview buffer fill, the overall performance gain is not as significant as I would have expected, but at least it injects less tags in the buffer.

TheJackiMonster:
We're gonna get rid of the whole tags long-term anyway. I've written first ideas about how the editor for outline needs to function in issue https://github.com/olivierkes/manuskript/issues/1419. You don't need to worry about optimizing the existing code for that. That wouldn't work for proper scaling with tools like spell checkers and and further plugins.

Kristof5933:
No trouble, I removed my optimisation, so it won't get in the way of a rewrite. Does the overlay work the way you expected ?

TheJackiMonster:
Well, it's kind of hard to see it with the changes I made that are merged. Any project I have loads instantaneously. So it skips the overlay.

Kristof5933:
That's true, you need a large project for this to happen. 

Here's a test project I use, just rename the .zip extension to .msk, the content is made of random words:

[large_no_revisions.zip](https://github.com/user-attachments/files/26165230/large_no_revisions.zip)

TheJackiMonster:
With the project I get short freezes now, switching to the editor or outline view. I think this is because the initial loading from upstream isn't fully asynchronous anymore. So that would need a fix.

Kristof5933:
The background loading is perfectly asynchronous and fast. However, I think that the TextView has its own performance issues that we won't be able to overcome with large texts. Rendering the TextView blocks the full gtk main loop, so nothing gets refreshed as long as this task isn't finished, giving the short freezes.

I did a small benchmark that leads me to think that when word wrapping is on (which is the case for most text editors), the algorithmic complexity is not linear.

Kristof5933:
Here's the benchmark I used to measure how much time it takes for Gtk to redraw a text view with increasing larger texts.

[textViewBench.py](https://github.com/user-attachments/files/26263404/textViewBench.py)

TheJackiMonster:
> Here's the benchmark I used to measure how much time it takes for Gtk to redraw a text view with increasing larger texts.
> 
> [textViewBench.py](https://github.com/user-attachments/files/26263404/textViewBench.py)

Nice! I'd suggest subtracting the fixed timeout from the duration for results but it definitely shows the effect on my system that numbers jump heavily once it reaches 100000 words.

I think in that case it might be smarter from us if we split the text editor into multiple TextView elements that could be appended into a vertical box layout with a single scrollbar for all of them. That way we could more easily separate the text from all scenes, manage changes in individual files with less complexity and don't run into these scaling issues from word wrapping.

It might just need a little more work to make it look and behave like one continuous text view. For example we would need to implement focus and cursor jumping on inputs like arrow keys or text selection via mouse or Ctrl+A.

Kristof5933:
A quick thought : I have a feeling that selection across multiple textviews might not work, because selection is tied to the underlying TextBuffer. 

However, this is already the case in the Qt version, so it wouldn't be a loss of functionnality.

Kristof5933:
The single TextView is now replaced by a series of TextViews. As I suspected, selection is not manageable across several TextViews, because GTK prevents us from doing so.

On startup, the editor on my large test project is still slow, maybe just a tad faster. However, once fully loaded and GTK stabilized (I don't really know what's happening, but Manuskript takes some time to be fully responsive once all of the python code has run), switching between every entry in the TreeView is smooth.

Kristof5933:
@TheJackiMonster when you have a moment, could you take a look at my last changes? I'd appreciate feedback on the new implementation of the editor. Thanks!

TheJackiMonster:
For some reason it doesn't load the content automatically with your changes. I have to switch selections or resize the overall view for it to show the latest text updates, it seems.

To reproduce this:
1. I start with the default project opened.
2. I switch to editor view.
3. I wait.

TheJackiMonster:
Overall it seems like the loading times of the editor view during startup is higher with your changes compared to the upstream `gtk` branch.

Kristof5933:
I did some testing, with my large document, on my branch between the version without and the version with the  the last commit, where the editor was changed to a series of textviews instead of a single one. The least I can say is that I'm confused. Depending on the path I take in the app, the behavior changes and the time it takes for the app to be ready sometimes favor one version against the other.

PR #1433, opened 2026-07-10
Add missing source files to translation settings

MrShark:
Two source files and two translations where missing in the `manuskript.pro` configuration

Then updated the `ts-`files via make translation


TheJackiMonster:
Please use Weblate in the future to contribute to translations to ease proper merges.
https://hosted.weblate.org/projects/manuskript/translations/

PR #1439, opened 2026-08-28
ci(packaging): fallback to PyQtWebEngine for Debian 13 compatibility

### LINKED ISSUE #1438: Debian package 0.17.0-1 depends on unavailable python3-pyqt5.qtwebkit on Debian 13 (Trixie)
The official `manuskript-0.17.0-1.deb` package cannot be installed on Debian 13 (Trixie) because it declares `python3-pyqt5.qtwebkit` as a dependency, but this package is not available in Debian 13.

I downloaded `manuskript-0.17.0-1.deb` from the official Manuskript release and inspected its package metadata:

```text
$ dpkg-deb -I ./manuskript-0.17.0-1.deb | grep -E '^( Package| Version| Depends:)'
 Package: manuskript
 Version: 0.17.0-1
 Depends: python3, python3-pyqt5, python3-pyqt5.qtwebkit, libqt5svg5,
```

My system:

```text
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
DEBIAN_VERSION_FULL=13.6

$ dpkg --print-architecture
amd64
```

The dependency is unavailable:

```text
$ apt-cache policy python3-pyqt5.qtwebkit
python3-pyqt5.qtwebkit:
  Installed: (none)
  Candidate: (none)
  Version table:
```

Attempting to install the dependency therefore fails:

```text
$ sudo apt install python3-pyqt5.qtwebkit

Package python3-pyqt5.qtwebkit is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'python3-pyqt5.qtwebkit' has no installation candidate
```

marosoft:
Debian 13 (Trixie) removed python3-pyqt5.qtwebkit due to QtWebKit deprecation. Update debian/control dependency metadata to allow python3-pyqt5.qtwebengine as the primary backend dependency while retaining python3-pyqt5.qtwebkit as a fallback.

Additional metadata updates:
- Parameterize Source archive URL using {PkgVersion}
- Fix invalid multi-value Section header to 'editors' per Debian Policy 2.4
- Remove obsolete Origin field

Closes: #1438

### DIFF
--- a/package/create_deb/control
+++ b/package/create_deb/control
@@ -2,18 +2,16 @@ Package: manuskript
 Version: {PkgVersion}
 Maintainer: Tobias Frisch <thejackimonster@gmail.com>
 Description: Manuskript open source tool for writers.
- Manuskript is an open source tool for writers.  It
+ Manuskript is an open source tool for writers. It
  provides a rich environment to help writers create
  their first draft and then further refine and edit
  their masterpiece.
-Section: office, text
+Section: editors
 Priority: optional
 Installed-Size: {PkgSizeInKb}
 Architecture: all
-Origin: Ubuntu 14.04
 Bugs: https://github.com/olivierkes/manuskript/issues
 Homepage: http://www.theologeek.ch/manuskript/
-Source: https://github.com/olivierkes/manuskript/archive/0.14.0.tar.gz
-Depends: python3, python3-pyqt5, python3-pyqt5.qtwebkit, libqt5svg5,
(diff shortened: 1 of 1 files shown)


TheJackiMonster:
Thank you for the fix.

